PR #5, opened 2021-08-22
Disable button when timer ends

spencerdezartsmith:
This PR disables the save button when the timer runs out. It also adds a subtle pulse animation to the timer for the final minute to help draw the users attention. Once the timer runs out a message is shown explaining to the user the need to save their farm every 30 minutes. 

I wasn't too sure what the messaging should be so I have just used the language listed in the ticket.

I have added a final state to the machine to handle the timer expiring as the user is required to refresh the page in order to continue.

https://user-images.githubusercontent.com/17863697/130372439-0f806616-cf68-451d-902d-1515abe6b46e.mov

Screen Shot 2021-08-23 at 8 48 30 am


adamhannigan (approved):
Awesome work, love the use of the state machine for this!

PR #27, opened 2021-10-10
Rewards Farm.sol soft fork

adamhannigan:
This PR introduces a fix to avoid the total supply dropping below halvening events.

When upgrading, the Farm contract will store the funds instead of burning them.

Every 3 days a farmer has the ability to gain rewards. This will take the funds out of the Farm contract based on the balance the user has.


spencerdezartsmith (approved):
This looks good to me. How do you imagine we will sync this to the UI? Would we set a timer on the front end once a reward is open to be able to sync with when to say that new rewards are ready. We wouldn't want people to have a pay gas fees to be told there were no rewards ready.

 It's also not really "Every 3 days" I mean it will be if you open the rewards every 3 days but they won't accumulate every 3 days if you don't open it.

adamhannigan:
@spencerdezartsmith I imagined that when the farm is opened a request would be made to check if there are any rewards available. If so display the present

spencerdezartsmith:
> @spencerdezartsmith I imagined that when the farm is opened a request would be made to check if there are any rewards available. If so display the present

Ahhh yep that makes sense.

PR #35, opened 2021-11-14
hide not upgraded fruits from inventory

Rofrtd:
This PR will hide disabled (yet not upgraded) fruits from inventory.

## To test:
* Open inventory and check if fruits yet not upgraded show.

## Image:
* Radishes yet not upgraded...
Screen Shot 2021-11-20 at 2 03 39 pm

### DIFF
--- a/src/dapp/components/ui/Plants.tsx
+++ b/src/dapp/components/ui/Plants.tsx
@@ -39,7 +39,7 @@ export const Plants: React.FC<Props> = ({
         <div id="crafting-items">
           {fruits.map((fruit) => (
             (fruit.landRequired > land.length) ?
-              (<Box />)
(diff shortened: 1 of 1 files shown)


adamhannigan:
@Rofrtd I like this. It would also be nice to include empty boxes here so they know that more plants will become available ;)

Rofrtd:
@adamhannigan Should we add multiple boxes? Or only add the number of empty boxes still to unlock?

adamhannigan:
@Rofrtd I think the exact number of boxes. We won't be able to add additional plants so it might be a bit confusing otherwise.

adamhannigan (commented):
LGTM 🚀 

I will merge this in as a few players have risen this in beta. Awesome job!

PR #40, opened 2021-11-23
Audio player

Rofrtd:
This PR will add an Audio Player with the default song.
I paired with RickFurt on this one as he wants to start contributing to the project.

Should we release this before or with the BETA release?

player

### To Test: 
* hover over the musical note
* See the player showing
* test controls (Note the `nextTrack` button won't work as we only have one song at the moment)


adamhannigan (approved):
Amazing work mate! 

Let's get this onto Beta now which we can release this weekend!

PR #44, opened 2021-12-05
Resolve animation looping issue

chris-calo:
Noticed that the loop behaviour within the Pickaxe library implementation seems to be broken, at least as of the most recent updates. This should resolve it, as well as give a little bit more flexibility to components animated via Pickaxe by granting them mutable `className` props.

This bug occurred both on the `baldman` chat icon in the `Market` component, as well as on the "Upgrade" indicator on the `Barn` component (see example below for the latter before/after).

### Example of old behaviour
https://user-images.githubusercontent.com/3837103/144729501-ebb602d4-fcf4-4424-8a2e-5e3f509221be.mov

### Example of fixed behaviour
https://user-images.githubusercontent.com/3837103/144729504-13e96463-05f1-40cd-8c42-23f34f43fd3c.mov

### DIFF
--- a/src/dapp/components/farm/Barn.tsx
+++ b/src/dapp/components/farm/Barn.tsx
@@ -102,7 +102,7 @@ export const Barn: React.FC<Props> = ({ farmSize, balance }) => {
                 </div>
                 {
                     farmSize < 17 && (
-                        <Pickaxe onClick={onUpgrade} />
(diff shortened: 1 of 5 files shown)


adamhannigan (approved):
Great spot! Thanks for improving the game.

I will approve and merge this in. The Github Actions CI will automatically deploy this within 3-5 minutes.

PR #45, opened 2021-12-05
Resolve NaN NFT count display

chris-calo:
Presently, for items not wholly available yet in the `Blockchain` component that are flagged as NFTs, the number "left" is listed as `NaN`. This PR will remove such behaviour _only_ if invalid, in order to prevent confusion with users.

### Example of old behaviour:
Bildschirmfoto 2021-12-04 um 20 20 38

### Example of fixed behaviour:
Bildschirmfoto 2021-12-04 um 20 20 28


adamhannigan (approved):
Good find. I'm sure `NaN` would confuse a lot of people 😄

Issue #24, opened 2021-10-07
Warn farmer about incorrect network


spencerdezartsmith:
When a user is on a network other than Polygon we still allow a transaction to occur which will ultimately fail. We should give feedback to the user to change to the correct network.

j923sneaks:
Maybe we can go to `loading` state whenever the network switches? I tried adding `NETWORK_CHANGED` event under `farming` and it seemed to work snice the Error modal appears. 

If the user has unsaved changes such as harvested plants, perhaps we can opt not to call `blockchain.resetFarm()` during state transition. I'm not sure if it's safe or advisable though. Might introduce some form of exploit?

PR #49, opened 2021-12-29
Transition to Loading state on NETWORK_CHANGED event

j923sneaks:
This PR is resolution for #24 

On `farming` state, include `NETWORK_CHANGED` event which transitions to `loading` state

Pros(?):
- similar behavior with `ACCOUNT_CHANGED` event

Cons:
- unsaved actions will be reset -- do users interact w/ dapps on other chains while waiting to harvest?


adamhannigan:
Good spot! This will prevent people from accidentally spending funds on the wrong chain 🚀

PR #52, opened 2021-12-30
Debugging the high gas prices issue

ethernetweb:
Haven't tested the code but do a compare to see changes. Make appropriate changes / fixes if needed.
~LiveFree

### DIFF
--- a/src/dapp/Blockchain.ts
+++ b/src/dapp/Blockchain.ts
@@ -281,9 +281,9 @@ export class BlockChain {
     onboarded();
   }
 
-  public async estimate() {
+  public async estimate(incr = 1) {
     const e = await this.web3.eth.getGasPrice();
-    let gasPrice = e ? Number(e) * 1 : undefined;
+    let gasPrice = e ? Number(e) * incr : undefined;
     const minimum = MINIMUM_GAS_PRICE * 1000000000;
     if (!gasPrice || gasPrice < minimum) {
       gasPrice = minimum;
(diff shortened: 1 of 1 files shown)


j923sneaks:
have you tried running the app locally and check the gas estimates?

ethernetweb:
> will try to run the app locally and check

I do see something that I need to fix. I'll update.
I haven't set up the environment locally to be able to run the tests.. but I may have to if I keep contributing.

ethernetweb:
Made the changes again.

It also occurs to me that the return values from getGasPrice() may be a problem with the API or the RPC endpoint, so testing locally might not even provide the necessary data for debugging. It may require seeing the console.log from mainnet.

adamhannigan:
@ethernetweb Looks good to me. Testing and working well on mainnet.

Thanks for helping out and joining the developer community! This will hopefully help thousands of players worldwide!

FYI, we have Github Actions hooked up which automatically deploy the app on merge into the `main` branch. Your fix will be live in approximately 5 minutes. 🚀

ethernetweb:
I have found that getGasPrice() is not returning proper -or- `let gasPrice = e ? Number(e) * incr : undefined;` is not working correctly because the gas price is always being set to the MINIMUM fallback of 40 gwei.

I'm still working with the team a bit to figure out how to resolve the high-gas bug. I don't think that it is related to the Blockchain.ts anymore but is somehow triggered by the interface. TheRoulette will be working with me to replicate the problem so we can sequence the events and hopefully come up with a fix.

PR #53, opened 2021-12-30
Change token symbol to $SFF

Rofrtd:
This PR will change how the token is referenced on the game! 

Instead of $10 use 10 $SFF.

To test:
* Check the welcome modal
* Check Total Supply modal
* Check Fruit Board


adamhannigan (approved):
Great work mate! I have also gone through and changed the Sunflower Icon for consistency.

This icon will be used on QuickSwap so I think we should make it consistent in the game 👍

PR #54, opened 2021-12-30
Show Christmas Tree NFT only during December and before/on the 25th

chris-calo:
Does what it says on the tin – figure people are likely getting tired of seeing a Christmas tree at this point. Still retains its implicit location, even when invisible, and also maintains it in the inventory.


adamhannigan (approved):
LGTM. Nice work! 🤗

PR #55, opened 2021-12-30
Resolve z-fighting of the bald man NPC and the nearest gold miner NPC

chris-calo:
If the gold miner positioned closest to the bald man is active, the bounding box has z-fighting issues. This makes it difficult to click either one. This PR resolves that.

#### Before:

https://user-images.githubusercontent.com/3837103/147791419-b6221bb2-dfc1-477e-b168-a5a53e09c1c9.mov

#### Fixed:

https://user-images.githubusercontent.com/3837103/147791426-4562ea47-f96a-4935-b15c-439c7f2c2f80.mov

### DIFF
--- a/src/dapp/components/farm/Gold.tsx
+++ b/src/dapp/components/farm/Gold.tsx
@@ -129,12 +129,14 @@ export const Gold: React.FC<Props> = ({ inventory }) => {
         return (
           <div
             style={gridPosition}
-            className={classnames("gather-tree", {
-              "gatherer-selected": isHighlighted,
-              gatherer: isNextToChop,
-              "game-object": true,
-              gold: true,
-            })}
+            className={classnames(
+              "gather-tree", 
+              "game-object",
+              "gold", {
+                "gatherer-selected": isHighlighted,
+                gatherer: isNextToChop,
+              }
+            )}
             onClick={isNextToChop ? open : undefined}
           >
             <img src={rock} className="rock-mine ore" alt="tree" />
--- a/src/dapp/components/farm/Gold.tsx
+++ b/src/dapp/components/farm/Gold.tsx
@@ -30,6 +30,7 @@ import {
 import { Inventory, items } from "../../types/crafting";
 
 import "./Trees.css";
+import "./Gold.css";
 
 const ROCKS: React.CSSProperties[] = [
   {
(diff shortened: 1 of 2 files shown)


src/dapp/components/farm/Gold.tsx
```diff
@@ -69,7 +69,7 @@ export const Gold: React.FC<Props> = ({ inventory }) => {
       load();
       setAmount(0);
     }
-  }, [machineState.value]);
+  }, [machineState, machineState.value]);
```
adamhannigan, line 72: Will the reference of `machineState` change in this case?

src/dapp/components/farm/Gold.tsx
```diff
@@ -125,6 +125,32 @@ export const Gold: React.FC<Props> = ({ inventory }) => {
           !machineState.matches("mining") &&
           (isNextToChop || isHighlighted);
 
+        const waitingBoundLimiterContainerStyle = {
```
adamhannigan, line 128: Where possible try to create a sibling css file and import that.

E.g. `Gold.css` 

In the future we are looking to use JSS stylesheets with Typescript support but for now we do it the old fashion way with CSS 😅

adamhannigan (commented):
Great initiative! Just a few small bits of feedback. @chris-calo 

Splitting the CSS into a separate file should help us refactor into JSS in the future.

src/dapp/components/farm/Gold.tsx
chris-calo: Oh, local linter must have caught this one. I'll resolve it.

src/dapp/components/farm/Gold.tsx
chris-calo: Yeah, that makes sense @adamhannigan – can do!

chris-calo:
@adamhannigan this has been broken out into a new `Gold.css` file, pending approval 👍

src/dapp/components/farm/Gold.tsx
chris-calo: @adamhannigan I retained this, as it doesn't appear to be affected erroneously by updates to `machineState`, since it's not a deep check of the parent object. It's proper to include it as, if the reference is no longer valid the `value` child must also have changed.

Rofrtd (approved):
Awesome work @chris-calo! 
This looks good to me!

src/dapp/components/farm/Gold.tsx
```diff
@@ -131,10 +132,12 @@ export const Gold: React.FC<Props> = ({ inventory }) => {
             className={classnames("gather-tree", {
               "gatherer-selected": isHighlighted,
               gatherer: isNextToChop,
+              "game-object": true,
```
spencerdezartsmith, line 135: Rather than putting these in the object with `true` value, you can pass them as a string along with `gather-tree`. 

```
classnames('gather-tree', 'game-object', 'gold', { ... })
```

spencerdezartsmith (changes requested):
Great job @chris-calo. Just one change in the `classnames` structure. Once that is done I will approve and merge.

src/dapp/components/farm/Gold.tsx
chris-calo: @spencerdezartsmith this is resolved in the latest commit! 😁

spencerdezartsmith (approved):
Legend!

PR #56, opened 2021-12-30
Adds linting and pre-commit hook

chris-calo:
@adamhannigan It's worth noting I chunked this out into 3 distinct pushes:

1. [0x26eff25](https://github.com/sunflower-farmers/sunflower-farmers/pull/56/commits/26eff259652ada1ff56bcc026a46ec61ac40dd64) adds the mechanisms for linting and Git pre-commit hooks, based on repo requirements. The implementation is done via a combination of ESLint and Prettier (both of which are compatible with more text editors and IDEs; can provide ideal ViM config if at all helpful).
2. [0x856bc2a](https://github.com/sunflower-farmers/sunflower-farmers/pull/56/commits/856bc2af8085766dc1b810e616bd603b9c5f203e) actually runs `yarn lint-fix`, resolving all "issues" in the repo (this likely bears discussion, as it just uses my preferred linting styles for such a project – also ignores Solidity files, however that can be implemented in a follow-up PR).
3. [0x818f9f](https://github.com/sunflower-farmers/sunflower-farmers/pull/56/commits/8183f9fa7a9e5fc5a5666e34229375a2f21042ba) adds the pre-commit hook, which will be installed for all new users – this is done to ensure proper code formatting, even if someone's dev environment is improperly configured.

PR #57, opened 2021-12-31
Lint project in GitHub Action, if `lint` NPM script is present

chris-calo:
This piggybacks on the work done in #56, however is not dependent on it – if merged, it will just begin working once the `lint` NPM script is present 😉


spencerdezartsmith (approved):
🚀

PR #61, opened 2021-12-31
Redesign of chickens to match overall aesthetic

chris-calo:
It wasn't matching well, so made some custom ones. Only a single animation, but can always add more.

#### Example of old:

https://user-images.githubusercontent.com/3837103/147840839-312e41d3-a650-417c-a6f9-84e99c18dfbe.mov

#### Example of new:

https://user-images.githubusercontent.com/3837103/147840843-7be510cb-23ed-4bac-9c2e-9993437ffef3.mov


adamhannigan (approved):
Another great PR 👏

Issue #62, opened 2022-01-01
[BUG] NPCs resized when mining Gold only


j923sneaks:
reported by RovingRover on discord
https://discord.com/channels/880987707214544966/881648372249939980/926684256716087386

![image](https://user-images.githubusercontent.com/89294757/147843700-164da0d4-9e7e-4570-b04d-f9a97ce256e9.png)

Steps:
1. Mine Gold -- some NPCs resized after clicking Mine and MM prompt

Does not happen when clicking other resources, only gold

j923sneaks:
verified. will close issue now
![image](https://user-images.githubusercontent.com/89294757/147844358-41c85093-7432-40ee-9bcd-3782513eb911.png)

PR #64, opened 2022-01-01
Fix UI reflow when mining gold

chris-calo:
This solves issue #62 by enabling `overflow` only when the mining animation is playing.


spencerdezartsmith (approved):
🎉

PR #72, opened 2022-01-02
Create dynamic message for supply halving. Closes #69

### LINKED ISSUE #69: Create dynamic message for supply halving
Presently, the UI shows a static message for 1.000.000 token halving event. We've already surpassed that; should be pegged to [the maths in the white paper,](https://docs.sunflower-farmers.com/tokenomics#the-halvening) instead.

#### Example in UI:

Bildschirmfoto 2022-01-01 um 20 10 31

chris-calo:
#### Example before:

Bildschirmfoto 2022-01-01 um 20 10 31

#### Example after:

Bildschirmfoto 2022-01-01 um 20 44 22

### DIFF
--- a/src/dapp/components/ui/HalveningBanner.tsx
+++ b/src/dapp/components/ui/HalveningBanner.tsx
@@ -5,7 +5,12 @@ import React, { useState } from "react";
 
 import closeIcon from "../../images/ui/close.png";
 import alert from "../../images/ui/expression_alerted.png";
-import { service } from "../../machine";
+import {
+  BlockchainEvent,
+  Context,
+  BlockchainState,
+  service,
+} from "../../machine";
 import { numberWithBreaks } from "../../utils/number";
 import { getNextHalvingThreshold } from "../../utils/supply";
 
@@ -49,8 +54,9 @@ export const Banner: React.FC = () => {
       <div>
         <span>
           When total supply reaches{" "}
-          {numberWithBreaks(nextHalvingThreshold)}, crop and upgrade prices
-          will be divided by 1/5. Be prepared!
+          {numberWithBreaks(nextHalvingThreshold.amount)} crop and upgrade
+          prices will be divided by {nextHalvingThreshold.halveningRate}.
+          Be prepared!
         </span>
         <a
           href="https://docs.sunflower-farmers.com/tokenomics#the-halvening"
--- a/src/dapp/components/ui/HalveningBanner.tsx
+++ b/src/dapp/components/ui/HalveningBanner.tsx
@@ -41,7 +41,6 @@ export const Banner: React.FC = () => {
 
   if (!show || !totalSupply) return null;
 
-  console.log("totalSupply : ", totalSupply);
   const nextHalvingThreshold = getNextHalvingThreshold(totalSupply);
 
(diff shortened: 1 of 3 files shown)


src/dapp/components/ui/HalveningBanner.tsx
```diff
@@ -1,24 +1,57 @@
...
+
+    load();
+  }, [machineState]);
+
+  if (!show || !totalSupply) return null;
 
-  if (!show) {
-    return null;
-  }
+  console.log("totalSupply : ", totalSupply);
```
zinkkrysty, line 44: this should be removed I think

chris-calo: Good catch @zinkkrysty – thanks! Resolved!

src/dapp/components/ui/HalveningBanner.tsx
```diff
@@ -1,24 +1,56 @@
...
   return (
     <div id="halvening-banner">
       <img src={alert} />
       <div>
         <span>
-          When total supply reaches 1 Million, crop and upgrade prices will be
-          divided by 1/5. Be prepared!
+          When total supply reaches{" "}
+          {numberWithBreaks(nextHalvingThreshold)}, crop and upgrade prices
+          will be divided by 1/5. Be prepared!
```
adamhannigan, line 53: Each halvening period alternates between 1/2 and 1/5.

For instance the next halvening will be 1/2. Consider adding some logic to display the correct division as well :)

src/dapp/components/ui/HalveningBanner.tsx
```diff
@@ -1,24 +1,56 @@
...
+
+  const [machineState, send] = useService<
+    Context,
+    BlockchainEvent,
+    BlockchainState
+  >(service);
+
+  React.useEffect(() => {
+    const load = async () => {
+      const notReady =
```
adamhannigan, line 24: Consider checking `isConnected` on the `machineState.context.blockChain`.

Do you think we could use this to determine if we can show the warning or not?

adamhannigan (commented):
Great work! This will make halvening periods a lot easier.

I have just left a few suggestions

src/dapp/components/ui/HalveningBanner.tsx
```diff
@@ -1,24 +1,56 @@
...
 import closeIcon from "../../images/ui/close.png";
-
-import "./Banner.css";
+import alert from "../../images/ui/expression_alerted.png";
+import { service } from "../../machine";
+import { numberWithBreaks } from "../../utils/number";
+import { getNextHalvingThreshold } from "../../utils/supply";
 
 export const Banner: React.FC = () => {
   const [show, setShow] = useState(true);
```
trentrand, line 13: Would be nice to only show the banner as we approach the final ~25% of the next threshold

adamhannigan:
@chris-calo I hope you don't mind but I added the dynamic halvening rate in there.

Thank you for your contribution! 🤗

adamhannigan (approved):
Great work! Can't wait to see all the amazing features you have in mind!

This game is going live to hundreds of thousands of people, you should be proud!

PR #77, opened 2022-01-03
Fix incorrect value for getMarketRate after supply hits >=1b tokens

zhongfu:
after token totalSupply reaches 1e27 (1 billion tokens), the market rate becomes totalSupply / 10000

that's >= 1e23, or 100 sextillion, instead of the intended 100000

which means that seed/fruit/land prices will become 0 (since they're all much lower than 1e23)

### DIFF
--- a/src/contracts/Farm.sol
+++ b/src/contracts/Farm.sol
@@ -462,7 +462,7 @@ contract FarmV2 {
         }
 
         // 1 Farm Dollar gets you a 0.00001 of a token - Linear growth from here
-        return totalSupply.div(10000);
+        return totalSupply.div(10000 * 10**decimals);
     }
 
     function getMarketPrice(uint price) public view returns (uint conversion) {


j923sneaks:
isnt 1 billion 1e9 ?

1e9 / 10000 = 100000

zhongfu:
> isnt 1 billion 1e9 ?
> 
> 1e9 / 10000 = 100000

ERC20 `totalSupply()` returns total supply in wei; 1 token = 10**decimals wei

`SFF.decimals() == 18`, so 1 billion tokens = 1e18 * 1e9 wei = 1e27 wei

j923sneaks:
> > isnt 1 billion 1e9 ?
> > 1e9 / 10000 = 100000
> 
> ERC20 `totalSupply()` returns total supply in wei; 1 token = 10**decimals wei
> 
> `SFF.decimals() == 18`, so 1 billion tokens = 1e9 * 1e18 wei = 1e27 wei

you are right. every check also has this multiplier so it makes sense

adamhannigan:
@zhongfu You are right, good catch. 

There is essentially a soft total supply limit at 1 billion tokens. You can farm but rewards are miniscule :)

PR #79, opened 2022-01-03
[FIX] Display nextHalvingThreshold in the MarketModal

mattheuspirovani:
The current version of MarkeModal its not using the amount variable, from nextHalvingThreshold object, ocurring in a wrong exihbition of next halving threshold. Changed to use the amount variable and to format in the screen its using numberWithBreaks from util/number.


adamhannigan (approved):
Great spot!

PR #81, opened 2022-01-04
Change NGO Ordering

Rofrtd:
We've noticed the first NGO option is getting a lot more donations, and we want the other NGOs to have the same chance on getting more donations.

This is a quick PR just for changing the order the NGOs show on the donation modal

NEW ORDER:

Screen Shot 2022-01-04 at 11 11 17 am
.


adamhannigan (approved):
You are doing God's work ⭐

PR #82, opened 2022-01-04
Add bull nft

spencerdezartsmith:
This is the initial work done for the bull nft. Price reflects the fact that you will be able to make more cows using this resource. Contract has been deployed `0xd9145CCE52D386f254917e481eB44e9943F39138`


adamhannigan:
@spencerdezartsmith Great work mate.

My only concern is the similarities with this bull - https://www.dreamstime.com/stock-illustration-bull-vector-illustration-cartoon-pixel-design-image70220266 - Could we tweak it to avoid any copyright (especially since we are using @sunflowerfarmz)

Also, it appear that address isn't where the smart contract is. The game is failing to load and I couldn't connect to it via remix 🤔 

Screen Shot 2022-01-04 at 9 44 20 am

spencerdezartsmith:
> @spencerdezartsmith Great work mate.
> 
> My only concern is the similarities with this bull - https://www.dreamstime.com/stock-illustration-bull-vector-illustration-cartoon-pixel-design-image70220266 - Could we tweak it to avoid any copyright (especially since we are using @sunflowerfarmz)
> 
> Also, it appear that address isn't where the smart contract is. The game is failing to load and I couldn't connect to it via remix 🤔
> 
> Screen Shot 2022-01-04 at 9 44 20 am

Ahhh yep ok. Lets hold off them. I will put some more time into it tonight.

PR #91, opened 2022-01-05
Reduce gas used in farm contract

alisonjsilva:
Reduce the amount of gas used in functions by decreasing the number of variables in farm contract.
I know with this changes you need deploy a new contract farm, but I think this will reduce the gas when sync our farms.
As polygon at this time having issues with gas, a little minus of gas maybe is important.


adamhannigan:
@alisonjsilva Great contribution!

Every variable counts :) The team is currently playing out a few different approaches. One of the potential solutions is a hard fork. 

We will keep these improvements in mind if we go that route 🤗

alisonjsilva:
Thanks for the answer. 
If the price of gas increases a little more players may not gain by syncing.
Right now it costs 0.28 Matic to save. And the farm contract is responsible for consuming 43% of all gas in the network 😬

martymonero:
Hi @alisonjsilva ,

don't want to discredit your work, but have you double checked that your changes really reduce gas costs? Redundant assignments are something I would expect a compiler to optimize away :) (see : [https://docs.soliditylang.org/en/v0.8.10/internals/optimizer.html#redundantassigneliminator](https://docs.soliditylang.org/en/v0.8.10/internals/optimizer.html#redundantassigneliminator)

Just want to make sure your work is not in vain ;)

alisonjsilva:
Hi @martymonero,
Course, any input is valid :)
Yes, I tested the gas consume, simulating the functions execution.
In the Solidity documentation, about RedundantAssignEliminator, its can remove redundant, but you still having the variable using memory. E.g.
```solidity
uint one = returnOne();
uint two = returnTwo();
uint sum = one + two;
```
The variable one and two will not be removed because this point to on function return.
If you look the solidity example:
```solidity
{
    let a_1 := 1
    let a_2 := mload(a_1)
    let a_3 := sload(a_2)
    sstore(a_3, 1)
}
```
Even a_1 pointing to a fix number, it not removed and it passed to `mload(a_1)`
What I do to reduce this cost?!

`uint sum = returnOne() + returnTwo();`

And in the RedundantAssignEliminator you have 3 variables, to reduce cost in this case the only approach needs to be:
```solidity
{
    sstore(sload(mload(1)), 1)
}
```
But! The real problem in this contract is the for loop, testing and saving event by event.
It's good, because you can't exploit the contract without pay for that, but as Polygon haven't many validators, the game congest the entire network.

Thank you very much for your comment!

martymonero:
Ok great job! Thanks for the fact checking :)

Perhaps some food for thought: Maybe the struct of the Farm could be reduced to one Integer , like Bombcrypto does for their BHero NFT : https://bscscan.com/address/0xca0a16f4c0f34fb93383f9aaad88384aedffff39#code

They encode all details in one uint256 and encode/decode them via bit shifting .  For the farm some parts will be dynamic, so one would basically need to apply some techniques from network protocols, e.g. 00 02 ... where 02 is the number of lands and so on. Needs to be discussed if this could be applied for Sunflower Farmers, as well. And if it really would reduce costs :) Keep me in the loop! It is an interesting project for sure!

Here is the example encoding from BHero:

```
function encode(Details memory details) internal pure returns (uint256) {
    uint256 value;
    value |= details.id;
    value |= details.index << 30;
    value |= details.rarity << 40;
    value |= details.level << 45;
    value |= details.color << 50;
    value |= details.skin << 55;
    value |= details.stamina << 60;
    value |= details.speed << 65;
    value |= details.bombSkin << 70;
    value |= details.bombCount << 75;
    value |= details.bombPower << 80;
    value |= details.bombRange << 85;
    value |= details.abilities.length << 90;
    for (uint256 i = 0; i < details.abilities.length; ++i) {
      value |= details.abilities[i] << (95 + i * 5);
    }
    return value;
  }
```

adamhannigan:
@martymonero That makes a lot of sense. That array struct is really killing the gas at the moment 😓

adamhannigan:
Closing for now as the new contract design does little business logic. These contracts will be added to the repo shortly.

PR #93, opened 2022-01-05
[feat] load farm from other wallets id

GianfrancoCorrea:
### Feature: Farms explorer.
View other farms searching by his wallet address.

First time developing with TypeScript, Web3 and xstate, so feel free to review/comment me anything.

QA:
- should open a modal on clicking explore farm button.
- should display loading message on search.
- should display new farm after search.
- should not have actionable buttons on exploring farm.
- should go to own farm after click "Go Back".
- should log in console if get blockchain data fails (could be a error text in the modal as improvement)

TODO: 

- ~~clean code~~
- ~~remove hardcoded wallet id~~
- ~~add input for wallet id (modal maybe)~~
- ~~add loading when fetching data~~
- ~~remove interactions in exploring farm~~
- ~~rewording labels~~
- ~~ask for css & design help :)~~ _nobody answered :(_
- ~~label to show actual farm address~~

actual state: 

https://user-images.githubusercontent.com/46355516/148410788-3ec8ce4c-4f6a-4c95-9e95-bb5e1ce19542.mp4

image


src/dapp/Blockchain.ts
```diff
@@ -372,6 +374,52 @@ export class BlockChain {
     };
   }
 
+  private async getExploringAccount(exploringAccount): Promise<Account> {
+    if (!this.web3 || this.isTrial) {
```
adamhannigan, line 378: We can assume that if they were able to explore an account, that they are already connected to web3 and not in trial mode.

I would consider removing this check here.

src/dapp/components/farm/Farm.tsx
```diff
@@ -69,6 +71,14 @@ export const Farm: React.FC = () => {
     BlockchainState
   >(service);
 
+  const [showModal, setShowModal] = React.useState(false);
```
adamhannigan, line 74: We have got a few modals in the game, consider calling this something a bit more descriptive.

E.g. `showExploreModal`

src/dapp/Blockchain.ts
```diff
@@ -372,6 +374,52 @@ export class BlockChain {
...
+            createdAt: 0,
+            fruit: Fruit.None,
+          },
+        ],
+        balance: 0,
+        id: exploringAccount,
+      };
+    }
+
+    const farm = await this.alchemyFarm.methods
```
adamhannigan, line 407: I'm curious if we may be able to simplify the approach. Instead of creating 2 new methods in Blockchain.ts,  could we instead update `this.account` in Blockchain.ts to point to the new farm address?

If we did this, we could just reload the farm and it should pull in the correct values.

src/dapp/components/farm/Farm.tsx
```diff
@@ -227,15 +237,84 @@ export const Farm: React.FC = () => {
...
     send("SAVE", { action: "SYNC" });
   };
 
   const safeBalance = balance.toNumber();
 
-  return (
+  const getExploringFarm = async (id: string) => {
+    send("SEARCH");
+
+    const data = await getFarm(id).catch(() => {
```
adamhannigan, line 252: XState uses an event driven approach which is fire and forget. It's good to decouple business logic (from the React component) and place it into the machine.

XState has an `invoke` field which allows you to run some logic when the state is entered. In this case you could move this async call inside of the `displayLoading` state.

src/dapp/machine.ts
```diff
@@ -202,6 +214,13 @@ export const blockChainMachine = createMachine<
         ],
       },
     },
+    displayLoading: {
```
adamhannigan, line 217: I would suggest calling this something more related to the functionality. 

`loadingFriendsFarm`

adamhannigan (commented):
This is awesome, was great to finally see some other people's farms 🤗

I have left a few suggestions around how we can restructure the Blockchain.ts/machine.ts

I also found one small issue that the music stops playing when you swap to someone else's farm. Do you have any idea what could be causing that?

src/dapp/components/farm/Farm.tsx
GianfrancoCorrea: interesting, so i can get the data returning in the src of invoke? 
`invoke: {
        id: "",
        src: async () => { return DATA HERE }
        },
        onDone: {
          target: "exploring",
        },
        onError: {
          target: "failure",
        }`

and get it like: 
`const data = send('SEARCH')`

or i have to create a state in blockchain.ts?

0xCiv:
Great stuff. Are you guys working on mumbai testnet? Seems can't make it work.

GianfrancoCorrea:
@0xCiv im on polygon mainnet, because i only dev in front is not really necessary change the network. You can ask on Discord #community-dev channel

src/dapp/components/farm/Blacksmith.tsx
```diff
@@ -137,19 +157,21 @@ export const Blacksmith: React.FC<Props> = ({
       <div
         style={{ gridColumn: "4/5", gridRow: "9/10" }}
         id="minter"
-        onClick={() => setShowModal(true)}
+        onClick={() => handleClick()}
```
Trobyss, line 160: Nit picking: Avoid recreate arrow function each render with `onClick={handleClick}`

src/dapp/components/farm/Farm.tsx
```diff
@@ -227,77 +237,159 @@ export const Farm: React.FC = () => {
...
+            balance={safeBalance}
+            onHarvest={() => {}}
+            onPlant={() => {}}
+            account={exploringFarm.data.id}
+            inventory={exploringFarm.data.inventory}
+            totalItemSupplies={totalItemSupplies}
+          />
+          <span id="save-button">
+            <Panel hasInner={false}>
+              <Button onClick={() => handleGoBack()}>Go back</Button>
```
Trobyss, line 306: Np: `onClick={handleGoBack}>`

src/dapp/components/farm/Farm.tsx
```diff
@@ -227,77 +237,159 @@ export const Farm: React.FC = () => {
...
+              </Button>
+              <Button
+                onClick={() =>
+                  window.open("https://docs.sunflower-farmers.com/")
+                }
+              >
+                About
+                <img src={questionMark} id="question" />
+              </Button>
+              <Button onClick={() => openExploreFarmModal()}>
```
Trobyss, line 362: Np: `onClick={openExploreFarmModal}> `

GianfrancoCorrea:
@Trobyss thanks for the review, its a good nit picking, in each render im creating a new memory reference for that functions, right?

src/dapp/components/farm/Farm.tsx
j923sneaks: @GianfrancoCorrea saw your [comment ](https://discord.com/channels/880987707214544966/905567246452129873/929373689710006333) on discord. i havent used xstate before but for reference you might want to check `Gold.tsx`. it sends `MINE` event with payload. then in `machine.ts` > `mining` state, invoke `src: (context, event) => console.log(event)`

So for your case

Farm.tsx
```
send({ type: "SEARCH", id: id });
```

then in `invoke`
machine.ts
```
src: async ({ blockChain }, event) => console.log(event.id)
```

let me know if that helps, still in the process of learning :P

Trobyss:
> @Trobyss thanks for the review, its a good nit picking, in each render im creating a new memory reference for that functions, right?

That's right, in fact you should think of components as functions that are called at each rendering. Everything that is declared inside (apart from the methods given by React) is re-created each time. Hence the use (in some cases) of useMemo and useCallback :)

src/dapp/components/farm/Farm.tsx
GianfrancoCorrea: @j923sneaks ohh dude thanks you, the last line is what i need
`(event as SearchFarmEvent).id`
I was doing it without the .id 😅

GianfrancoCorrea:
> > @Trobyss thanks for the review, its a good nit picking, in each render im creating a new memory reference for that functions, right?
> 
> @GianfrancoCorrea That's right, in fact you should think of components as functions that are called at each rendering. Everything that is declared inside (apart from the methods given by React) is re-created each time. Hence the use (in some cases) of useMemo and useCallback :)

yeah the useMemo and useCallback are very useful to avoid computing logic every render. With eslint for react hooks, there are warning to this.

adamhannigan:
@GianfrancoCorrea Apologies for a delayed response!

This is looking good to me. The team is doing a bit of a refactor at the moment with Sunflower Land.

I'm personally still extremely excited for this feature so I can show off my farm 💪 

The new structure of the code base will be released in a week or two. It would be best to sync up (and perhaps jump on a call) so we can figure out the best way to get this feature in 🚀

adamhannigan:
@GianfrancoCorrea I will close this PR for now as there are quite a few conflicts.

Let's bounce some ideas how we can integrate this into the new system

PR #104, opened 2022-01-07
Change display decimals

aayusharyan:
On the farm screen, the decimal points for the SSF are less which is confusing (specially for new farmers `such as me` as they are not able to understand how much they have). Change that to show more decimal points.

Before Change:
image

After Change: 
image

Another point to highlight, because of the incorrect rounding, not able to calculate the coins to farm effectively.
Also this might be helpful as the upcoming halvening will reduce the base price of sunflower even more to 0.0001 so that means using the old UI, anyone will need to harvest 5 sunflowers to see some change in balance.

### DIFF
--- a/src/dapp/components/farm/Farm.tsx
+++ b/src/dapp/components/farm/Farm.tsx
@@ -281,7 +281,7 @@ export const Farm: React.FC = () => {
           <div id="inner">
             <img src={coin} />
             {machineState.context.blockChain.isConnected &&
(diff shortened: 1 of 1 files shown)


alessandrodias (approved):

adamhannigan:
Great fix! It has been a bit busy lately.

I will close this for now, as the team has done a refactor of the code base. We would love to get this enhancement in the new setup though 🤗

PR #115, opened 2022-01-12
[Draft] DAO

ethernetweb:
I have not tested it and it is likely full of bugs. The multi-dimensional struct arrays are a bit complicated and I know they're difficult to get right (I've only done them once before) ... but this is proof of concept that should be very useful.

The features of this contract are that a user must have at least 20 tokens to place vote or create a referendum to be voted on. It will verify that they still hold the tokens at the time of tally. If they don't, their vote won't count. This vote count can be voted on and changed in the contract. As can the token address that they must hold.

This should handle multiple referendums simultaneously and people should be able to vote on each of them. It also has a feature in creating the referendum that the voter must hold X token (IE: an NFT). If this feature is not used (null argument passed), it uses the default tokens that currently require >= 20.

Let me know how to proceed from here... probably debugging and writing a rudimentary web3 front end to communicate with it... or to stop because someone else is already working on one LOL...

One thing though, I don't have a lot of time to commit to helping out but I did say that I would help when I could so hopefully this is a good start and I don't know when I'll have more free time.

### DIFF
--- a/src/sfl_dao
+++ b/src/sfl_dao
@@ -126,8 +126,12 @@ contract SunflowerFarmersDAO {
   }
 
   function getReferendum(uint _ref) public view returns(Referendum[] memory _referendum) {
-    require(Referendums.length>=_ref,"That referendum doesn't exist.");
-    return Referendums[_ref-1];
+    require(
+      _ref>-1 &&
+      (_ref-1)<_referendums.length,
+      "That referendum doesn't exist."
+    );
+    return _referendums[_ref-1];
   }
 
 
--- a/src/sfl_dao
+++ b/src/sfl_dao
@@ -1,0 +1,135 @@
+// SPDX-License-Identifier: Unlicense
+// DAO
+// Version: 1.0 for Sunflower Farmers / Sunflower Land - (unfinished)
+// Written by: LiveFree @ Discord, (ethernetweb @ github)
+// January 11, 2022
+
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+
+
+
+contract SunflowerFarmersDAO {
+  uint public ReferendumCount;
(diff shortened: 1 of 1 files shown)


adamhannigan:
@ethernetweb This is great work. You have beat us to it!

The plan is to reward players with a DAO token (in the future). This DAO would vote on gameplay mechanics and would actually get proceeds from the game (e.g. 2% of sales).

This is definitely the vision of the project. For the next few months, our focus will be on the gameplay aspects.

It is great we can start the conversation around this now. 

Do you have any other ideas how the DAO would work inside of the game?

ethernetweb:
> @ethernetweb This is great work. You have beat us to it!
> 
> The plan is to reward players with a DAO token (in the future). This DAO would vote on gameplay mechanics and would actually get proceeds from the game (e.g. 2% of sales).

Interesting concept. I could likely implement that into the contract. I assume the more DAO tokens they earn by participating in the DAO, the more of the 2% they would earn from proceeds? If it's in the future we can postpone this work until the more critical stuff is done.

> 
> This is definitely the vision of the project. For the next few months, our focus will be on the gameplay aspects.
> 
> It is great we can start the conversation around this now.
> 
> Do you have any other ideas how the DAO would work inside of the game?

I have some ideas how that could be done. It would all be decentralized of course so really it's just a matter of building a UI and interacting with the contract.

ethernetweb:
I should also say that with many DAO's, the biggest problem I always see is that it becomes a majority MOB rule where only the "richest" hold all the voting power. It would like to avoid this if at all possible.

adamhannigan:
@ethernetweb It would be interesting challenge trying to get an equilibrium of power in the voting.

Some members would ideally vote on mechanisms that would make them profitable, others would focus on building a great game.

adamhannigan:
@ethernetweb I will park this PR for now as we want be getting the DAO live pre-launch.

It would be better for collaboration/discussion if we moved this to an issue. Have you thought any more how we could achieve the DAO?

Issue #114, opened 2022-01-10
Fruit counter on favicon


avalla:
Hello, I was thinking that would be useful to add a badge on favicon with fruit counter.

I can develop it, I would add a State for Counter to App.tsx and use Context to add / remove items in Field.tsx

When timeleft goes to 0 the counter the counter will be increased, when fruit is harvested it will be decreased.

Thanks :)

adamhannigan:
@avalla This is a great idea!

Now we have the new code structure, would you be interested in jumping on this?

avalla:
Yes, I will try :)
Do you have any idea to add this feature in a cleaner way?

Thanks!

PR #134, opened 2022-01-20
[feat] issue #114 show harvestable counter on favicon

avalla:
- added a Context inside App component
- added tinycon dependency
- when timer goes to zero increment the counter
- when plant is harvested decrement the counter


src/App.tsx
```diff
@@ -11,8 +13,10 @@ import "./styles.css";
  */
 export const App: React.FC = () => {
   return (
-    <Auth.Provider>
-      <Navigation />
-    </Auth.Provider>
+    <AppProvider>
```
adamhannigan, line 16: Does the AppProvider need to wrap the entire app?

Consider moving it down to only the level on it is used. (I.e. `<Crops />`)

src/AppProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    (action: AppAction) => {
+      const newState = eventReducer(state, action);
+      setState(newState);
+      return newState;
+    },
+    [state]
+  );
+
+  return (
+    <AppContext.Provider value={{ state, appDispatcher }}>
```
adamhannigan, line 58: Do we need to expose the `state` to the components that are using this context?

It looks like they only need access to the `appDispatcher`

src/AppProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  throw new Error(`Unexpected event dispatched`);
+}
+
+export const AppProvider: React.FC = ({ children }) => {
+  const [state, setState] = useState<AppState>({
+    harvestable: 0,
+  });
+
+  const appDispatcher = React.useCallback(
```
adamhannigan, line 48: Given that there is only 1 event, consider going a simpler approach and not using an event reducer.

E.g.

```
(value: number) => {
	if (value != -1 && value !== 1) {
      throw new Error("This counter should increment or decrement by 1 only");
    }
    const harvestable = state.harvestable + value;
    Tinycon.setBubble(harvestable);
    setState({
      ...state,
      harvestable
    } as AppState);
}
```

src/AppProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      ...state,
+      harvestable
+    } as AppState;
+  }
+
+  throw new Error(`Unexpected event dispatched`);
+}
+
+export const AppProvider: React.FC = ({ children }) => {
```
adamhannigan, line 43: I would suggest giving this a unique name - AppIconProvider

adamhannigan (commented):
Great job, works well.

I have just left some comments around the structure/simplifying the state changes

src/App.tsx
avalla: I was using GameContext before, but I was thinking that it was better to not mix different states (application stuff and game stuff).

I put it at App level because maybe can be useful to keep other things :)

I will refactor it as CropsContext!

src/AppProvider.tsx
avalla: Right, but maybe in future it will be useful :)

Ok... we can add it in future :)

avalla:
There are some bugs, if you crop and seed fastly the counter will not be incremented, I also suppose there are some concurrencies bugs, I'm adding some checks to prevent multiple counter update

src/AppProvider.tsx
avalla: Done

src/AppProvider.tsx
avalla: Done, removed useCallback, I think that is not useful because there is a side-effect

avalla:
I found the bug, it was inside Soil.tsx, added the reset of setBadgeUpdated inside useEffect unmount.

I'm closing the PR and I will create a new one clean and merged with main branch, see you there ;)

Issue #129, opened 2022-01-20
Shortcuts for buying and selling in bulk


adamhannigan:
<img width="197" alt="Screen Shot 2022-01-20 at 10 10 32 am" src="https://user-images.githubusercontent.com/11745561/150345095-78e947ba-da9e-4a4c-aa1f-61d1d070003f.png">

Currently players can only buy and sell one item at a time. This is quite annoying and bad UX.

We should consider adding buttons that help them perform these actions in bulk. '+5', '+10', '+20'

Clicking these buttons would enable people to buy or sell multiple items at a time.

**Includes**

- UI buttons
- Update game state
- Unit tests
- Disable button if they do not have enough

PR #142, opened 2022-01-21
[FEAT] Bulk Buy/Sell/Craft

j923sneaks:
## Description

This PR implements issue #129. I decided to go with `+1` and `+10` increments to avoid overpopulating the modal. I feel it's a good middle ground in terms of quantity but feel free to suggest better ways.

### Buy
image

image

image

### Sell
image

image

image

### Craft
Restricted bulk crafting to ERC20 items for now
image

### Sample video
https://user-images.githubusercontent.com/89294757/150477117-440be59b-06c7-446f-87cb-8695cc4557f0.mp4


j923sneaks:
I also pushed some changes to `GameProvider` to test selling. Please comment if need to revert :)

src/features/blacksmith/components/CraftingItems.tsx
```diff
@@ -14,23 +14,25 @@ import { Craftable } from "features/game/events/craft";
...
-  const hasIngredients = selected.ingredients.every(
-    (ingredient) => (inventory[ingredient.item] || 0) >= ingredient.amount
+  const lessIngredients = (amount = 1) => selected.ingredients.some(
+    (ingredient) => (inventory[ingredient.item] || 0) < ingredient.amount * amount,
   );
-  const hasFunds = state.balance >= selected.price;
+  const lessFunds = (amount = 1) => state.balance < selected.price * amount;
 
-  const craft = () => {
+  const craft = (e, amount = 1) => {
```
adamhannigan, line 31: Consider omitting the click event parameter (`e`) as it is not needed for this function.

Below we can call it as `onClick={() => craft(10)}`

adamhannigan (changes requested):
Good code and nice work with the tests 💪 I am so grateful you are in the code again! 🤓 

I have found one small regression when manually testing. The buttons are not disabling correctly when selling crops. It works for crops which have an amount, but for other ones with no amount the buttons are still active.

See below, you can still click "Buy 1" and "Buy 10" on the potatoes even though I have no potatoes

Screen Shot 2022-01-21 at 9 15 26 am

src/features/blacksmith/components/CraftingItems.tsx
j923sneaks: Thanks for the feedback! 

> See below, you can still click "Buy 1" and "Buy 10" on the potatoes even though I have no potatoes

I forgot to default to 0 if selected is `undefined`

> Below we can call it as onClick={() => craft(10)}

My bad. At first single craft had errors because it accepted `SyntheticBaseEvent` as argument. Then I saw I still had  `onClick=${craft}` 😅

j923sneaks:
Thanks for the feedback! Pushed the fixes 😄 

image

Pranav-yadav:
@j923sneaks great 🚀
I would like to suggest some changes:
- instead of proving 2nd option (button) as `Buy / Sell 10` we can add **`input`** tag there with min limit and max limit set, where player can enter exact amount that he wants to buy or sell
> inshort why hardcode and limit to 10 only.

What do u think @adamhannigan ?

j923sneaks:
hi @adamhannigan i synced with the main branch and updated my changes :)

@Pranav-yadav I agree that it gives users more freedom to input their desired amount. Surely it can be done, in this PR or the next. What I'm thinking of is the UX, especially with this game running on mobile.

Pros of input
- more flexible

Cons(?) of input
- more user actions
  - can be considered as Pro (less actions) if setting high value e,g. 100 as opposed to clicking 10 times
- prone to error, need validation

adamhannigan:
@Pranav-yadav @j923sneaks 

The challenge we face is building a mobile friendly experience. The simple click nature makes it quite easy to navigate around and a fluid experience.

We will have the opportunity to add an input field, but we will have to get the UX/UI perfect so it doesn't disrupt the experience.

adamhannigan (approved):
Works well. Great job mate! 💪

PR #223, opened 2022-02-05
[FEAT] Map movement with WASD keys and Space to center map

ihsanser:
I am adding listeners to window to move around the map with keyboard

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

I have manually tested on my local.

# Checklist:

### DIFF
--- a/src/features/game/Session.tsx
+++ b/src/features/game/Session.tsx
@@ -1,4 +1,4 @@
-import React, {useRef} from "react";
+import React, {useRef, useEffect} from "react";
 import ScrollContainer from "react-indiana-drag-scroll";
 
 import background from "assets/land/background.png";
@@ -6,11 +6,19 @@ import background from "assets/land/background.png";
 import { Game } from "./Game";
 import { GameProvider } from "./GameProvider";
 import { ToastProvider } from "./toast/ToastQueueProvider";
-import mapMovementSetup from "./lib/mapMovement";
+import mapMovement from "./lib/mapMovement";
 
 export const Session: React.FC = () => {
+  // catching and passing scroll container to keyboard listeners
   const container = useRef(null);
-  mapMovementSetup(container)
+
+  useEffect(() => {
+    mapMovement.addListeners(container.current)
+    return () => {
+      mapMovement.removeListeners()
+    }
+  }, [container])
+
   // Load data
   return (
     <GameProvider>
--- a/src/features/game/Session.tsx
+++ b/src/features/game/Session.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, {useRef} from "react";
 import ScrollContainer from "react-indiana-drag-scroll";
 
 import background from "assets/land/background.png";
@@ -6,13 +6,16 @@ import background from "assets/land/background.png";
 import { Game } from "./Game";
 import { GameProvider } from "./GameProvider";
 import { ToastProvider } from "./toast/ToastQueueProvider";
+import mapMovementSetup from "./lib/mapMovement";
 
 export const Session: React.FC = () => {
+  const container = useRef(null);
+  mapMovementSetup(container)
   // Load data
   return (
     <GameProvider>
       <ToastProvider>
-        <ScrollContainer className="bg-green-background overflow-scroll relative w-full h-full">
+        <ScrollContainer className="bg-green-background overflow-scroll relative w-full h-full" innerRef={container}>
           <div
             id="gameboard"
             className="relative h-gameboard w-gameboard"
--- a/src/features/game/lib/mapMovement.ts
+++ b/src/features/game/lib/mapMovement.ts
@@ -1,71 +1,93 @@
 type MovementIntervals = {
-  moveUp?: any,
-  moveDown?: any,
-  moveLeft?: any,
-  moveRight?: any
+  moveUp?: ReturnType<typeof setInterval>,
+  moveDown?: ReturnType<typeof setInterval>,
+  moveLeft?: ReturnType<typeof setInterval>,
+  moveRight?: ReturnType<typeof setInterval>
 }
 
-
+let container: HTMLElement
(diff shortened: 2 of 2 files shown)


src/features/game/Session.tsx
```diff
@@ -1,18 +1,21 @@
...
 import background from "assets/land/background.png";
 
 import { Game } from "./Game";
 import { GameProvider } from "./GameProvider";
 import { ToastProvider } from "./toast/ToastQueueProvider";
+import mapMovementSetup from "./lib/mapMovement";
 
 export const Session: React.FC = () => {
+  const container = useRef(null);
+  mapMovementSetup(container)
```
adamhannigan, line 13: If we only need to set this when the ref changes consider putting it inside of a useEffect.

E.g.

```
useEffect(() => {
  mapMovementSetup(container)
}, [container])
```

src/features/game/lib/mapMovement.ts
```diff
@@ -0,0 +1,71 @@
...
+  moveRight?: any
+}
+
+
+const sensitivity = 3
+const movementIntervals: MovementIntervals = {}
+const initialCoordinates = [1024,1214]
+// TODO dynamic initial coordinates based on tile dimensions
+
+const setup = (container: any) => {
```
adamhannigan, line 14: Avoid using `any` where possible. Using types reduces bugs and will ensure if someone comes in to change the code there is less of a chance they will break something.

In this case you can type it as `container: RefObject<HTMLDivElement>`

Some other nullable warnings may pop up - this is a good thing, we will need to check if the container is null (could happen if the component unmounts)

src/features/game/lib/mapMovement.ts
```diff
@@ -0,0 +1,71 @@
...
+  // adding setintervals for smooth movement for each relevant keypress and removing them on keyup 
+  window.addEventListener('keydown', (event) => {
```
adamhannigan, line 16: We've got 2 window listeners here that don't get cleaned up.

When a user changes their Metamask account we will take them back to the login page. Once they log back into the game 2 more window listeners would be added. This will cause some strange behaviour with too many listeners.

I would suggest adding some code that cleans up these listeners when the component (`Session.tsx`) is unmounted.

You can achieve this in the components use effect by providing a disposer function.

```
useEffect(() => {
	// do some code

       return () => {
          // clean up listeners
      }
})
```

adamhannigan (changes requested):
Nice work 👏 Welcome to the dev community 🏆 

I know there are a few people discussing using the up and down arrow keys to cycle through the inventory items. It would be worth asking the devs which hot keys we should use for what.

I have left a few suggestions on the PR. The main area that needs addressing is the leaky window listeners. This will cause some strange behaviour when the user goes back to the login screen.

Pranav-yadav:
Great feat @ihsanser 🚀💪.

Few suggestions, if possible then try to add another events for commonly used/visited locations in game like,
following keys to focus/**open modals**, for;

- C: Craft
- F: Crops/Farm _(foucuses the crops/main-farm area, similar to what, the **Crops button** under menu does. Ik u've used Space for centering, but space suits best for playing/pausing music)_
- I: Inventory
- M: Shop/Market

And most important, 😊
- P/Space: Play or Pause the music 🎶
- Esc: Close the current open modal only :)

Optional,
B: display true Balance (i.e., expand the balance)


What do u think @adamhannigan ?

adamhannigan:
@Pranav-yadav These look good to me 👌 

I'm wondering what shortcut we should use for cycling the inventory items as this will be a common action. I.e. switching between selected seeds or tools 🤔

Pranav-yadav:
> @Pranav-yadav These look good to me 👌
> 
> I'm wondering what shortcut we should use for cycling the inventory items as this will be a common action. I.e. switching between selected seeds or tools 🤔

Yep, @adamhannigan, adding key shortcuts for cycling through inventory as well as shop items will be used frequently.

I think, in these case we'll first need to check if the (any) modal is opened, and the logic of key event listeners must be modified to perform specific things based on which modal is currently opened.

In short, we need to modify the logic inside event listeners such that, when a modal open the code should perform cycles operations and not move the map itself.

Will try to explain the logic in code if I get time.

Pranav-yadav:
@adamhannigan
Edit: I almost forgot for time being we can cycle through items using Tab key
<kbd>Tab<kbd>

ihsanser:
@adamhannigan thank you very much for the review! I think I have managed to get things the way you wanted. Updated types and improved event listener handling in my last commit. I haven't worked with React or TypeScript before so I am grateful for your output.
@Pranav-yadav I liked your suggestions. I wanted to keep this PR quick and simple. I can work on your suggestions right after this one.
About the "keys might be used on another menu" topic, I totally agree with you, we should only move the map if there is no open menu or something. Maybe we can add a state if there isn't already. I will check that also.

src/features/game/lib/mapMovement.ts
```diff
@@ -0,0 +1,93 @@
...
+const intervalTime = 1
+const movementIntervals: MovementIntervals = {}
+const initialCoordinates = [1024,1214]
+// TODO dynamic initial coordinates based on tile dimensions
+
+const keyDownListener = (event: KeyboardEvent) => {
+  const key = event.key.toLowerCase()
+  if (container) {
+    if (key === 'w' || key === 'arrowup') {
+      if (movementIntervals.moveUp == undefined) {
```
spencerdezartsmith, line 19: Can we update this to use `===`? It's good practice to always use strict equality in JS.

ihsanser:
@spencerdezartsmith of course. Thanks for the feedback. Updated now

adamhannigan (approved):
You have picked up Typescript fast. Great work 💪 LGTM

PR #266, opened 2022-02-14
[FEAT] Inventory tabs

jcads:
# Description

Inventory UI gets cluttered as we obtain more inventory items. I used IDs to group the items into different categories. 

In the mobile view, I opted with using arrows to move through the tabs as `touchEvents` are a little tricky to work with and can add more complexity to the code.

In desktop view, we can scroll on the tabs to view all tabs or click and drag. I was thinking of handling just the `onWheel` event but others first intuition might be to click and drag to view tabs. So I also added `mouseEvents`.

In the future as we add more categories, we would just update it in `types/index` and `InventoryItems`.

## Type of change

# How Has This Been Tested?

I ran tests by making sure the desktop and mobile view works as expected. To view this new feature just go to Inventory, there we can scroll, drag horizontally to move (desktop), and click the arrow to go to the prev/next tab.

**Scrolling**
scroll

**Click and drag**
drag

**Mobile view**
mobile

# Checklist:


jcads:
@adamhannigan @spencerdezartsmith @Rofrtd what do you think?

Pranav-yadav:
@jcads great 🚀 
Few suggestions from my side;

### You can show the Tabs **vertically** on _mobile screens_, instead of creating horizontal slider. 
_(what do u think @spencerdezartsmith ?)_

This will make the code more optimized and easy to understand.

@adamhannigan will provide you with appropriate instructions.

PS: You might need to decrease the size of the elements little bit, if you go with vertical approach. 👍

adamhannigan:
Definitely looks more decluttered 👏 @jcads 

However, we will have a usability issue when it comes to people using trackpads (mainly MacBooks). It is not very common on web interfaces to drag across to scroll. I'm worried that users would not figure out how to view the other items here. It took me awhile on my mac to figure out how to drag to scroll and even then it was a bit buggy.

Ideally, we want have an extremely clear 'affordance' (indicator) that the content is scrollable and that more categories exist.

My gut feeling is that the tabs are constraining us here (especially on mobile). I've just edited the following (only in CSS) to see what another approach could look like here.

@spencerdezartsmith @jcads @Pranav-yadav What are your thoughts on the following (for web and mobile)?

Screen Shot 2022-02-15 at 2 43 43 pm

jcads:
I just tested it on my trackpad and it works fine with the classic 2-finger scroll. However, I agree that we need a much clearer indicator that the content is scrollable.

> What are your thoughts on the following (for web and mobile)?

This is definitely cleaner in terms of showing the category. 

But I think UX issues may arise once we begin to add more categories (e.g. a category with id 10 would take a lot of clicks/taps to get there). One thing we could do later is to add some sort of a dropdown _(maybe on the top-left side)_ that contains  all the categories and the player can just click to go to that category.

What do you think @adamhannigan?

spencerdezartsmith:
> Definitely looks more decluttered 👏 @jcads
> 
> However, we will have a usability issue when it comes to people using trackpads (mainly MacBooks). It is not very common on web interfaces to drag across to scroll. I'm worried that users would not figure out how to view the other items here. It took me awhile on my mac to figure out how to drag to scroll and even then it was a bit buggy.
> 
> Ideally, we want have an extremely clear 'affordance' (indicator) that the content is scrollable and that more categories exist.
> 
> My gut feeling is that the tabs are constraining us here (especially on mobile). I've just edited the following (only in CSS) to see what another approach could look like here.
> 
> @spencerdezartsmith @jcads @Pranav-yadav What are your thoughts on the following (for web and mobile)?
> 
> Screen Shot 2022-02-15 at 2 43 43 pm

Firstly great job @jcads.

Yeah I don’t think tabs is the best UX here. They don’t quite fit which makes them looked cluttered. The arrows sitting outside is a very bespoke solution. We want to stick with common UI patterns that are intuitive. I think this example is much better.

adamhannigan:
@jcads Agreed with more tabs it will get hard to manage, for now I think it will be fine. 

A drop down could work in this case. When we get to the point of 5-6+ different categories we should take a step back and assess why players are coming to the inventory.

I imagine at that part we will figure out what the 'hotkey' items are. Things like NFTs would live in a trophy room, resources in a storage room etc :)

jcads:
Thank you for the feedback @spencerdezartsmith & @adamhannigan. I will take your suggestions into account 💯

Pranav-yadav:
@adamhannigan great, that way it'll be more clutter free.

@jcads you can set the `minWidth` for the div surrounding the Tab, as arrow positions are getting changed every time, this creates accessibility (UX) concerns as we need too move the mouse every time. (at least on PC)
Adding `minWidth` there, will improve the UX.

And, in future we can even bind KBD events such that, player would be able to move through tabs using arrow keys on KBD.
@adamhannigan  @spencerdezartsmith ?

src/features/hud/components/InventoryItems.tsx
```diff
@@ -4,35 +4,154 @@ import { OuterPanel, Panel } from "components/ui/Panel";
...
+import crop from "assets/crops/sunflower/crop.png";
+import tool from "assets/tools/hammer.png";
+import nft from "assets/nfts/gnome.png";
+import food from "assets/crops/wheat/flour.png";
+import resource from "assets/resources/wood.png";
+interface Props {
+  onClose: () => void;
+}
+
+export const categories: CATEGORY[] = [
```
adamhannigan, line 24: Instead of declaring a separate category ID type (which is directly typed to the items), consider the following pattern.


```
type Tab = 'Seeds' | 'Crops' | ...
export const CATEGORIES: Record<Tab, { img: any, items: object }  = {
  Seeds: {
   img: seed,
   // import the items
   items: SEEDS
},  Crops': {
   img: crop,
   // import the items
   items: CROPS
},
...
}
...
```

src/features/hud/components/InventoryItems.tsx
```diff
@@ -4,35 +4,154 @@ import { OuterPanel, Panel } from "components/ui/Panel";
...
+    name: "Resources",
+    img: resource
+  },
+];
+
+export const InventoryItems: React.FC<Props> = ({ onClose }) => {
   const { gameService, selectedItem, shortcutItem } = useContext(Context);
+  const scrollDiv = useRef<any>(null);
+
+  const [tab, setTab] = useState<CATEGORY_ID | number>(1);
```
adamhannigan, line 61: Again, magic numbers can get quite confusing for a new engineer in the repo. The value `1` does not provide much meaning to them. They would need to spend more time reading other parts of the code just to understand this.

Consider the following:

```
const [tab, setTab] = useState<Tab>('Crops');
```

adamhannigan (changes requested):
@jcads I have left a few suggestions around how we could use Typescript here to prevent errors.

Where possible, reduce the use of 'Magic Numbers'. It makes it a bit difficult for other engineers reading the code and can also make it more flaky.

https://itqna.net/questions/878/what-are-magic-numbers-javascript

adamhannigan (approved):
@jcads Awesome work 👏 This feels like a huge UX improvement, especially on mobile devices. Players will love this ❤️ 

Drop your Polygon address and we will whitelist you for some upcoming content :)

jcads:
> @jcads Awesome work 👏 This feels like a huge UX improvement, especially on mobile devices. Players will love this ❤️

Thank you!

> Drop your Polygon address and we will whitelist you for some upcoming content :)

I think I might have dropped it a while ago. But I'll put it here again in case you missed it: 0x068BAb172FA62635DFB3cAbCF330EdbF63522cE4

PR #310, opened 2022-02-22
[FIX] Visit farm button fix

j923sneaks:
# Description

This PR covers the ff:
- `Enter` key now loads the visiting farm
- ~~Counter now excludes locked crops~~
- ~~Counter now ignores `readonly/visit` mode~~
- Fixed inventory check for zone four in `visitUtils`

Fixes #issue N/A

## Type of change

# How Has This Been Tested?

Manual testing
~~INITIAL_FARM with Pumpkin Soup~~
Steps:
1. Load your farm -- counter should be correct and excludes locked crops~~

Visit Farm
Steps
1. Sign in metamask
2. Explore a friend's farm
3. input farm ID
4. press `Enter` key -- should load the visiting farm

# Checklist:


adamhannigan:
@j923sneaks Some great improvements! 👏 

One area that is flaky at the moment and keeps breaking is the favicon counter. Adding more business logic into the components `canHarvest` and handling logic in the `Soil.tsx` is a quick fix, but it won't make the code stable and will likely break again when one of these components is changed.

I would suggest committing only the Visit farm fix in this PR and actually removing the favicon from the code.

We need to reconsider how we can fix the root problem of the FavIcon counter instead of bandaiding it at the component level. Otherwise this issue will likely resurface. The most sustainable approach would be stripped out all of the logic from the components and handling the fav icon count based on events that are fired in the system. This would require significant work at the moment.

j923sneaks:
Hey @adamhannigan i understand. Reverted some code and retained what's not related. apologies for the many commits 😅 

I too was reluctant to have the same code for each zone only differing in the inventory item to be check. you're right that event based system from `Soil` to `AppIconProvider` is the best way. I will see how it's done in React, but is it possible or ideal to emit events in vanilla way?

adamhannigan:
@j923sneaks Luckily we have got xstate in our stack. It supports a range of different APIs that let us listen to events. Here is one for example `gameService.onSend(event => ...)`

adamhannigan (approved):
LGTM 👏

PR #318, opened 2022-02-24
[CHORE] Pixel image rendering issue

adamhannigan:
# Description

This PR introduces support for users

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "events": "^3.3.0",
     "graphql": "^16.2.0",
     "html2canvas": "^1.4.1",
+    "jwt-decode": "^3.1.2",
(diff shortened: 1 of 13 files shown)


spencerdezartsmith (approved):
Code looks good on this. Was unable to test as I was getting a cors rejection using the `https://api-dev.sunflower-land.com` url. Not sure if you have it deployed on a different instance. Also I don't think its this PR but the sign message is overflowing the metamask popup. We should probably fix as it may make people feel wary.

Approving but acknowledging I couldn't test the flow.
Screen Shot 2022-02-25 at 1 20 32 pm

adamhannigan:
Merging in for testing on testnet.

@spencerdezartsmith That is strange, in my Metamask it wraps onto the next line 🤔

PR #330, opened 2022-03-03
[FIX] polish visit feature

### LINKED ISSUE #328: [BUG] Selected shortcuts persist even if farm is not loaded
There are 2 scenarios for this bug:

**A. Selecting items in inventory when visiting are cached as shortcuts and are displayed when going back to loading your farm**
reported by [AEP] Vael - Niels#1701 from Discord

Given:
Account A -- your farm that does not have NFTs / resources e.g. Golden Cauliflower
Account B -- a farm that has NFTs / resources e.g. Golden Cauliflower

Steps:
1. Visit Account B
2. Click Golden Cauliflower under Inventory -> NFTs
3. Go back and load Account A
4. Check shortcut items in Hud -- ERROR: this will include Golden Cauliflower which Account A does not have

Expected:
In step 2, it should not cache selected item as a shortcut since it is only visiting farm

**B. Selecting items in inventory persist when switching accounts**

Given:
Account A -- your farm that does not have NFTs / resources e.g. Golden Cauliflower
Account B -- your other farm that has NFTs / resources e.g. Golden Cauliflower

Steps:
1. Load Account B
2. Click Golden Cauliflower under Inventory -> NFTs
3. Switch to Account A using Metamask
4. Load Account A
5. Check shortcut items in Hud -- ERROR: this will include Golden Cauliflower which Account A does not have

Expected:
Should cache shortcuts per farm? 

##
### Impact
Medium. It's not game breaking and it cannot be exploited. For example, selecting Axes in one account then switching to another (one who doesn't have any) cannot use those axes to chop trees. I'd argue for low severity but it may confuse users.

### Cause
Caching shortcuts only save the item names and does not map per `farmId`. So in either case, they will persist if farm changes.

### Resolution
For A, we can have a check that will only execute `shortcutItem()` if not in `readonly` mode. Then for B, maybe we can change the logic itself to include mapping between `farmId` and selected items? What do you think? 🤔

jcads:
# Description

- Remove "equip axe" popup on trees when visiting
- Visitors can shake trees on other farms
- Don't set `shortcutItem` on "readonly" state (fixes #328) 

## Type of change

# Checklist:

### DIFF
--- a/src/features/forest/components/Tree.tsx
+++ b/src/features/forest/components/Tree.tsx
@@ -71,6 +71,11 @@ export const Tree: React.FC<Props> = ({ treeIndex }) => {
   };
 
   const shake = async () => {
+    if (game.matches("readonly")) {
+      shakeGif.current?.goToAndPlay(0);
+      return;
+    }
+
     if (selectedItem !== "Axe") {
       return;
     }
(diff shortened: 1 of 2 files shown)


adamhannigan (approved):
Great fixes. Good work 👏

PR #346, opened 2022-03-06
[FEAT] Preload images / allow tx rejections

spencerdezartsmith:
# Description
This PR add a preload hook for our modal images/gifs. This prevents the case where the modal doesn't have the image when it first opens which looks janky and makes the modal jump once the image does load.

I have also set up some handling that allows the user to cancel a sync, team donation and withdraw transaction and continue playing the game rather than showing the `Something went wrong` modal forcing the user to refresh.

I have also block the ability for a user to continue clicking around (after deleting modal from the dom) if a fatal error has occurred.

Fixes #issue

## Type of change

# How Has This Been Tested?

- Run the app and see the images loaded during the during the `readyToStart` state.
- Go to the menu and hit `Sync on chain`
- The syncing image should show in the modal immediately
- Reject the tx in metamask. You should be able to continue playing the game

# Checklist:

### DIFF
--- a/src/features/game/components/GameError.tsx
+++ b/src/features/game/components/GameError.tsx
@@ -1,9 +1,17 @@
-import React from "react";
+import React, { useEffect } from "react";
 import deathAnimation from "assets/npcs/human_death.gif";
 
 export const GameError: React.FC = () => {
+  useEffect(() => {
+    const body = document.querySelector("body");
+
+    if (body) {
+      body.style.pointerEvents = "none";
+    }
+  }, []);
+
   return (
-    <div className="flex flex-col items-center p-2">
+    <div id="gameerror" className="flex flex-col items-center p-2">
       <span className="text-shadow text-center">Something went wrong!</span>
       <img src={deathAnimation} className="w-1/2 -mt-4 ml-8" />
       <span className="text-shadow text-xs text-center">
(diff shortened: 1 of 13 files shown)


src/features/game/components/GameError.tsx
```diff
@@ -1,9 +1,17 @@
-import React from "react";
+import React, { useEffect } from "react";
 import deathAnimation from "assets/npcs/human_death.gif";
 
 export const GameError: React.FC = () => {
+  useEffect(() => {
```
adamhannigan, line 5: Is the purpose of this code to prevent click actions behind the modal?

Is it possible to achieve this using a HTML overlay in this component instead that prevents pointer events behind the modal?

I'm concerned some bugs may pop up if we start transitioning out the `error` state that we are not cleaning up these values

adamhannigan (approved):
You are doing god's work mate. Tested and works well for me.

I've left just a small comment but happy for this to go in 👍

src/features/game/components/GameError.tsx
spencerdezartsmith: What would prevent the user from removing the html overlay from the dom? I would add a clean up in this use effect to remove this? That way when it’s closed pointer events are back in action?

PR #372, opened 2022-03-15
[FIX] Resource refresh bug

### LINKED ISSUE #361: [BUG] Rock/Trees reappear
**Details**

If you chop down a tree or mine a rock during the middle of an autosave, once the autosave is completes the rock/tree will reappear. If you try again you will receive an error on the next save

**Steps to reproduce**

1. Chop a tree during the middle of autosave
2. Try chop tree again

**Technical details**

This happens because the random values are coming back from the server side and updating the game state. Instead of updating the entire game local state we should instead only update the values generated on the server (i.e. Random Number values)

adamhannigan:
# Description

This PR fixes the issue that a resource would randomly reappear after being mined/chopped.

Fixes #361 

# How Has This Been Tested?

Throttle your internet to make it slow. Click save and then begin chop a tree. By the time the game saves the tree should remain chopped

### DIFF
--- a/src/features/game/lib/transforms.ts
+++ b/src/features/game/lib/transforms.ts
@@ -66,6 +66,25 @@ export function makeGame(farm: any): GameState {
   };
 }
 
+type Rocks = Record<number, Rock>;
+
+/**
+ * Updates a rock with the new amount of mineral inside of it
+ */
+function updateRocks(oldRocks: Rocks, newRocks: Rocks): Rocks {
+  return Object.keys(oldRocks).reduce((rocks, rockId) => {
+    const id = Number(rockId);
+    const rock = oldRocks[id];
+    return {
+      ...rocks,
+      [id]: {
+        ...rock,
+        amount: newRocks[id].amount,
+      } as Rock,
+    };
+  }, {} as Record<number, Rock>);
+}
+
 /**
  * Merges in server side changes into the local state
  */
@@ -80,7 +99,7 @@ export function updateGame(
   // Only update random number values generated from the server
   return {
     ...oldGameState,
-    // Update any random numbers from the server
+    // Update tree with the random amount of wood from the server
     trees: Object.keys(oldGameState.trees).reduce((trees, treeId) => {
       const id = Number(treeId);
       const tree = oldGameState.trees[id];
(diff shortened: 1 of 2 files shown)


src/features/game/lib/transforms.ts
```diff
@@ -65,3 +65,65 @@ export function makeGame(farm: any): GameState {
...
+): GameState {
+  if (!newGameState) {
+    return oldGameState;
+  }
+
+  // Only update random number values generated from the server
+  return {
+    ...oldGameState,
+    // Update any random numbers from the server
+    trees: Object.keys(oldGameState.trees).reduce((trees, treeId) => {
```
spencerdezartsmith, line 84: Could we do something like this to avoid the repetition? This could be further simplified if tree had the key `amount` instead of `wood` 😅 . Update resource could then be a generic function.

```
type ResourceData = Record<number, Tree | Rock>;

function objectKeys<Obj>(obj: Obj): (keyof Obj)[] {
  return Object.keys(obj) as (keyof Obj)[];
}

function isTree(resource: Tree | Rock): resource is Tree {
  return (resource as Tree).wood !== undefined;
}

const updateResource = (oldState: ResourceData, newState: ResourceData) => {
  return objectKeys(oldState).reduce((resourceData, itemId) => {
    const id = Number(itemId);
    const resource = oldState[id];

    if (isTree(resource)) {
      return {
        ...resourceData,
        [id]: {
          ...resource,
          wood: (newState[id] as Tree).wood
        }
      }
    }

    return {
      ...resourceData,
      [id]: {
        ...resource,
        amount: (newState[id] as Rock).amount
      }
    }
  }, {})
} 

function updateGame(
  newGameState: GameState,
  oldGameState: GameState
): GameState {
  if (!newGameState) {
    return oldGameState;
  }

  return {
    trees: updateResource(oldGameState.trees, newGameState.trees),
    stones: updateResource(oldGameState.stones, newGameState.stones),
    iron: updateResource(oldGameState.iron, newGameState.iron),
    gold: updateResource(oldGameState.gold, newGameState.gold)
  }
}```

spencerdezartsmith (commented):
Looks good mate. Just a comment about drying up the code.

src/features/game/lib/transforms.ts
adamhannigan: In hindsight it would of been good to have similar interfaces for wood + rocks. Right now it would require a migration though which is a bit out of scope 😢 

I've created a reusable function as your suggestion above but only applied it to the rocks 😄

PR #448, opened 2022-03-27
[FEAT] Skill Tree lock-unlock

jcads:
# Description

Mute the panel colors of each locked skill. Doing this gives emphasis to unlocked skills and can give players a sense of accomplishment when unlocking a new skill.

Also, I changed the bg-color of the skill level indicator for Gathering. Let me know your thoughts on this one

## Type of change

# How Has This Been Tested?

manually

image

image

# Checklist:


Z1kou:
@jcads I love this idea!  is there  any chance of further upgrading this by reducing the saturation more and or even adding a lock on screen, as a player I don't think dampening the colors is enough to show Something is unavailable!

jcads:
@Z1kou yes we can reduce the saturation more.

I also initially had thought of adding a lock but that would require additional art. Where do you think the best placecment of the lock should be? I was thinking, a) beside the skill name (**Lumberjack 🔒**); b) replace the locked badge with a lock 🤔

adamhannigan:
lock

 @jcads I created this a while ago but never used it. 

Also, I would suggest playing around with opacity instead of colours. Changing HSV values often makes the color palette inconsistent with the game. To make the UI look beautiful I would suggest one of the following:

1. Using opacity
2. Manually specifying a different colour (perhaps you could use the grey Message component)

jcads:
@adamhannigan 

## **Reduced some of the grayscale values**
image

## **Combination of grayscale, opacity, and a bg-color**

image

image

Z1kou:
@jcads  I think the one you have now is cool, also like its placement! saw the two prototype you have and yet i still feel like they are not enough i would like if it was something like this !
bvzvq8bs9pk31

spencerdezartsmith:
> @adamhannigan
> 
> ## A. **Reduced grayscale values**
> image
> 
> ## B. **Combination of grayscale, opacity, and a bg-color**
> image
> 
> image

What about a combination of grayscale and opacity. I like A. but think some reduced opacity might help. I love the little lock <3

adamhannigan:
@jcads  The palettes still look off in these examples. Unfortunately I can't give the exact colors since I'm not an artist myself. I don't think us devs can give much more feedback how to perfect that 😅 

From my view, I would prefer to have an aesthetically pleasing skill tree preview instead of the unlocking functionality that does not match the color scheme in rest of the game.

If the idea is to "give a sense of accomplishment when achieving them" can we just use the lock icon instead of the displaying the skill icon?

I also think the previous modal does a good job of a sense of accomplishment. They choose the skill and then it appears in their bagdes.

By the way, the background colour of the "Gathering" is much better 👏

jcads:
> If the idea is to "give a sense of accomplishment when achieving them" can we just use the lock icon instead of the displaying the skill icon?

image


image

adamhannigan (approved):
LGTM 🚀 We went through a lot of iterations but the UI is looking great ❤️

PR #458, opened 2022-03-29
[Fix] Update Selling Price With Skills

Z1kou:
# Description

The green thumb effect was giving 10%price increase instead of 5% stated in the game, also tried fixing the popup modul of selling when you have the skill, you will still get the updated amount but the pop over still showed the old value of before the skill

## Type of change

# How Has This Been Tested?
Yarn dev
Yarn dev Sunflower with skill
NG)

Yarn test 
Yarn test Skill

# Checklist:

### DIFF
--- a/src/features/crops/components/Plants.tsx
+++ b/src/features/crops/components/Plants.tsx
@@ -41,7 +41,7 @@ export const Plants: React.FC = () => {
      }
       else
      {
-      setToast({ content: "SFL +$" + selected.sellPrice.mul(amount).toString() });
+       setToast({ content: "SFL +$" + selected.sellPrice.mul(amount).toString() });
     }
   };
 
--- a/src/features/crops/components/Plants.tsx
+++ b/src/features/crops/components/Plants.tsx
@@ -29,20 +29,17 @@ export const Plants: React.FC = () => {
 
   const inventory = state.inventory;
 
+
+
   const sell = (amount = 1) => {
     gameService.send("item.sell", {
       item: selected.name,
       amount,
     })
-      
-    if (inventory["Green Thumb"]?.greaterThanOrEqualTo(1)) 
-     {
-       setToast({ content: "SFL +$" + selected.sellPrice.mul(1.05*amount).toString() });
-     }
-      else
-     {
-       setToast({ content: "SFL +$" + selected.sellPrice.mul(amount).toString() });
-    }
+      {
+        setToast({ content: "SFL +$" +   displaySellPrice(selected).mul(cropAmount).toString() });
+      }
+
   };
 
   const cropAmount = new Decimal(inventory[selected.name] || 0);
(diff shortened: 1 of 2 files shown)


src/features/crops/components/Plants.tsx
```diff
@@ -33,8 +33,16 @@ export const Plants: React.FC = () => {
     gameService.send("item.sell", {
       item: selected.name,
       amount,
-    });
-    setToast({ content: "SFL +$" + selected.sellPrice.mul(amount).toString() });
+    })
+      
+    if (inventory["Green Thumb"]?.greaterThanOrEqualTo(1))
```
j923sneaks, line 38: why not use `getSellPrice()` from `pricing.ts` instead?

```js
setToast({ content: "SFL +$" + (getSellPrice(selected.name, inventory).mul(amount)).toString() });
```

Z1kou: Spent alot of time trying to set it up with that, errors kept popping left and right so i came up with the alternative that I used here !

Z1kou:
After playing around with the variables for a while trying to find a format that is functioning,consistent with the code and the future code, and without errors I got this! Huge thanks to @j923sneaks  for.... Oh wait You updated the comment with the displaySellPrice() format, OH MY GOD I SPENT SO LONG TRYING TO FIX  THE ERRORS AND IT WAS RIGHT HERE 😂 ,Well this was a great learning experience got to learn more about functions with trial and error and research, @j923sneaks  Thank you bro again ❤️

src/features/game/lib/pricing.ts
```diff
@@ -7,8 +7,8 @@ import { Crop } from "../types/crops";
 export const getSellPrice = (crop: Crop, inventory: Inventory) => {
   let price = crop.sellPrice;
 
-  if (inventory["Green Thumb"]?.greaterThanOrEqualTo(1)) {
-    price = price.mul(1.1);
+  if (inventory["Sunflower Seed"]?.greaterThanOrEqualTo(1)) {
```
j923sneaks, line 10: I think we should retain `Green Thumb` here :)

src/features/crops/components/Plants.tsx
```diff
@@ -29,12 +29,17 @@ export const Plants: React.FC = () => {
...
+
+
   const sell = (amount = 1) => {
     gameService.send("item.sell", {
       item: selected.name,
       amount,
-    });
-    setToast({ content: "SFL +$" + selected.sellPrice.mul(amount).toString() });
+    })
+      {
```
j923sneaks, line 39: you can omit the square brackets

j923sneaks:
also @adamhannigan some were raising that the 10% bug persists when saving? shouldn't it be also handled in the backend?

jcads:
> After playing around with the variables for a while trying to find a format that is functioning,consistent with the code and the future code,

@Z1kou you can run `yarn format` to run the linter and format the code.

src/features/crops/components/Plants.tsx
Z1kou: It is because there is only one command in them right ?

adamhannigan (approved):
LGTM. Nice work again! 💪

PR #459, opened 2022-03-29
[FIX] Remove craft button after crafting 2 flags

jcads:
# Description

Removes craft button and ingredients after reaching flag crafting limit.

## Type of change

# How Has This Been Tested?

image

# Checklist:

### DIFF
--- a/src/features/tailor/components/TailorSale.tsx
+++ b/src/features/tailor/components/TailorSale.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useContext, useState } from "react";
 
 import close from "assets/icons/close.png";
 import flag from "assets/nfts/flags/sunflower_flag.gif";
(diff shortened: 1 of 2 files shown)


adamhannigan:
Good idea 💡  @jcads 

One concern around this approach is that we are introducing specific logic into a component that is used not only for flags. Taking this approach won't scale well and could introduce bugs in other areas that use this component.

I would suggest passing in a separate property into the `Rare` component - i.e. `canCraft`

This will enable us to lift the flag specific logic up into the parent component 😄

src/features/tailor/components/TailorSale.tsx
```diff
@@ -24,6 +32,10 @@ export const TailorSale: React.FC<Props> = ({ onClose }) => {
       return obj;
     }, {} as typeof FLAGS);
 
+  const maxFlags = Object.values(FLAGS).filter(
+    (flag) => flag.name in state.inventory
+  ).length === 2;
+
```
jcads, line 38: @adamhannigan ready for review. Moved this conditional to the parent component

adamhannigan (approved):
LGTM. Nice work! 🚀

PR #476, opened 2022-04-02
[FEAT] Skill upgrades toast notifications

jcads:
# Description

Shows a toast notification once skill upgrades are avaialable.

## Type of change

# How Has This Been Tested?
Manually set level and upgrades in `skillUpgradeToast.ts`

https://user-images.githubusercontent.com/50954098/161379048-8d75cbf5-4052-4082-9585-834d3800f628.mp4

https://user-images.githubusercontent.com/50954098/161379054-f50f6593-e823-4c31-92d5-11d37a9d9558.mp4

# Checklist:

### DIFF
--- a/src/features/blacksmith/components/Rare.tsx
+++ b/src/features/blacksmith/components/Rare.tsx
@@ -63,7 +63,12 @@ const Items: React.FC<{
     </div>
   );
 };
-export const Rare: React.FC<Props> = ({ onClose, items, hasAccess, canCraft = true }) => {
+export const Rare: React.FC<Props> = ({
+  onClose,
+  items,
+  hasAccess,
+  canCraft = true,
+}) => {
   const [selected, setSelected] = useState<Craftable>(Object.values(items)[0]);
   const { gameService } = useContext(Context);
   const [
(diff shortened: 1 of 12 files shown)


jcads:
**Update:** moved the function call to Game.tsx to show the notification after saving, not just when farming and gathering.

adamhannigan:
@jcads Nice initiative. 

One concern around this is that the logic gets called on every single game re-render instead of just the actions that would cause a skill upgrade.

We want to keep the Game.tsx component as lightweight as possible. Adding more functionality into that component leads to more critical code that needs to be maintained. Given that a user may only ever see this message once every couple weeks I suggest that we take an approach that is not coupled to every re-render.

Consider moving this logic into the `House.tsx` component. On the first render, we can check if a skill upgrade is available and display a call to action on the House (I.e. an exclamation mark). This won't show in real time, but the next time a user enters the game they will see the notification and open the house.

What do you think?

jcads:
@adamhannigan I initially did something like this on every harvest and every mining/chopping but it resulted in more code duplication. 

I think moving the logic to House.tsx is a better approach.

jcads:
image

adamhannigan (changes requested):
Awesome work. I manually tested it and it feels very natural!

It is also very clear to the player what they need to do 🤗 

Apologies for the delay, been extremely busy with launch 😅 I have left just one small suggestion to help improve render performance and reduce some bugs in the future

adamhannigan (approved):
LGTM 🚀 Apologies on the delay. Awesome work 👏

PR #490, opened 2022-04-06
[FEAT] Banking Withdraw Split Modal

adamhannigan:
# Description

This PR splits the withdraw functionality into a token and a items workflow. It makes the UX a bit clearer.

**Excludes**

UI tidying. It is still quite messy and will need to be addressed before we turn it on for the masses

Screen Shot 2022-04-06 at 1 57 01 pm


InuBakaBo:
Should put a little note to sync before you withdraw, and also after you withdraw. Or do you only need to sync once?

PR #516, opened 2022-04-10
[FEAT] Blacklist visited farms

adamhannigan:
# Description

This change ensures that when you visit a blacklisted farm, it will display as blacklisted


Z1kou:
Suggestion : adding rain effect when visiting blacklisted farms which is not necessarily needed but it will serve as a visual queue that the farm visited is blacklisted which a lot of people need 😅

Rofrtd:
I think @rickfurt was working on some sort of rain implementation!

rickfurt:
I think for blacklisted farms we could display something a bit more clear to show as blacklisted. I can work into some other visual that could be applied to. 

Like black clouds floating all over the farm or something else.

The rain effect was actually for the halving. Also I wanted to implement some sort of weather and day and night cycles for the game, although still having figured out a good way of doing it, but working on it.

Pranav-yadav:
> The black ominous clouds would be easier to implement, just need the assets

💯 Let's scare 'em 💀

adamhannigan:
Closing PR for a graphical + gameplay implementation

PR #539, opened 2022-04-13
[FEAT] Add Time Left Panel in Resources (On-Hover)

tiffanydys:
# Description
- Added `TimeLeftPanel` component
- Applied component to the following resources:
     - Trees
     - Stone
     - Iron
     - Gold 

## Type of change

# How Has This Been Tested?
`yarn test`
Screen Shot 2022-04-14 at 2 04 11 AM
`yarn dev`
Wood:
Screen Shot 2022-04-14 at 1 48 55 AM
Stone:
Screen Shot 2022-04-14 at 1 48 25 AM
Iron:
Screen Shot 2022-04-14 at 1 48 17 AM
Gold:
Screen Shot 2022-04-14 at 1 52 32 AM

# Checklist:


ahh1539:
Great change IMO. I think it's natural to add this if it's already done for crops.

adamhannigan:
@tiffanydys Great initiative 👏 

One concern is that every resource is going to be re-rendering every second. This could cause some unexpected lag especially while we scale.

I wonder if we could move the 1 second interval + update into the `TimeLeftPanel` itself. That way we only trigger updates and calculate time while the panel is actually visible.

1. User hovers tree
2. Panel is rendered and begins the interval
3. User "unhovers"
4. Panel removes the interval

What are your thoughts?

j923sneaks:
Like in `Soil.tsx` Can we not also pass the existing `timeLeft` prop to the Panel component? 

If the re-rendering is the issue then maybe the timers themselves need to be reworked. the computation for `timeLeft` and `percentage` always update anyways

tiffanydys:
> @tiffanydys Great initiative 👏 
> 
> 
> 
> One concern is that every resource is going to be re-rendering every second. This could cause some unexpected lag especially while we scale.
> 
> 
> 
> I wonder if we could move the 1 second interval + update into the `TimeLeftPanel` itself. That way we only trigger updates and calculate time while the panel is actually visible.
> 
> 
> 
> 1. User hovers tree
> 
> 2. Panel is rendered and begins the interval
> 
> 3. User "unhovers"
> 
> 4. Panel removes the interval
> 
> 
> 
> What are your thoughts?

Yup agreed. Might also want to detect if timeLeft is less than 60 minutes so that we only start the interval timer when it displays the seconds left on the Panel.

tiffanydys:
> Like in `Soil.tsx` Can we not also pass the existing `timeLeft` prop to the Panel component?
> 
> If the re-rendering is the issue then maybe the timers themselves need to be reworked. the computation for `timeLeft` and `percentage` always update anyways

Yeah I think the timers do need a rework. Might probably work on that in another PR though, or if anybody wants to get it ✌🏻😄

adamhannigan:
@tiffanydys Code looks good. I just need to test it on a few different devices. For transparency, it may take me a couple days before I can merge this in <3

tiffanydys:
> @tiffanydys Code looks good. I just need to test it on a few different devices. For transparency, it may take me a couple days before I can merge this in ❤️ Apologies for the delay

Hi adam 👋 
No worries, take your time 😉

adamhannigan (approved):
Awesome work @tiffanydys 

This is a great UX improvement 🤗

PR #553, opened 2022-04-15
[CHORE] Remove underlines from UI

adamhannigan:
# Description

Remove underline since it is not a link


aayusharyan:
Ohhh, I was also about to make this change. 😅

Issue #592, opened 2022-04-18
[PROPOSAL]  DECORATION RANK


SrCouve:
I think the concept of decorative NFTS is cool, but I think it would be interesting to add a decoration rank, where each NFT would give a score (Based on its scarcity and  mint price) and depending on your rank position you would earn a certain amount of SFL (Daily , weekly or monthly) to avoid abusing this system when you wanted your nft to score in that rank you would choose a "hold" option where it would only be withdrawn after the rank score deadline, I think it would be a system cool to encourage the purchase of cosmetic Nfts and gives functionality to them, even Nfts with utilities could enter this rank 

I believe it would be a very interesting mechanic to bring competitiveness to the game, we could also add a new NPC in the city, which would be where we would see the farms of the rank and we could visit them directly by clicking on the rank,
in addition to stimulating trade and the creation of decorative NFTS, it would also stimulate interaction between players, making it more meaningful to visit other farms

adamhannigan:
Some great ideas. Let's continue the chat over in #607

PR #607, opened 2022-04-19
[FIX] Open Sea Bunny Decimal Point Values

adamhannigan:
# Description

When a value has 0 decimal points we need to specify 0, instead of 1.

Using 1 signifies that 1 token would equal = 0.1


RogGarcia:
Hi, I'm not very used to the way github works (I came here because of a comment from Adam), but I would like to mention two proposals regarding decorative items and their extended function in the game mechanics. I hope they contribute to a more engaged gameplay:
https://github.com/sunflower-land/sunflower-land/issues/609
https://github.com/sunflower-land/sunflower-land/issues/592

PR #615, opened 2022-04-20
[FEAT] Add crop ready sound

brian-man:
# Description

This change adds a sound emission when crops are reaching maturation for harvesting.

## Type of change

# How Has This Been Tested?

1. Plant a single crop and verify the sounds plays at maturation
2. Plant several crops at once to verify the sound plays multiple times

# Checklist:


brian-man:
@adamhannigan 
The ready.mp3 is a placeholder (it's a copy of frog.mp3).  Please don't merge until the desired sound has been crafted and the ready.mp3 has been updated.

src/features/crops/components/Soil.tsx
```diff
@@ -73,6 +75,9 @@ export const Soil: React.FC<Props> = ({
   if (timeLeft > 0) {
     const percentage = 100 - (timeLeft / crop.harvestSeconds) * 100;
     const isAlmostReady = percentage >= 50;
+    if (timeLeft < 2) {
```
Rofrtd, line 78: Also here, the sound would play twice as it goes to 1 and 0..

Rofrtd (commented):
Hi @brian-man, 

This implementation would play a sound for each crop ready! 
I feel like this would be a bit overwhelming when all your 22 crops are ready to harvest and the sfx plays 22 times on top of each other. 

Any ideas on how we could solve this? 

I think something like, if there are other crops almost ready to harvest (say withing 30 seconds with just a couple of seconds in between each other) we wait for the last one to trigger the sfx.

src/features/crops/components/Soil.tsx
newb23: I think what might work best is a sound plays only on the first crop done, then every X seconds after that. Bonus points for X being user configured. Triple bonus points for user configurable ready sounds. (Chooser for in game sounds, or, user"upload" to local cache?)

Rofrtd: Yeah, good idea for it playing on the first crop ready.

We just need to take into considerations players can plant different crops at different times.
Having crops ready with a couple minutes in between.

I think the time being configure by player could be done based on the player feedback for this first implementation.

And the uploading sounds I don't think is ideal, this would be unnecessary complexity and also we want the sfx to align with the other sfx in game.

brian-man: > Also here, the sound would play twice as it goes to 1 and 0..
I originally had "< 1" and it never fired.  With "< 2" it fired reliably and sounded good to me.

brian-man:
> This implementation would play a sound for each crop ready! I feel like this would be a bit overwhelming when all your 22 crops are ready to harvest and the sfx plays 22 times on top of each other.

That was the intent.  The idea was to [a] effectively amplify and/or prolong the sound if more crops are ready and [b] properly handle crops arriving in a staggered fashion aligned with how they are planted.

Choosing the right sound is important (meaning you don't want a 15sec sound for example).

src/features/crops/components/Soil.tsx
brian-man: > I think what might work best is a sound plays only on the first crop done

I don't know what this means when crops arrive 1 sec, 5 sec, 10 sec, and 5 minutes apart.

brian-man:
I added specific replies to the above.

Broadly, assuming I can figure out how to implement it I'm happy to do whatever the team prefers.  Any auditory hint of "crop ready" is an improvement for my farm. :)

ahh1539:
Would love to see this implemented in some capacity but also not too certain how to go about it 🤔

ahh1539:
Or we could simply just also have a timer for the `readyNoiseRecentlyPlayed` like we do for the crops now where we set it for 5 minutes each time after a crop plays the noise

0xsha1man:
A couple of things that should come into consideration

What is the intent of the sound?
- give audio feedback on the harvest?
- just a nice polish?

Way back in the day a game IP I worked on was licensed out to Bally's as a slot machine interactive. As the largest and probably still the only producers of slot machines, they knew from their own inside research what years later was officially studied by the NIH. Thus we had to spend a lot of time revamping our game audio cues to meet certain standards of theirs.

> [The Impact of Sound in Modern Multiline Video Slot Machine Play](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4225056/)
The results showed that the sound influenced the arousal of participants both psychophysically and psychologically. The sound also influenced players’ preferences, with the majority of players preferring to play slot machines that were accompanied by winning sounds.

Basically, there's a reason that coin dropping into a tin tray exists to this date when slot machines no longer accept coins. Also why it is sooooo loud in casinos.

Now moral questioning reasons aside... 

-**If the sound is to drive player intent and give positive enforcement and need to keep coming back for more...** 
Then yes after every harvest. Just please make sure that sound doesn't blow out the player's ears with the rapid firing.

-**If the intent is an audio cue, keep in mind audio assistive tech such as amplified audio or CC**
Then play the audio cue in a progressive sequence with diminishing returns. You can't ignore effective research to drive the longevity of the product. HOWEVER, to prevent the immediate reaction that rapid-fire sound can cause of muting ALL audio, that audio sequence is best. 

Let me try to explain...

> When audio cues of the same set trigger to play within X ms (say twice within 1/4 of a second), reduce the audio level of the consecutive play, repeat diminishing the sound even lower until after the 3rd or 4th request then ignore all others.
> 
> If no audio of that set is playing or there was a pause in the requests for it after 1 or 2 seconds (imagine clearing a field, player pans to the next section and begins again), then immediately start the original sequence again at 100%, reducing levels like normal.

-100%, 85%, 60%, 30%, ignore other requests
-a pause of 1 or 2 seconds or no audio cue of that set is playing, start over at 100%

The numbers would have to be adjusted with testing maybe stop after the third, or the % doesn't feel right etc.

Having not heard the audio, its length, plus just observing it in action in regards to the game those numbers really depend on this situation once it is implemented. Also, keep in mind the reward situation that happens with a harvest that can occur. That sequence would still work just fine and actually would be great because of that.

Z1kou:
@0xsha1man  hey bro i read all of you text, and i totally agree Looking at the purpose for the sound, 

**Which in our case is to serve as an alert for players that their crop are ready**        I suggest using your second format of having a decreasing system, or How about add a  1 cue for all the crops that has a 30 second cooldown so people don't get spawned, you planted 22 sunflowers? when the first one is ready you will get the sound effect,


This way we ensure : - No spam , -Purpose fullfilled , -Less complicated code writing since you wouldn't need to design a function with decreasing volume ect...


What do you guys think?

0xsha1man:
That wouldn't work quite with my own playstyle. My crop organization is haphazard at best.  So having the cue keyed to each plant growth duration wouldn't work for me. If that is the going idea then just the third option would work best.  Set the audio to be a set starting value and vary it +- a certain percent each play.  Varying the volume like that will break the perception of it and be enough of a mix to not get too annoying.

Example: Base volume is 80 %.
One sound cue, randomly select a number between 0-20%.  Add it to the base and play the cue at that volume. to get a final value somewhere between 80 and 100%.

Doing it that way simplifies it and also allows us to restrict certain ones in the future or in general from playing too loud.  Set the blacksmith on repeated open closing sounds to not escalate in loudness and annoyance as an example.

Rofrtd:
@0xsha1man nice write up.
But I believe we don't need to "spam" the sound on every crop ready even with the volume variations.
I believe it would be better to lets say:
- you have a batch of crops ready the sound plays once even if there are multiple crops ready within the next minute.
- The sound would play again if the time in between crops ready is longer than 3 min (just an example)

adamhannigan:
Closing for inactivity

brian-man:
@adamhannigan The inactivity is because I need core dev input on which road they want to go down. The initial PR was very simple and a cavalcade of ideas basically made it required for core devs to pick a path.

PR #632, opened 2022-04-21
[FIX] Floor SFL balance to 18 decimal places after selling

brian-man:
# Description

When acquiring the minimal amount of Wheat (meaning 1e-18 units) via trading, deposit works fine.  What does not work fine is selling the Wheat.  If you sell the Wheat (with or without NFTs), the sale price is 7e-19.  The Shop vendor happily does the transaction.  From there forward, Sync is broken for the farm for all attempts going forward.

The problem is that the in-game SFL balance can store more than 18 decimal places but the on-chain token support 18 so a Sync is impossible.

This change floors the SFL balance to 18 decimal places after completing a sell operation.  This prevents this issue in this case and similar cases that are likely to happen after a few halvings.

I'm hopeful it will also fix my currently un-sync-able farm in Polygon mainnet. :|

Screenshot of the Wheat about to be sold:
image

Screenshot of the message after sale:
image
The 20th and 21st decimal places are due to 5% from the Green Thumb skill.

## Type of change

# How Has This Been Tested?

Validated locally with 1 and 2 as the floor values.  I don't have a good way to test the Sync scenario though as it's not available on local.

# Checklist:

### DIFF
--- a/src/features/game/events/sell.ts
+++ b/src/features/game/events/sell.ts
@@ -38,7 +38,7 @@ export function sell({ state, action }: Options): GameState {
 
   return {
     ...state,
-    balance: state.balance.add(price.mul(action.amount)),
+    balance: state.balance.add(price.mul(action.amount)).toDecimalPlaces(18, Decimal.ROUND_DOWN),
     inventory: {
       ...state.inventory,
       [crop.name]: cropCount.sub(1 * action.amount),


adamhannigan (approved):
LGTM. Let's investigate how we can set this precision at a global level in the app.

Ideally all of our calculations should follow the 18 point decimal precision.

PR #673, opened 2022-04-24
[FIX] Decimal handling of selling all crops

### LINKED ISSUE #637: [BUG] - Change the decimal handling of the Game frontend to match with the Contract.
The contract has about 10 decimal places, (give or take) but the game UI can only handle 1. Because of which the game tries to round the number and tries to sell, resulting in insufficient crops error.

Something like this.
image

Now, when I try to sell it, It tries to 1843.4 which obviously cannot sell because it is more than what I have. I cannot perform sell all operation.
image

Fix -> Instead of rounding. round down to the nearest decimal and then try to sell.

Priority -> Possibility before the halving.

pnkthrepwood:
# Description

Fixes #637 

Casting a "Decimal.js" number to a "number" javascript primitive causes a floating point precision loss that, in some cases, can prevent crops to be sold ( see example in #637 )

Fixed by adding a global precision setting and using `Decimal` instead of `number`.

## Type of change

# How Has This Been Tested?

Manually tested the scenario proposed in #637

1. Change `cropCount` of `harvest.ts` to the testing number `new Decimal('1843.399999999999999999')`
2. Start a new farm
3. Harvest a crop
4. Open market
5. Navigate to 'Sell' tab
6. Select the crop type harvested in (3)
7. Press 'Sell all' button
8. Confirm the modal
9. Observe that the bug is fixed and crops can be successfully sold

# Checklist:

### DIFF
--- a/src/features/game/Game.tsx
+++ b/src/features/game/Game.tsx
@@ -37,6 +37,10 @@ import { screenTracker } from "lib/utils/screen";
 import { Withdrawn } from "./components/Withdrawn";
 import { EasterEggHunt } from "features/easter/Area";
 import { Resetting } from "features/auth/components/Resetting";
+import Decimal from 'decimal.js-light';
+
+const DECIMAL_PRECISION = 78;
+Decimal.set({ precision: DECIMAL_PRECISION });
 
 const AUTO_SAVE_INTERVAL = 1000 * 30; // autosave every 30 seconds
 const SHOW_MODAL: Record<StateValues, boolean> = {
--- a/src/features/crops/components/Plants.tsx
+++ b/src/features/crops/components/Plants.tsx
@@ -31,7 +31,7 @@ export const Plants: React.FC = () => {
 
   const inventory = state.inventory;
 
-  const sell = (amount = 1) => {
+  const sell = (amount = new Decimal(1)) => {
     gameService.send("item.sell", {
       item: selected.name,
       amount,
@@ -46,11 +46,11 @@ export const Plants: React.FC = () => {
   const displaySellPrice = (crop: Crop) => getSellPrice(crop, inventory);
 
   const handleSellOne = () => {
-    sell(1);
+    sell(new Decimal(1));
   };
 
   const handleSellAll = () => {
-    sell(cropAmount.toNumber());
+    sell(cropAmount);
     showSellAllModal(false);
   };
 
@@ -129,7 +129,7 @@ export const Plants: React.FC = () => {
               sell all your {selected.name}?
             </span>
             <span className="text-sm text-center text-shadow mt-1">
-              Total: {cropAmount.toNumber()}
+              Total: {cropAmount.toString()}
             </span>
           </div>
           <div className="flex justify-content-around p-1">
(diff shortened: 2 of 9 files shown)


src/features/game/Game.tsx
```diff
@@ -37,6 +37,10 @@ import { screenTracker } from "lib/utils/screen";
 import { Withdrawn } from "./components/Withdrawn";
 import { EasterEggHunt } from "features/easter/Area";
 import { Resetting } from "features/auth/components/Resetting";
+import Decimal from 'decimal.js-light';
+
+const DECIMAL_PRECISION = 78;
```
adamhannigan, line 42: What is the thought process behind 78?

Could we use 18 decimal points to support ERC20 & other smart contract precision?

adamhannigan:
We will need to support this on the server as well.

When this payload gets sent to the API, the Decimal will be stringified and will lose the precision it would of previously had once it gets converted back to a number.

I wonder if we support a extra type `all` which can be used specifically when a user wants to sell all.

`amount: number | 'all'`

src/features/game/Game.tsx
pnkthrepwood: According to the docs, the [precision ](https://mikemcl.github.io/decimal.js/#precision)setting in decimal.js is actually used to store the number of significant digits.

In the bug, the number is "1438.3999...99":   4 digits before point and 18 decimals --> 22 significant digits

ERC20 tokens usually have 18 decimals places, but it's up to the contract to use any number arbitrarily.

As smart contracts only use uint256, the maximum value that can be represented (2^256 - 1) can have at most 78 significant digits.

Although 78 is the most extreme case, I would say that 18 and "some more" would be enough in practice.

pnkthrepwood:
We can use `all` to send that action instead of the stringified number. although the bug is not limited to the "sell all" action.

For example, if a farm has a very large quantity of tokens and the seed price is small enough, the farm could buy seeds without spending token.

bug_

This example is not feasible right now (there even isn't such large quantity of SFL token minted), but as more items are introduced and more halvings keep happening, a similar bug could arise in the long term.

src/features/game/Game.tsx
adamhannigan: Good point, thanks!

adamhannigan:
Screen Shot 2022-04-28 at 9 09 10 am

@pnkthrepwood I get this error when trying to run locally.

**Steps to reproduce**

1. Log in to farm
2. Buy seeds
3. Sell crops
4. Click save.

Any idea what is causing this?

pnkthrepwood:
I guess this is coming from `autosave.ts`, in squashEvents. 
I can't reproduce it though. Maybe only happens with testnet?

adamhannigan:
@pnkthrepwood Have we set you up with testnet access yet? 🤗 What is your discord name?

pnkthrepwood:
I don't have access yet, my Discord is pnkthrepwood#6838

adamhannigan:
@pnkthrepwood We have fixed this issue and is being handle server side + contract side

PR #681, opened 2022-04-24
[FEAT] Show current release version tag on splash screen

ShiningMinions:
# Description

As a player, I would really like to know the client version I'm currently playing on, so that I don't have to go in-game to know if my browser has loaded the most latest features. It would also be nice if this version was "clickable", and brought me to SF Land's releases page where I could read more.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

**PLEASE NOTE! This change requires some testing with github actions. This is needed to correctly calculate the latest release tag for new .env variable. I highly recommend we take a close look at this new github action before merge.**

Since this is a UI feature, I tested visually. I made sure that this element only appears on the Splash screen by putting it within the "Splash" component. I also ran the game in "view-only" mode to ensure this new element doesn't carry-over to appear in-game.

For local development, the text will say "Version: local-dev" (or whatever value is set in .env file). However, the end goal is that github action will pickup on the latest tag and correctly set this new .env variable for each new release (i.e. v0.2.10-end-hunt).

The new "Version" element:
splash-version-panel

# Checklist:

- [x] Title of the PR is relevant and is prefixed with [FEAT], [CHORE] or [FIX]
- [x] Screenshot if it includes any UI changes
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #715, opened 2022-04-27
[FEAT] Move logout button from settings into main menu dropdown.

ShiningMinions:
# Description

As a player, it feels a bit awkward and unnatural to me needing to go into the nested "settings" sub-menu in order to logout. I moved the "Logout" button from the "settings" sub-menu and into the bottom of the main menu.

## PLEASE NOTE:

This is an opinionated change - some players in dev-chat disagree, while others definitely agree. To voice your opinions, please do the following:

👍 if you agree with this change
👎 if you think logout button should stay where it's currently at (in nested "settings" sub-menu)
💬 Alternatively, please comment if you think the change is good but would like something tweaked.

## Type of change

# How Has This Been Tested?

**At time of writing, I do not have a test-net farm so I cannot test this. I will work on getting this, but I'm asking if another community member or reviewer can pull my branch and test this. All that needs to be tested is if the "Logout" button still works. I'm fairly confident that the code works because it was a simple cut-and-paste, but as a developer over the years I find it unwise to assume this. 😉**

PC View:
logout-button-pc

Mobile View:
logout-button-mobile

Notice Logout Button gone in "Settings" sub-menu:
logout-button-gone-settings

# Checklist:

### DIFF
--- a/src/features/hud/components/Menu.tsx
+++ b/src/features/hud/components/Menu.tsx
@@ -111,6 +111,11 @@ export const Menu = () => {
     authService.send("EXPLORE");
   };
 
+  // {Todo: Modify gameState and authMachine for Logout event}
+  const onLogout = () => {
+    authService.send("LOGOUT"); // hack used to avoid redundancy
+  };
+
   // Handles closing the menu if someone clicks outside
   useEffect(() => {
     document.addEventListener("mousedown", handleClick);
@@ -220,6 +225,14 @@ export const Menu = () => {
                     onClick={handleSettingsClick}
                   >
                     <span className="sm:text-sm flex-1">Settings</span>
+                  </Button>
+                </li>
+                <li className="p-1">
+                  <Button
+                    className="flex justify-between"
+                    onClick={onLogout}
+                  >
+                    <span className="sm:text-sm flex-1">Logout</span>
                   </Button>
                 </li>
               </>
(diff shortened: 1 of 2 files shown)


baron:
I support this for two reasons:

- most users expect a logout button on the first level of menu options
- the current "Settings" button is a bit of a misnomer as it includes no configurable user settings (Logout, Reset Session, Airdrop V1 Farm which is a temporary addition)

Also, I liked the red background version (again mostly from conditioned expectations)

adamhannigan:
@ShiningMinions Looks like there has been a bit of a discussion on Discord around this.

I'm leaning towards keeping it in the Settings. A few reasons:

1. The current Menu is too bloated
2. The Settings -> Logout approach is common for apps where users don't log out often. I.e. 1 device = 1 user.
3. It is not a common action - Only multiaccounters would be utilising this. Multiple on the same device need to be using separate MetaMask accounts anyway

ShiningMinions:
@adamhannigan yup and you definitely make some great points. This is 100% a PR I think we can take or leave, I just wanted to offer it up.

I think the thing itching at the back of my brain, is that I'm kind of half-expecting some more content to go on the splash page soon (i.e. announcements, FAQs, etc.), which gives me a reason as a player wanting to logout more often. And in fact, I saw my version PR just get merged in (cheers btw!): https://github.com/sunflower-land/sunflower-land/pull/681. As a player, I would be more inclined to logout to see info like this.

Again, I can take or leave it, just wanted to offer it to the community. Cheers!

17ss:
As a regular player, I open my browser when I want to play, and close it when I don't want to play, the logout function doesn't do anything. But it makes easy to use robots

adamhannigan:
@ShiningMinions Definitely will have some use cases for this.

For now let's park this PR until we have these features in the game and logging out becomes a more useful action for the player 🤗

PR #773, opened 2022-05-02
[FIX] token/inventory is null without API URL

Sliman4:
# Description

When running locally without API_URL, it displays an error in console and won't open some modal windows:

image
image
image
image

After change:
image
image
image

Since there's no reasonable value to put into SFL Supply, I've added a placeholder that says "You're running Sunflower Land locally!"
NFT interfaces are read-only and obviously you can't craft, but it may be useful if a new contributor like me wants to change something in UI.

## Type of change

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Manually and yarn test.

# Checklist:


adamhannigan (approved):
LGTM 🚀 

In the future we could consider adding a more explicit URL such as `IS_LOCAL`.

Ping us on Discord #coders-chat and we can set you up with testnet access 😄

PR #782, opened 2022-05-03
[FEAT] Adds toast message on {item} collection

newb23:
# Description

Adds a toast message on each {item} collection showing the amount of the item collected.

## Type of change

# How Has This Been Tested?

Test Suites: 12 passed, 12 total
Tests:       103 passed, 103 total
Snapshots:   0 total
Time:        24.951 s
Ran all test suites.
Done in 26.27s.

Local testing through both `yarn test` and local play.

image
image
# Checklist:


adamhannigan:
@newb23 What is the main goal of also showing the toast? Is the current amount that shows not clear enough to the user?

We want avoid too many animations/popups happening as it can get quite distracting for the user

newb23:
Hey @adamhannigan,

Yes, particularly when harvesting and re-planting using a "double-click", the popover text is obfuscated behind/in front of/around the other elements there, making it difficult to see what was actually gained and/or planted.

If the design intent is to limit these notifications, may I instead offer two other pieces of feedback?

1, offer it as a setting for the user? Three states, of: toast only, popover only, or both, and have this setting saved in localStarage as a key.

2, and maybe this regardless of 1, move the popover text further away from the "action", or otherwise make it more apparent? (Green color for the +X string?) Another option I think might work is making it "float" vertically higher then fade out, possibly also with a color change.

Obviously there are a lot of design considerations here, lol. I just wanted to use something both already built into the game as well as unobtrusive as possible, and this was both pleasant and informative, in my eyes.

adamhannigan:
You raise some good points @newb23 It sounds like we are uncovering a root problem.

If the player can't see those values that pop up, we should improve the UI around that area 🙌

spencerdezartsmith:
Great job! The main issue with this toast is on mobile. When you do a lot of actions fast this popover takes up a large part of the screen and stops you from continuing until it goes away. Like @adamhannigan I think we can improve the UI that currently exists.

newb23:
> Great job! The main issue with this toast is on mobile. When you do a lot of actions fast this popover takes up a large part of the screen and stops you from continuing until it goes away. Like @adamhannigan I think we can improve the UI that currently exists.

I completely get this. An alternative, or really, more fuel to the conversation in addition to the "vertically floating numbers" idea I had above, is that I can definitely limit the number of toast messages generated, I just didn't want to chu n too much work in the event the idea itself was rejected, as it partially has!

adamhannigan:
@newb23 Looks like there is still a bit of discussion around whether to add this to the crops or not.

I would suggest for now omitting that part from the PR and we can add the minerals + wood toasts to see how it feels and player's responses 😄

newb23:
> @newb23 Looks like there is still a bit of discussion around whether to add this to the crops or not.
> 
> I would suggest for now omitting that part from the PR and we can add the minerals + wood toasts to see how it feels and player's responses 😄 

@adamhannigan 
Absolutely, I will have this PR'd later this morning. There are any number of combinations of options I can do with this guy, and more than happy to test, try, and retry as needed.

newb23:
@adamhannigan @spencerdezartsmith
Just tagging to let you know this is ready for "player testing" with all comments to date, addressed.

spencerdezartsmith:
Screen Shot 2022-05-16 at 3 32 50 pm

Just tested this and its working well for me. Do you think it makes sense that we use the word for the resource and also the image? I think we should probably be consistent with this and use one or the other. Implementing the word would probably be the easiest here but the icons are probably more accessible given that not all players speak english. Thoughts?

newb23:
@spencerdezartsmith 

I think that your thoughts on both sides are correct. I wonder if we could do both without crowding too much. Something like {resourceName} ({resourcePic}). I would like to try this this, and see how she goes, if it looks gross, I think the universal {resourcePic} is going to be best across the board.

Further, in another PR, I want to attempt to "stack" the amounts gained/spent after a certain maximum. Meaning: once you get to 5 buy/sell "events", add/subtract the next event's total to the 5th element (pun!), so the window doesn't grow so large.

spencerdezartsmith:
> @spencerdezartsmith
> 
> I think that your thoughts on both sides are correct. I wonder if we could do both without crowding too much. Something like {resourceName} ({resourcePic}). I would like to try this this, and see how she goes, if it looks gross, I think the universal {resourcePic} is going to be best across the board.
> 
> Further, in another PR, I want to attempt to "stack" the amounts gained/spent after a certain maximum. Meaning: once you get to 5 buy/sell "events", add/subtract the next event's total to the 5th element (pun!), so the window doesn't grow so large.

Ok sounds like a plan. Please make sure that you test your solutions a lot on mobile screens. This is where we need to be most mindful of crowding. I look forward to seeing what you come up with!

newb23:
@spencerdezartsmith 
This is now ready for your testing with the change to icons in the toast messages. Once/if approved, and people like it, I will work towards another PR addressing "stacking" the toast messages to better utilize space in the event of many separate purchases/crafts/etc, as discussed. (Pic added for crafting an axe. Cheated on the cost though. 😄)

adamhannigan (approved):
LGTM. Nice work!

PR #801, opened 2022-05-05
[FEAT] Progress bars now display 1 decimal place

brian-man:
# Description

Concerns:
1. The progress bar timers are "weird" around the value of 1.  The "1min" progress is displayed for exactly 1 second.
2. When upgrading with Nancy, the alignment with integer units drops so Sunflower duration is "51s" and similarly for other crops.
3. For longer duration items like without-Woody trees, "1.4hrs" is more useful than showing "2hrs" for a full hour.

This change introduces 1 decimal place to resource progress bars.

For example, Potato's initial progress bar with Nancy is now "4.3mins" instead of "5min"

Crop examples:
image

Wood examples:
image
image

## Type of change

# How Has This Been Tested?

Watching the timers for various crops to ensure that the values look correct and the transitions across time units behavior properly.

# Checklist:

### DIFF
--- a/src/lib/utils/time.test.ts
+++ b/src/lib/utils/time.test.ts
@@ -17,13 +17,21 @@ describe("time", () => {
       expect(secondsToString(2 * ONE_SEC)).toBe("2secs");
       expect(secondsToString(59 * ONE_SEC)).toBe("59secs");
       expect(secondsToString(ONE_MIN)).toBe("1min");
-      expect(secondsToString(ONE_MIN + ONE_SEC)).toBe("2mins");
+      expect(secondsToString(ONE_MIN + ONE_SEC)).toBe("1.1mins");
+      expect(secondsToString(2 * ONE_MIN - 6 * ONE_SEC)).toBe("1.9mins");
+      expect(secondsToString(2 * ONE_MIN - ONE_SEC)).toBe("2mins");
+      expect(secondsToString(2 * ONE_MIN)).toBe("2mins");
       expect(secondsToString(59 * ONE_MIN)).toBe("59mins");
       expect(secondsToString(ONE_HR)).toBe("1hr");
-      expect(secondsToString(ONE_HR + ONE_SEC)).toBe("2hrs");
+      expect(secondsToString(ONE_HR + ONE_SEC)).toBe("1.1hrs");
+      expect(secondsToString(2 * ONE_HR - 6 * ONE_MIN)).toBe("1.9hrs");
+      expect(secondsToString(2 * ONE_HR - ONE_SEC)).toBe("2hrs");
+      expect(secondsToString(2 * ONE_HR)).toBe("2hrs");
       expect(secondsToString(23 * ONE_HR)).toBe("23hrs");
       expect(secondsToString(ONE_DAY)).toBe("1day");
-      expect(secondsToString(ONE_DAY + ONE_SEC)).toBe("2days");
+      expect(secondsToString(ONE_DAY + ONE_SEC)).toBe("1.1days");
+      expect(secondsToString(2 * ONE_DAY - 3 * ONE_HR)).toBe("1.9days");
+      expect(secondsToString(2 * ONE_DAY)).toBe("2days");
     });
   });
 
--- a/src/lib/utils/time.ts
+++ b/src/lib/utils/time.ts
@@ -31,16 +31,16 @@ export function secondsToString(seconds: number) {
   }
 
   if (seconds < ONE_HR) {
-    const minutesCeil = Math.ceil(seconds / ONE_MIN);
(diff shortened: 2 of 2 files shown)


adamhannigan:
@brian-man Using decimal places for time is often quite confusing for users. From building a timesheets platform, a lot of people got confused with `1.3` = 1 minute 30 seconds for example.

In this case here, consider showing it as `1min 30secs` -> `1min 49secs`.

This however may be too long to display on the plant. This is likely why the decimal point precision was not implemented in the first place

brian-man:
@adamhannigan 
Anybody that can't handle/understand 1 decimal place will already be unable to grasp the pricing for the first thing in the game: Sunflower Seed price.  That said...

"1hr 1min" tooltip displayed as "2hr" in the progress bar UI is just awful.

Recommendations?

MarketParticipant:
1h 1m

tiffanydys:
> @brian-man Using decimal places for time is often quite confusing for users. From building a timesheets platform, a lot of people got confused with `1.3` = 1 minute 30 seconds for example.
> 
> In this case here, consider showing it as `1min 30secs` -> `1min 49secs`.
> 
> This however may be too long to display on the plant. This is likely why the decimal point precision was not implemented in the first place

I agree with adam here, seeing the time as `1.3` will make people think it's 1 min 30 sec, and for the people who actually get that it's decimal placement then they're still going to divide the time in their heads and calculate how much time left is there. 

While I agree that showing "2hr" when it's below 2 hours is not that great, it's better and easier for people to see rather than a decimal placed time value.

Also people can just hover on the crop/resource to view the exact time itself - that's the purpose of why it was implemented.

JCtapuk:
Replace the time by a percentage from 0 to 100% and when pointing will show the remaining time

brian-man:
> Also people can just hover on the crop/resource to view the exact time itself - that's the purpose of why it was implemented.
As I stated in discord, the hover UI is nearly useless for me.  The static UI is what matters and the...

...2hr, 1hr for exactly 1 sec, 59min...

...approach currently used is lacking.

Also, as I stated above: If people can't manage that "1.3 hr" means "1.3 * 60 minutes" then they'll basically be a lost cause when trying to process that Sunflower Seed costs 0.0005 units of currency.

brian-man:
> Replace the time by a percentage from 0 to 100% and when pointing will show the remaining time

This would be useless and redundant, as the progress bar art already provides the % completion.

The value (for me at least) of the time above the progress bar is that I can see immediately when I plant and when glancing over later at the game from the other screen how much time I have left after showering, or getting lunch, or whatever.  Hovering over mixed crops to find this information is simply awful UX.

tiffanydys:
> 
> Also, as I stated above: If people can't manage that "1.3 hr" means "1.3 * 60 minutes" then they'll basically be a lost cause when trying to process that Sunflower Seed costs 0.0005 units of currency.
> 

We need to remember that we are building the game in the point of view of a person who is "dumb" - implementing decimal placed time will make it harder for them
and it would look like we aren't very inclusive with our updates.

brian-man:
> We need to remember that we are building the game in the point of view of a person who is "dumb" - implementing decimal placed time will make it harder for them and it would look like we aren't very inclusive with our updates.

To be clear: I wasn't the one suggesting people would be confused by "1.1min" despite having to understand "0.0001" priced seeds.  Moving on ...

> How about we show `1 hour 3 mins` to something like `1h3m` and will only show 2 units of time at a time.
> For example:
> 1 day 3 hours -> `1d3h`
> 1 hour 45 mins -> `1h45m`
> 5 mins 30 sec -> `5m30s`
> 30 sec -> `30s`

I think that's harder to read than the PR, but it's better than what is in main.  Improvement is good.

@adamhannigan Your thoughts?

newb23:
$.02, Personally, I like the thought of 
> 1 day 3 hours -> `1d3h`
> 1 hour 45 mins -> `1h45m`
> 5 mins 30 sec -> `5m30s`
> 30 sec -> `30s`

In place of the current implementation.

adamhannigan:
@brian-man 30s looks good, but the others look a bit crowded. Could we instead just show 1 minute for longer such as 60seconds - 90seconds.

I don't think the issue is seeing plants at 5 minutes or 6 hours. The main issue we need to solve is around the 1 minute display.

I think in this case we could solve just that issue and it would be a significant improvement

brian-man:
> I don't think the issue is seeing plants at 5 minutes or 6 hours. The main issue we need to solve is around the 1 minute display.

@adamhannigan Well, the reason I created the PR in the first place is because "2hr" for 61 minutes was annoying the hell out of me.

adamhannigan:
❤️  Great initiative but closing for inactivity. The proposed solution doesn't seem to have community consensus at this point

PR #827, opened 2022-05-07
[FIX] [HIGH PRIORITY]: Hide Goblin Village "Entry" & btn in Menu, in Visiting State 

Pranav-yadav:
# Description

- Added temporary readonly state and transitions to forbid entry to Goblin Village while in visit - readonly state.
- Hide's "Goblin Village Entry" from Humans Town as well as Button to go to goblin town from "Menu" while in Visiting (readonly) state.
- a lot more... 😉 

## Fix Scope: Vulnerability

Fixes Vulnerability Reported through Discord DM by _me_  😉 (`@pranavyadav#9917`)  AKA `@Pranav | (127.0.0.1)`

## Includes
- [x] added _temp_, readonly state and transitions to `goblinMachine`
- [x] show Forbidden Screen for `/goblins` when in visit - readonly state
- [x] added check to `/goblins` route forbid''s when in visiting state
- [x] added fallback for above cond, shows Forbidden screen instead of Blank.
- [x] When in Visiting (readonly) state; 1. hide Goblin Village Entry
- [x] 2. hide Go to Goblin Village button from Menu (under Map)
- [x] Added Forbidden R.FC to forbid entry to Goblin Village when visiting, i.e., in readonly state
image

## Type of change

# How Has This Been Tested?

- Manually
- `yarn test`
- `yarn tsc`

# Checklist:

- [x] I have added screenshots for UI changes/ Fixes
- [x] I have properly commented the code wherever necessary


Pranav-yadav:
@adamhannigan @spencerdezartsmith 
I have _**tried**_ to fix both UI as well as Code/Logic scopes of the vulnerability reported by **_me_** through Discord DM..

If possible merge this before `#827` - as I won't get time afterwards to resolve conflicts if any...

Side Note:
- the `goblinMachine` readonly states are temp fix, do refactor that state as well transitions for the same after merging.

adamhannigan (commented):
Nice work mate! Hiding the Goblin entry from the UI is a great step forward.

I would suggest going one step further and never letting a visiting user access the Goblin page at all. This is a common React auth pattern that we can use.

E.g.

```
          <Routes>
            <Route path="/" element={<Humans />} />
            {
              // Goblin town should never accessed while visiting someone else's farm
              !authState.matches("visiting") && (
                <Route path="/goblins" element={<Goblins />} />
              )
            }

            <Route path="/farm/:id" element={<Humans key="farm" />} />
            <Route path="/visit/:id" element={<Humans key="visit" />} />
            {/* Fallback */}
            <Route element={<Humans />} />
          </Routes>
```

I believe we also need to fix when a user manually enters `/#/visit/[ID]` into the URL bar to transition into the `visiting` state.

Pranav-yadav:
> I would suggest going one step further and never letting a visiting user access the Goblin page at all. This is a common React auth pattern that we can use.


Definitely. Had a thought about protected routes when encountered this problem and similar to what you mentioned tried changing path manually.. Hence implemented 'Forbidden' React.FC to be *more expressive* to the player..

As, you've shown adding check to the routes will increase the control over path changes...

Pranav-yadav:
> I believe we also need to fix when a user manually enters /#/visit/[ID] into the URL bar to transition into the visiting state.


💯 Yep.

Pranav-yadav:
One more thing, we should still show the Forbidden screen after blocking the path else, it'll show blank screen with no routes matched warning in console...

Pranav-yadav:
> I believe we also need to fix when a user manually enters /#/visit/[ID] into the URL bar to transition into the visiting state.


Definitely.. This is showing undefined behavior right now..

Pranav-yadav:
Also, we should opt for protected routes in future instead of manually performing checks every time..

When other functionalities are introduced which will involve funds and etc. It'll be very handy to have protected routes..

Pranav-yadav:
@adamhannigan 
added the check to /goblins routes as well as added fallback there as well.. So Forbidden screen is shown instead of Blank Screen.  :)

src/features/game/lib/goblinMachine.ts
```diff
@@ -126,10 +147,7 @@ export function startGoblinVillage(authContext: AuthContext) {
                   ),
               }),
             },
-            onError: {
```
adamhannigan, line 129: Without this, an error message would not be handled. The state machine would not know how to handle the error. Is there a reason this was removed?

src/features/game/lib/goblinMachine.ts
```diff
@@ -91,6 +92,19 @@ const makeLimitedItemsById = (items: LimitedItemRecipeWithMintedAt[]) => {
 };
 
 export function startGoblinVillage(authContext: AuthContext) {
+  // Temp check for determining if in visiting i.e. readonly state
+  const isVisiting = () => {
```
adamhannigan, line 96: In our application state, a user should never be visiting while the goblin machine is running.

I would suggest removing this logic from the machine as it should be handled elsewhere. We want to encapsulate only the state logic that should apply to goblin village 😃

src/Navigation.tsx
```diff
@@ -52,7 +53,18 @@ export const Navigation: React.FC = () => {
...
+                element={
+                  <Splash>
+                    <Forbidden />
+                  </Splash>
+                }
+              />
+            )}
             <Route path="/farm/:id" element={<Humans key="farm" />} />
             <Route path="/visit/:id" element={<Humans key="visit" />} />
             {/* Fallback */}
```
adamhannigan, line 70: Does this route fallback not work?

adamhannigan (commented):
Nice improvements mate 👌 

In this case here, we don't want to start supporting a visiting state in goblin village - this will lead to edge cases and bugs creeping in over time. It also adds complexity to the machine.

To keep things simple I would suggest just enforcing the business rule that "You can never visit Goblin town". It would be more scalable to enforce this on the routing level and never let them enter the goblin machine. Strange the fallback route was not working.

We have 2 options here:

1. Route them back to the main screen
2. Display a custom 404

Give that the goblins route is a valid route when playing, I would suggest they just fallback to the farming town

Pranav-yadav:
Okay.. 

2nd one is already implemented..
I've already added, the Custom **Forbidden** Component...

src/features/game/lib/goblinMachine.ts
Pranav-yadav: Okay. Reverted..

src/Navigation.tsx
Pranav-yadav: Checking...

adamhannigan (approved):
LGTM. Awesome work mate 🏆 🤗

PR #839, opened 2022-05-09
[FIX] Recipe Wei amounts

adamhannigan:
# Description

The wei amounts were not being converted correctly.

See before and after on testnet
Screen Shot 2022-05-09 at 12 10 26 pm
Screen Shot 2022-05-09 at 12 10 14 pm


spencerdezartsmith (approved):
🤩

PR #854, opened 2022-05-09
[CHORE] Token Redirect button

Amon-S:
I've added some buttons to redirect to the quick-swap pages for buying tokens and liquidity pairs into the wishing well. I've seen some people asking for some button on the game so if it's needed, here it is :D


spencerdezartsmith:
@Amon-S Thank you so much for this PR. We are currently reworking the Wishing Well modal. We will for sure add this functionality into our rework.

PR #892, opened 2022-05-12
[FIX] Seperate Visiting Context

spencerdezartsmith:
# Description

This PR aims to address the issues that we were facing with folks visiting other peoples farms. There were many issues that players faced when visiting contributors land eg. broken back button, unable to go back to your farm from visiting a contributor etc.

The main problem was that when a player was visiting a farm a new `gameMachine` was being invoked inside of the `authMachine` visiting state. This is the wrong context. This PR adds a `visitingMachine` and `VisitingProvider` which is simpler given the fact that visiting is essentially a read only state. The `authMachine` is the overarching machine which exists outside of navigation. We can send events to the `authMachine` from the `visitingMachine`. This helps because currently we have a visiting state for people who don't have a farm as well as a visiting state for a active land in a playing state. ie Town Hall.

This PR has also added a specific visiting file structure. This is to take this conditional logic out of the game components. All of these new components are basically read only.

This is part of a larger refactor of the state machines of the game as well as shared components. This is the first step.

_VisitingMachine_
Screen Shot 2022-05-13 at 7 52 52 pm

## Type of change

**NOTE**: Contributors ids are not available in dev/test mode. In order to test this you need to run locally and change line 79 to 

```
navigate(`/visit/${index % 2 > 0 ? 1 : 42}`);
```

 in `src/features/farming/townHall/components/Contributors.tsx`. The id's you add here can be any testnet id's

_To Test_
- From the starting screen click on `Explore ....`
- Add a farm number and visit a farm
- This should be a read only state ie. Crafting and Shop should be disabled

- Load your land and click on `Lets go`
- Go to the Town Hall
- Play around with visiting contributors lands

Sorry for the size of this. Its a massive chore 😬🍻 

# Checklist:


brian-man:
> Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

spencerdezartsmith:
> > Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

This is a draft PR. Not looking for feedback at this point 😄

j923sneaks (commented):
minor comments from me 🤗

c-py (approved):
Really, nice work. The refactor is simplifying the code a lot.

I tested using the the provided instructions and also just clicking around. During my clickfest the two notable points of interest:

1. After visiting a a non contributor farm via the Menu, the back button throws you back to the main auth screen rather than dropping you off back your authorised farm.

2. The house shows a level 2 farmer with no XP (I think regardless of the true level).

I don't believe either of these are exactly relevant to the refactor - It's working well, and good to go.

spencerdezartsmith:
> Really, nice work. The refactor is simplifying the code a lot.
> 
> I tested using the the provided instructions and also just clicking around. During my clickfest the two notable points of interest:
> 
> 1. After visiting a a non contributor farm via the Menu, the back button throws you back to the main auth screen rather than dropping you off back your authorised farm.
> 2. The house shows a level 2 farmer with no XP (I think regardless of the true level).
> 
> I don't believe either of these are exactly relevant to the refactor - It's working well, and good to go.

Yeah there is currently a bit of a disconnect between the visit states... one lives in the top level of the `authMachine` and the other is inside of the `connected` state which is like another machine. I think this is ok for now and will be easier to fix once the `visitingMachine` becomes a child of the `authMachine`.

The XP one I think can be looked at separately as I don't think it's related to this PR.

tiffanydys:
Not sure as to why the Shop got misplaced but everything else seems to be working just fine. Great work 😄😊

image

tiffanydys (approved):

spencerdezartsmith:
> Not sure as to why the Shop got misplaced but everything else seems to be working just fine. Great work 😄😊
> 
> image

Weird! I will take a look at this!!

spencerdezartsmith:
> > Not sure as to why the Shop got misplaced but everything else seems to be working just fine. Great work 😄😊
> > image
> 
> Weird! I will take a look at this!!

FIXED... just a missing `absolute` on the shop.

src/features/game/Humans.tsx
```diff
@@ -10,18 +11,29 @@ import mapMovement from "./lib/mapMovement";
...
+  const [authState, send] = useActor(authService);
   // catching and passing scroll container to keyboard listeners
   const container = useRef(null);
   const { id } = useParams();
   const [scrollIntoView] = useScrollIntoView();
 
+  useEffect(() => {
+    // Move auth machine out of visiting state if route changed via browser back button
+    if (
+      authState.matches({ connected: "visitingContributor" }) ||
```
adamhannigan, line 27: How can a player be in the `visiting` or `visitingContributor` state if they are in the `Humans` component?

I find putting logic in useEffect a bit disconnected at times. It is not clear what logic was fired to get into this state. It is a bit confusing how this logic would get called.

Is there somewhere else we could put this business logic and fire `send("RETURN");`?

src/features/game/lib/visitingMachine.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const setFarmDetails = assign<Context, any>({
+  farmId: (_context, event) => event.data.farmId,
+  address: (_context, event) => event.data.address,
+  isBlacklisted: (_context, event) => event.data.isBlacklisted,
+  owner: (_context, event) => event.data.owner,
+  state: (_, event) => event.data.state,
+});
+
+export function startGame({ farmToVisitID }: { farmToVisitID: number }) {
```
adamhannigan, line 50: A much cleaner machine. Well done 👏

src/features/visiting/house/House.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+    return <div className="flex flex-wrap">{badges}</div>;
+  };
+
+  const Content = () => {
+    if (isSkillTreeOpen) {
+      return <SkillTree back={open} />;
+    }
+
+    return (
```
adamhannigan, line 98: Consider componentizing this to ensure the visiting and farm House UI remains consistent.

src/features/visiting/hud/components/InventoryTabContent.tsx
```diff
@@ -0,0 +1,177 @@
...
+  inventory: Inventory;
+  inventoryItems: InventoryItemName[];
+  onClick: (item: InventoryItemName) => void;
+}
+
+const TAB_CONTENT_HEIGHT = 384;
+
+const isSeed = (selectedItem: InventoryItemName) => selectedItem in SEEDS();
+
+export const InventoryTabContent = ({
```
adamhannigan, line 36: Do we need to duplicate this component? It doesn't look like it is specifically reading from the `gameMachine`

src/features/visiting/hud/components/Share.tsx
```diff
@@ -0,0 +1,71 @@
...
+import farmImg from "assets/brand/nft.png";
+import { CopyField } from "components/ui/CopyField";
+
+interface Props {
+  farmURL: string;
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+export const Share: React.FC<Props> = ({ farmURL, isOpen, onClose }) => {
```
adamhannigan, line 18: Another component, do we need to duplicate it if it is a UI component?

src/features/visiting/water/components/Shark.tsx
```diff
@@ -0,0 +1,68 @@
...
+  "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg==";
+const imageSources = [sharkLeft, blankPng, sharkRight, blankPng];
+
+const getSharkPosition = () => {
+  const randomLeft = randomBetweenMaxInclusive(0, 73);
+  const randomTop = randomBetweenMaxInclusive(0, 90);
+  return { top: randomTop, left: randomLeft };
+};
+
+const Shark = () => {
```
adamhannigan, line 18: Same here, does this component differ in visiting and farm state?

adamhannigan (commented):
Nice work mate 👏 

The visiting machine is much cleaner.

There appear to be a few UI components that are duplicated. It makes sense to duplicate a component if it has a different data source (gameMachine vs visitMachine), but I feel we should still reuse functional UI components. The `House.tsx` is the main one that stands out to me - it is quite UI heavy and I feel we should abstract this so it remains consistent for visiting and farm states as we add more functionality in there

src/features/game/Humans.tsx
spencerdezartsmith: This is confusing, I agree but its because of the way our machines/navigation are set up.

When a user is authorised and playing the game they can visit a contributor via the town hall. When they do this their `gameMachine` dies and they spin up a `visitingMachine. This is controlled by the navigation ie `/visit...` loads `Visiting` which starts machine. Through an event we also move the `authMachine` into the `{ connected: 'visitingContributor' }` state since this machine is outside of the navigation. 
This makes for a better experience when a user hits in the `Back` button because they can just go back to their game rather than having to go through the entire connecting flow again which is unnecessary as we already have all of their details sitting in the authMachine. We have control over this because we control the back button with events. 
However, if a user uses the browser back button we are unable to fire an event to the authMachine because, like I mentioned above, it exists outside of our navigation. To get around this, inside of the `Humans` component which is loaded on the game route, on load, we check the current state of the authMachine and if it is in one of the visiting states then move it back into its correct state by sending the `RETURN` event. If we didn't do this then it would be stuck in a visiting state when a user is back playing the game therefore we would miss events.

We can definitely fix this we just need to think about a scaleable solution to our machine/navigation architecture. It's too big to include in this PR obviously. I have added a clearer comment.

src/features/visiting/house/House.tsx
spencerdezartsmith: I made the `House` reuseable

spencerdezartsmith:
> Nice work mate 👏
> 
> The visiting machine is much cleaner.
> 
> There appear to be a few UI components that are duplicated. It makes sense to duplicate a component if it has a different data source (gameMachine vs visitMachine), but I feel we should still reuse functional UI components. The `House.tsx` is the main one that stands out to me - it is quite UI heavy and I feel we should abstract this so it remains consistent for visiting and farm states as we add more functionality in there

I have made the shared components reuseable and moved them to `src/components`:

- Balance
- Share
- Shark
- Frog
- Inventory -> InventoryItems -> InventoryTabContent

These were also duplicated in Goblin Town. Now `farming`, `goblins` and `visiting` all share the same components.

adamhannigan (approved):
LGTM mate! 🚀

adamhannigan:
Thanks for the contribution and feedback! ❤️ 

 We are moving towards using an ideas portal to help capture and vote on community ideas. Github will only be used for bug/technical reports. We are only a small team and it is becoming difficult for us to respond individually to every case so we have began using community voting software.   

 If you could please upload your idea to the following community portal it would be greatly appreciated 🤗   

 https://sunflower-land.ideas.aha.io/

 You can vote on similar ideas and help bring ideas to our attention. Thanks again for being apart of this community!

PR #1007, opened 2022-05-25
[FIX] Remove copy button from Goblin Bank

c-py:
# Description

This PR removes the copy button from the Goblin bank.

Users have reported issues using the button on Android. Removing this button prevents players getting confused, or any future copy/paste issues. 

User's will now need to click the reveal "eye" icon prior to getting the farm address.

Screen Shot 2022-05-25 at 2 22 42 pm

## Type of change

# How Has This Been Tested?

This was tested manually

# Checklist:

- [X] My changes generate no new warnings


tiffanydys:
Hello, just a couple of questions: 
How will we copy the farm address after the change? 
Will there be another way to copy it?

Thanks 😄

c-py:
> Hello, just a couple of questions: How will we copy the farm address after the change? Will there be another way to copy it?
> 
> Thanks 😄

It's a great question, the new flow would include clicking the "eye" button to reveal the address, and then copy using the browser's inbuilt copy mechanism. I tested it on Android (Metamask), iOS (Metamask) and Firefox. In all cases it was possible to hold down to highlight the text and copy the address.

It's not the nicest experience. The copy button is more elegant, but it also incentives players to copy blindly without checking the destination address. 

At least not having this button, is that it is clear that the onus is also on the player to ensure they have copied the right address. But obviously there is a tradeoff on the user experience. 

@tiffanydys Do you have any ideas how we could find a middle ground or if one exists 😄 ?

ShiningMinions:
I want to preface by saying I don't mean any disrespect with what I'm about to say - I just want to give you some truthful criticism and feedback as a player in the community.

As a player I absolutely hate this feature, and I guarantee the player-base will hate it too. Why are we choosing to make the game more difficult to play? Why are you changing the game-play for everybody, instead of just honing-in for players on Android? Implementing this feature will just leads to MORE confusion and LESS convenience to the player, and is making the game less fun to play overall. So as a player, I'm giving you a check-and-balance on this one. 

Ref: https://sunflower-land.ideas.aha.io/ideas/SFL-I-117

brian-man:
> Users have reported issues using the button on Android. Removing this button prevents players getting confused, or any future copy/paste issues.

I'm confused.  So some Android users are having with the most useful button in the bank and everyone else loses the functionality because of it?  Speaking for part of everyone else, that kind of sucks. :|

adamhannigan:
Hey all, thought I would weigh in some thoughts here.

From a liability perspective, we should reduce any action that could cause issues when dealing with sending and receiving of users assets. 

By enabling a "copy" button we face 2 issues:

- Copy button breaks and the fault is on us
- Clipboard/click jacking software on a user's device injects a different address onto the clipboard.

For both of these cases, if a user manually copies the address and pastes it will reduce the risk. When a user needs to manually highlight and copy it adds that extra sense of awareness to what they are performing and they can visible check the address matches that they copy pasted.

Whilst it makes the UX a bit more difficult, leaving a copy button introduces a range of risks for us as a project and users transferring their assets.

In essence if that copy button breaks again and someone sends 100K USD of resources to the wrong address we **DO NOT WANT TO BE LIABLE**

ShiningMinions:
@adamhannigan everything I do in crypto ever has that copy button - every exchange, game, etc. For crying out loud, even openSea has that button - I just checked it now:

open-sea-copy-button

It may not seem like a big deal from your perspective, but back-door decisions like this really frustrates a player like me to the max. From my perspective, it simply doesn't make sense when literally everything else in crypto has it. It makes the game harder to play and less fun. **If any other player reads this and disagrees with me, please feel free to down-vote this comment.**

Please consider an alternative approach:
- Only apply this feature to the people who are actually having the problems: i.e. Android users. Why inconvenience everyone?
- Update your TOS to say you're not liable or something like this (to be fair, I'm not a legal expert here).
- Add a disclaimer right below the copy button "PLEASE DOUBLE CHECK YOUR ADDRESS AFTER PASTING", or something like this.

brian-man:
ShiningMinions:
> ...
Agree with everything said.

adamhannigan:
@ShiningMinions Fair points.

I feel the disclaimer is important to add on this screen. Is there a way to detect which devices our solution won't work on?

From my point of view I would still like to see the whole address and select, copy and paste it myself. As a rule of thumb, I never use copy/clipboard functionality when sending crypto. This is quite common practice especially when sending valuable items to avoid any malware or dodgy behaviour.

ShiningMinions:
Really appreciate you hearing me and the others out about this. Ideally, this conversation should have happened before valuable development hours were consumed (shameless plug: https://sunflower-land.ideas.aha.io/ideas/SFL-I-117).

> I feel the disclaimer is important to add on this screen. Is there a way to detect which devices our solution won't work on?

What if we parsed the User-Agent string of the browser somehow? That might be the solution.

> From my point of view I would still like to see the whole address and select, copy and paste it myself. As a rule of thumb, I never use copy/clipboard functionality when sending crypto. This is quite common practice especially when sending valuable items to avoid any malware or dodgy behaviour.

I'm a well-versed crypto user myself, and I know you're technically correct and your way is best, but surprisingly even I don't do what you do. I find that as long as I verify the first 4 and last 4 digits of my address, that's good enough for me. That's because I know even if my clipboard got hijacked, the likelihood of the hijacker to have an address matching the first 4 and last 4 digits of my own address is almost impossible.

### Here's a potential middle-ground approach:

1. Don't show the copy button on Android phones only.
2. On all other devices:
 a. Show the copy button
 b. Once the "copy" button is clicked, immediately show the entire address. Continue to show the "copied" tooltip also.
 c. Finally, once the full address is shown, show one of those red warning texts underneath: "WARNING! Please double check your pasted address is the same as above!" or something like this. I'm not a fan of those warning texts, but I would 100% want one here.

adamhannigan:
@ShiningMinions That solution sounds good to me.

I would add that we initially display the entire address. For those that want to manually highlight and copy, it would not be clear that they first need to click the copy button to see that address

c-py:
Closing in favor of https://github.com/sunflower-land/sunflower-land/pull/1027

PR #1044, opened 2022-06-02
[FEAT] Add progress bar and hover details to chickens

spencerdezartsmith:
# Description
This PR adds a progress bar and details of time remaining till egg is ready.

An optimisation was made on the chickens so the component wouldn't re render on every interval tick as it was unnecessary and could potentially end up 20x when boosts are applied. I have kept this in place since the progress bar only changes quarterly ie. progress > 25, progress > 50 etc.

The progress bar will be visible when the chicken is in the `fed` state therefore the percentage will update whenever the chicken moves between `happy` and `sleeping` states (max 20 seconds).

However, hovering over the chicken when in these states will give you the exact time to egg and will count down via an interval that exists only in the popover. This prevents the intervals running always for each chicken. I have also added a debounce on the hover because the chickens are so close to each other it smooths out the hover events.

chickens-progress

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "not op_mini all"
   ],
   "dependencies": {
+    "@types/lodash.debounce": "^4.0.7",
     "@types/lodash.shuffle": "^4.2.6",
     "@xstate/inspect": "^0.6.2",
     "@xstate/react": "^1.6.3",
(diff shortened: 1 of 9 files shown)


c-py (approved):
Looking sweet!

looks awesome

**question (non-blocking):** Do you think we should add the resource in the on mouse over?

On Hover Sunflower

spencerdezartsmith:
> Looking sweet!
> 
> looks awesome
> 
> **question (non-blocking):** Do you think we should add the resource in the on mouse over?
> 
> On Hover Sunflower

I don't think we need it. Crops have it because the resource could be one of many... in the case of chickens it could only be an egg. Also we don't have a lot of space so I think it makes sense to leave the display lean.

PR #1088, opened 2022-06-15
[FEAT] Enable Chickens

adamhannigan:
# Description

This PR enables chickens for production. It also tidies up the following:

- A missing chicken
- Prevents them from minting more chickens than the max


spencerdezartsmith (approved):
🐔

Issue #1089, opened 2022-06-15
[CHANGE] Custom Input in Buying & Selling


hujikol:
## How the Feature is Currently Working

Currently, the buying and selling experience is quite exhausting, we need a lot of clicking when we want to sell custom amounts and not all of the stock, and also so much clicking with custom buying amounts.

### _i.e. we need to click 69 times the 'buy 1' button when we want to buy 69 seeds of beetroot._
<img src="https://user-images.githubusercontent.com/12151051/173821290-8aeb3b44-0d3a-4023-b708-98f648d12f9b.png" width="540" />

### _or, we should click 100 times to sell the pumkins if I want to keep the remaining 300's_
<img src="https://user-images.githubusercontent.com/12151051/173821772-f9d339e5-c875-4fbb-8bfd-5870904b813d.png" width="540" />

## How the Feature Should Be Changed

In that matter, I suggest a solution to add a custom input field, and a stepping button, but also do pick max avaliable to buy or to sell. This custom input can also be implemented in the future in-game market feature or even in the withdraw panel.

> I'm also aware there is already a great solution in PR #1079 , which is to withdraw by long-pressing the mouse button.
> But in my opinion, it is just not so optimal way to do so, i.e. we want to withdraw 5k items at a time but we need to wait like 5 or more seconds to fill up the 5000 items.

## Proof of Community Demand
This issue is based on my aha proposal: https://sunflower-land.ideas.aha.io/ideas/SFL-I-130

Lot of farmers also need the same feature to be implemented (source discord suggestion channel ):
<img src="https://user-images.githubusercontent.com/12151051/173818721-16f61233-4679-4474-a031-924db046c2d4.png" width="640" />

<img src="https://user-images.githubusercontent.com/12151051/173820970-45049df9-0d1c-4096-ad84-51de5a8444d0.png" width="640" />

<img src="https://user-images.githubusercontent.com/12151051/173822742-b785a251-8948-49e5-9883-a2e9f2e980d7.png" width="640" />

<img src="https://user-images.githubusercontent.com/12151051/173822783-32dd176d-4d8e-4d32-9a2c-1bd6d0947bb6.png" width="640" />

<img src="https://user-images.githubusercontent.com/12151051/173822854-1846a8e4-0a0c-4f76-b1ae-c3c8c82f2811.png" width="640" />

[...]

PR #1090, opened 2022-06-15
[FEAT] Custom Buying & Selling Input

hujikol:
# Description

This idea brings freedom and avoids too much clicking when we want to sell or buy in the custom amount (Bulk Buy/ Sell).
Feature change #1089 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Manual testing has been conducted (see demo video), and local development testing has been conducted.

# UI Changes

## Buying
image

## Demo Video
https://user-images.githubusercontent.com/12151051/173825180-499b5100-551b-4b9f-a8b6-3465ea1c35a8.mp4

# Checklist:

- [x] Title of the PR is relevant and is prefixed with [FEAT], [CHORE], or [FIX]
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] Existing unit tests pass locally with my changes
- [ ] Code has been refactor.


Demonbane18:
nice i hope this get approved

hujikol:
Hello @tiffanydys, thank you for the follow-up ^_^, 
and thank you for the PR references, I've been read them and have a summary of my own, 
...
on that 2 PR have so many variations of the UIX, and also I'm aware that the team also considering the simplicity of the current flow, while the already mentioned PR has more complicated UIX (in my opinion).

as mentioned by @adamhannigan 
> The typically workflow right now is plant -> harvest -> sell. For the majority of users, I predict they would click the 'Sell All' button.
> 
> Adding an extra step would add friction to this flow. I would suggest ensuring that the workflow is firstly suited to accommodate the majority of users and how they play. It should be the 5-10% who want to do a custom bulk sell that needs to do an extra step.

So I came up with this new variation of the UIX as I do my own research on it, it's a more simple and intuitive way. A lot of the UIX I'm using are utilizing the current component, and in the way of interaction principle, as I learn it, it should be no longer than 3 steps to achieve the goals. While I also arranged the button in that way so the user usually scans top to bottom.

But I see the team right now is focusing to build the land expansion thing and want to manage all the features easily, but sooner or later it will be useful to be implemented in the in-game marketplace too... but yet I'm open to all opinions.

spencerdezartsmith:
Hey there @hujikol. Thank you so much for your contribution. As @tiffanydys mentioned we have had a few PR's with a similar implementation. We want to stay away from inputs as much as possible. We don't like the UX of them. 

The `0` not removing ect and mobile is also an issue with things overflowing...

Screen Shot 2022-06-17 at 8 56 58 am

Screen Shot 2022-06-17 at 8 56 30 am

I think a better approach here and something we would be more willing to consider is to just add a `Sell 10` button to align with the `Buy 10`. 

Again we're super appreciative of the effort but we will close this one for now. Please feel free to implement the suggestion above.

hujikol:
> I think a better approach here and something we would be more willing to consider is to just add a Sell 10 button to align with the Buy 10.

thankyou for the input, I think it's another solution too but still, we need to encourage the artist to make a better layout to be implemented soon.

I'm also thinking about adding max buy to align it with your suggestion, how about that idea?
so there will be 3 buttons in each section, one, ten, all.

I also keep trying to design a new UI for the mobile version.

Polysemouse:
If this feature is implemented in the future, could we also have a confirmation when buying items?  Sometimes I would try buying an item but I mistakenly bought the wrong one.

It would be nice to have a unified UI/UX for both buying and selling (both having a confirmation).

PR #1112, opened 2022-06-21
[FIX] Use available balance in bank

adamhannigan:
# Description

The Goblin bank was showing items/tokens that had already been used in off chain - i.e. just showing the on chain value. This would cause an API error if attempting to withdraw these items.

**Solution** - Use the state that is stored on the goblin state.

See example below where Goblin Village only has 2995 but the bank displays 2999

Screen Shot 2022-06-21 at 2 22 59 pm

Fixes #issue

## Type of change


c-py:
@adamhannigan Do you think it's possible for the opposite to happen? That off-chain SFL may have already been spent but we display on-chain?

Perhaps we need to display the minimum of the two

c-py (approved):
LGTM!

PR #1158, opened 2022-06-30
[Feat] Land Expansion Pebble

Rofrtd:
# Description

This PR introduces the pebble to land Expansion.

Fixes #issue

## Type of change

# How Has This Been Tested?

Added tests

# Checklist:


src/features/game/expansion/components/resources/Pebble.tsx
```diff
@@ -0,0 +1,254 @@
...
+      await new Promise((res) => setTimeout(res, 2000));
+      setCollecting(false);
+    } catch (e: any) {
+      displayPopover(
+        <span className="text-xs text-white text-shadow">{e.message}</span>
+      );
+    }
+  };
+
+  const timeLeft = getTimeLeft(pebble.stone.minedAt, PEBBLE_RECOVERY_TIME);
```
adamhannigan, line 140: Screen Shot 2022-07-08 at 9 25 43 am

It looks like the API payload does not include stone. This is causing a white screen. I would suggest updating the types to make `stone` optional and then checking whether it exists here.

Screen Shot 2022-07-08 at 9 26 16 am

src/features/game/expansion/components/resources/Pebble.tsx
```diff
@@ -0,0 +1,254 @@
...
+    >
+      {/* Unmined pebble which is strikeable */}
+      {!mined && (
+        <div
+          ref={containerRef}
+          className="group cursor-pointer w-full h-full"
+          onClick={strike}
+        >
+          <Spritesheet
+            className="group-hover:img-highlight pointer-events-none z-10"
```
adamhannigan, line 156: Screen Shot 2022-07-08 at 9 27 42 am

Looking closely the pebble is a bit blurry.

Similar to the water PR, we need to make sure that a pixel width is the exact same throughout the entire game. In this case here the pebble pixel appears smaller than the grass pixels.

If you think a smaller pebble would look better, I would suggest designing a pebble that is smaller than the current one. Otherwise, scale it to ensure it fits the rest of the art in the game.

We need to be diligent with this and set the standard for the community. They will base their art of what we have in the game so it is extremely important we set the initial elements as pixel perfect 👍

src/features/game/expansion/components/resources/Pebble.tsx
```diff
@@ -0,0 +1,254 @@
...
+          top: "-28px",
+          opacity: collecting ? 1 : 0,
+          transition: "opacity 0.2s ease-in",
+          imageRendering: "pixelated",
+        }}
+        className="pointer-events-none z-20"
+        getInstance={(spritesheet) => {
+          minedGif.current = spritesheet;
+        }}
+        image={dropSheet}
```
adamhannigan, line 193: Screen Shot 2022-07-08 at 9 30 50 am

The green border on the stone looks strange on top of the water. I would suggest changing the art to use a grey that has some transparency so it would look good on top of water or on top of land.

Also, once the sprite sheet plays it just seems to stay there. Not sure if this should be working yet.

src/features/game/expansion/components/resources/Pebble.tsx
```diff
@@ -0,0 +1,254 @@
...
+        )}
+      >
+        <HealthBar percentage={collecting ? 0 : 100 - (touchCount / 2) * 100} />
+      </div>
+      {/* Recovery time panel */}
+      {mined && (
+        <div
+          className="absolute"
+          style={{
+            top: "41px",
```
adamhannigan, line 229: Currently there is no visual feedback to a player that a pebble is in 'recovery mode'. This would be confusing for players. I would suggest adding some opacity to the image to make it appear inactive.

E.g.

Screen Shot 2022-07-08 at 9 32 59 am

adamhannigan (commented):
Time to get some rock! Excited to get this live 🚀 

Similar to the water PR, we need to be pixel perfect with the artwork and set the right standards for the community.

I've also noticed that the artwork that has been added is twice the size it needs to be (i.e. 1 pixel is actually 2px wide). All art that gets added should retain a 1px = 1px. This makes the files smaller and makes it easier to place throughout the game.

adamhannigan (changes requested):
Previous review was meant to request changes 🤦

src/features/game/expansion/components/resources/Pebble.tsx
spencerdezartsmith: This API payload is not correct due to you having an old record `pebbles` in your gameState in testnet. This was the shape of the payload in the past iteration. You will need to delete your current `pebbles` and then reload to get a clean one. 

The `undefined` issue should still be handled though. `stone` should definitely be optional. It is on the api side.

spencerdezartsmith:
> Time to get some rock! Excited to get this live 🚀
> 
> Similar to the water PR, we need to be pixel perfect with the artwork and set the right standards for the community.
> 
> I've also noticed that the artwork that has been added is twice the size it needs to be (i.e. 1 pixel is actually 2px wide). All art that gets added should retain a 1px = 1px. This makes the files smaller and makes it easier to place throughout the game.

The increase in the size of the image is my bad. I did that when I was struggling to get the Spritesheet to position correctly. Its not needed anymore because it actually wasn't the issue in the end. Romy you can use the original images for this again. Reach out if you need help.

adamhannigan (approved):
Nice work mate. LGMT 🚀

Issue #1116, opened 2022-06-21
[FEAT] Allow Players to Remove Growing Crops (Shovel)


ShiningMinions:
## Pixelized Crib Trello Ref:
https://trello.com/c/pARkxAiI

## Summary of Request

As a player, I want a way to remove crops from my field (which are currently growing), so that I can correct seeds I mistakenly planted, and I don't have to wait the entire grow cycle to re-plant.

## Proof of Community Demand

- [Aha! Proposal #107](https://sunflower-land.ideas.aha.io/ideas/SFL-I-107)
- [Aha! Proposal #68](https://sunflower-land.ideas.aha.io/ideas/SFL-I-68)
- VIP Channel Poll:
![VIP Poll](https://trello.com/1/cards/629e0275b4a9622820eb2aa6/attachments/62a6949b7b8f570653028139/previews/62a6949b7b8f570653028155/download/community-demand-remove-crops-vip.png)
- HODL Channel Poll:
![HODL Poll](https://trello.com/1/cards/629e0275b4a9622820eb2aa6/attachments/62a6949a7902252256c2b27a/previews/62a6949a7902252256c2b28d/download/community-demand-remove-crops-hodl.png)

## Contributors

If know, please provide the list of contributors who will be working on this feature.

- TBD

## README - Pre-Notes:

We believe every farm should have the ability to remove unwanted crops by default - it's very frustrating as a player not being able to remove crops, and we feel every player should have this ability by default. However, we do like the idea of minting "better shovels" at some point in the future. For example, maybe a "wooden shovel" will give you 10% chance to recover seed. Maybe "stone shovel" would be 25%, etc. Inspiration for this idea came from [Aha! Proposal 107](https://sunflower-land.ideas.aha.io/ideas/SFL-I-107).

As a quick future brainstorm, we think the ability to mint better shovels could be unlocked only after completing some new "quest". This would be extremely fun. The quest lore could have something to do with the goblin who steals your shovel. He could "teach you the secret of crafting better shovels", or something like this. This is out-of-scope for this feature request, but could be extremely fun content for future.

## Technical Requirements

1. All farms should be able to remove crops. Therefore, every farm should start with exactly one "rusty shovel" by default. This "rusty shovel" is a non-withdrawable inventory item, and its only function is to remove unwanted crops currently growing in the player's plots.
 a. Description: "An old rusty shovel. Equip to start removing unwanted crops."
[...]

PR #1203, opened 2022-07-09
[FEAT] Allow Crop Removals and Rusty Shovel

ShiningMinions:
# Description

Please see details, pre-approved requirements, and original Issue here: https://github.com/sunflower-land/sunflower-land/issues/1116
This PR also closes this Github Discussion: https://github.com/sunflower-land/sunflower-land/discussions/1126

This PR adds a new "rusty shovel" item and allows players to remove unwanted growing crops. The "rusty shovel" will do nothing more than remove crops which are currently growing.

This PR also sets-up the possibility of adding "better shovels" to be added into the game in the future. These "better shovels" are out-of-scope for this particular PR.

## Discrepancies from Original Requirements:

1. In original requirements, I missed providing details how existing farms would currently get the "rusty shovel". Suggestions:
 a. Sell the "Rusty Shovel" at shop for 0 SFL. Will then be permanently sold-out for that farm.
 b. Air-drop the "Rusty Shovel" to existing farms. New farms will be provided one upon new mint.

2. Some various English grammar has been changed in an attempt to make it easier to understand.

3. I didn't add the "Cancel" button to the warning modal, as the "X" icon seemed sufficient.

4. I made the warning modal cool-down is set to **5 seconds** instead of **10 seconds**, so that it pops-up more frequently.

## Outstanding Items:
- [ ] The "Rusty Shovel" is currently being force-added to gameState on load for dev purposes. This code needs to be removed before merge.
- [ ] CoreDevs need to provide the new assetID for the "Rusty Shovel".
- [ ] Decision will need to be made by CoreDevs how the "Rusty Shovel" should be added to existing farms (see options above).
- [ ] Consider adding a sound-effect to the "removeCrop" action.

## Type of change

# How Has This Been Tested?

I have created a new test suite which test crop removals (see removeCrop.test.ts).

rust-shovel-test-results

I also tested some parts visually, specifically, the "warning modal":

# Checklist:

# Credits:

- Feature Proposer, Requirements Writer, Coder: @ShiningMinions
- Critics: Polysemouse (nezumijeep#3310 - Discord), Noisy#0666 (Discord)
- Artist (Rusty Shovel) - LittleEinstein#0439 (Discord)


Demonbane18:
great work

ShiningMinions:
Note for CoreDevs:

Please see "Outstanding Items" checklist above :+1:

src/features/farming/crops/components/RemoveCrop.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { Modal } from "react-bootstrap";
+import { Panel } from "components/ui/Panel";
+import { Button } from "components/ui/Button";
+
+import close from "assets/icons/close.png";
+import rustyShovel from "assets/tools/rusty_shovel.png";
+
+interface Props {
+  doShow: boolean;
```
spencerdezartsmith, line 10: Could this prop be renamed to "showWarning" to make it a little clearer?

src/features/farming/crops/components/RemoveCrop.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onConfirm: () => void;
+  onClose: () => void;
+}
+
+export const RemoveCrop: React.FC<Props> = ({ doShow, onConfirm, onClose }) => {
+  return (
+    <Modal centered show={doShow} onHide={onClose}>
+      <Panel>
+        <img
```
spencerdezartsmith, line 19: Screen Shot 2022-07-12 at 12 59 10 pm

The alignment of the header with the absolute positioned close icon is a bit too squashed. Might be worth trying to position both using flexbox.

src/features/farming/crops/components/RemoveCrop.tsx
```diff
@@ -0,0 +1,46 @@
...
+          className="h-6 top-4 right-4 absolute cursor-pointer"
+          onClick={onClose}
+        />
+        <div className="p-2">
+          <h1 className="text-xl text-center">Remove Growing Crops?</h1>
+          <div className="flex my-4 justify-center">
+            <img src={rustyShovel} style={{ width: "50px" }} />
+          </div>
+          <p className="text-sm mb-4">
+            Warning! You have your shovel equipped.
```
spencerdezartsmith, line 30: I think this would read better as `Your shovel is equipped!` Let the exclamation point be the warning.

src/features/farming/crops/components/RemoveCrop.tsx
```diff
@@ -0,0 +1,46 @@
...
+          </p>
+          <p className="text-sm mb-3">
+            Are you sure you want to start removing crops? You will not recover
```
spencerdezartsmith, line 33: Also here maybe `lose` instead of `not recover`. 

`You will lose both the crop and its seed!`, Also we can just use string interpolation here instead of the `&apos;` code.

src/features/farming/crops/components/RemoveCrop.tsx
```diff
@@ -0,0 +1,46 @@
...
+            the crop or it&apos;s seed!
+          </p>
+        </div>
+
+        <div className="flex">
+          <Button className="text-sm" onClick={onConfirm}>
+            Yes, Start Removing
```
spencerdezartsmith, line 40: How about `Start digging!` which I think is a little more playful?

spencerdezartsmith (commented):
This is excellent @ShiningMinions 🤩 . Very high quality starting with the description all the way to the code. Awesome job. I left a little bit of feedback mostly in regards to copy. I would totally support this feature being added. 🥇

src/features/farming/crops/components/Field.tsx
```diff
@@ -40,6 +43,36 @@ export const Field: React.FC<Props> = ({
   const clickedAt = useRef<number>(0);
   const field = game.context.state.fields[fieldIndex];
   const [showCropDetails, setShowCropDetails] = useState(false);
+  const [showRemoveWarning, setShowRemoveWarning] = useState(false);
+  const [lastCropRemovedTime, setLastCropRemovedTime] = useState<number | null>(
```
0xsunflowerstudios, line 47: I would suggest showing the modal everytime that you need to remove a crop. Since the main use case is fixing up a few errors (i.e. incorrectly planted crops), then we should design around this core focus.

There is quite a bit of extra logic to do the conditional rendering of the modal which adds a bit of bloat, maintenance and potential bugs in the future.

src/features/game/events/removeCrop.ts
```diff
@@ -0,0 +1,105 @@
...
+  LOCKED_LAND = "Goblin land!",
+  NO_CROP_PLANTED = "There is no crop planted!",
+  READY_TO_HARVEST = "Plant is ready to harvest!",
+  INVALID_PLANT = "Invalid Plant!",
+}
+
+export type RemoveCropAction = {
+  type: "item.removed";
+  item?: InventoryItemName;
+  index: number;
```
0xsunflowerstudios, line 21: For readability, could we call this `fieldIndex`

src/features/game/events/removeCrop.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  if (action.index < 0) {
+    throw new Error(REMOVE_CROP_ERRORS.FIELD_DOESNT_EXIST);
+  }
+
+  if (!Number.isInteger(action.index)) {
+    throw new Error(REMOVE_CROP_ERRORS.FIELD_DOESNT_EXIST);
+  }
+
+  if (
```
0xsunflowerstudios, line 54: Do we need to check the locked fields since there won't be any growing crops on them?

All of the crops are already to harvest on the locked land, which means the clauses below should throw and error.

src/features/game/types/index.ts
```diff
@@ -11,6 +11,7 @@ export enum LimitedItemType {
 }
 
 export const KNOWN_IDS: Record<InventoryItemName, number> = {
+  "Rusty Shovel": 100, // TODO - get correct ID from coreDevs
```
0xsunflowerstudios, line 14: We can put this in the tools category as ID `307`

0xsunflowerstudios (commented):
Awesome work @ShiningMinions 

I've left a few suggestions.

In terms of the functionality:

1. We can show a rusty shovel on the farm that users pick up for the first time
2. I still feel that we want users to earn this shovel - If we are spending time building this into the game we want player's to experience to joy of grinding and earning that shovel. This shouldn't be too much work adding a 'Basic Shovel' to the Blacksmith on the UI side of things - this mainly uses config objects that are ready to go (see `craftables.ts`)

src/features/farming/crops/components/RemoveCrop.tsx
ShiningMinions: In fact, I just discovered the word "it's" is incorrect grammar here anyways. "it's" should only ever represent "it is". I will simply change to "its" instead :+1: .

ShiningMinions:
@spencerdezartsmith new UI for warning modal:

new-warning-ui

src/features/farming/crops/components/Field.tsx
ShiningMinions: > I would suggest showing the modal everytime that you need to remove a crop. Since the main use case is fixing up a few errors (i.e. incorrectly planted crops), then we should design around this core focus.

@0xsunflowerstudios This is definitely not a blocker for me - I am willing to change this if needed so we can get the feature out to the community :+1: . However, it's worth mentioning that the tech requirements of this feature went through a few iterations on my end with several community members. The feedback I have received by community has greatly indicated the desire to freely remove crops, without a warning modal constantly hounding the player.

> There is quite a bit of extra logic to do the conditional rendering of the modal which adds a bit of bloat, maintenance and potential bugs in the future.

@0xsunflowerstudios perhaps a refactor would be best here? Or alternatively, perhaps we create a new "Chore Request" to clean-up this logic after merge? I agree that this adds bloat, but I also wouldn't want to deny the community a feature that provides them value.

src/features/game/events/removeCrop.ts
ShiningMinions: @0xsunflowerstudios you are correct - I overlooked this detail. Will remove this logic and irrelevant tests next push :+1:

ShiningMinions:
@0xsunflowerstudios responding to your other comments:

> 1. We can show a rusty shovel on the farm that users pick up for the first time

Sounds good :+1: Ideally, I should have defined this portion in original issue's requirements, but I definately missed it. What I'm thinking is we show the "Rusty Shovel" next to the player's farm sign and make it clickable. Upon click, we should a modal something like "You found a rusty shovel!", and in the body say something like "You can equip this to start removing unwanted crops", and finally a button that says "Claim". This shovel will only appear on the ground if the player doesn't already have on in inventory. What do you think?

> 2. I still feel that we want users to earn this shovel - If we are spending time building this into the game we want player's to experience to joy of grinding and earning that shovel. This shouldn't be too much work adding a 'Basic Shovel' to the Blacksmith on the UI side of things - this mainly uses config objects that are ready to go (see `craftables.ts`)

Here's what I think what we should talk about: What does this next "better shovel(s)" look like, such that a player would "want to grind for it/them"? Should we talk about the possibility of adding wood/stone/iron/gold shovels? Perhaps these should be future NFTs, then when crafted, give you a chance to recover the seed?

I do feel like these "better shovels" are out of scope for this feature, and can be introduced in the future - but I agree that perhaps we should at least get **some idea** of what these "better shovels" will do.

src/features/farming/crops/components/Field.tsx
```diff
@@ -40,6 +43,36 @@ export const Field: React.FC<Props> = ({
...
+    null
+  );
+
+  // Record when some crop was last removed
+  useEffect(() => {
```
adamhannigan, line 52: One concern of this approach is that the component becomes more dependent on external conditions and events. Previously most of the state was internalised or focussed around a single field, now other fields can affect each other.

We are beginning land expansion and adding more functionality every week. To help scale and avoid bugs/maintenance we want to try keep the `Field.tsx` more contained.

After playing around with the modal/removing crops I was wondering if we could simplify this a bit and bring it back to the 'tappy' nature of the game.

When a Rusty Shovel is equipped, we can employ a double tap to remove a crop. This would work similar to chopping a tree, but instead if would remove the crop. Going down this avenue would help with the consistence of user actions in the game (tapping to harvest/mine/remove a resource).

To assist with the 'danger' aspect, I would consider using the red progress bar.

redbar_03

adamhannigan (commented):
In terms of equipping every player with a Rusty Shovel, I realised this may just be easier that we hardcode this on the API that every farm has the shovel. This will simplify the workflow of an 'airdrop'.

I see your point on not wanting to accept a modal each time. Makes sense 👍 

On the technical side of things I'd suggest we make some alterations that make the Field.tsx more contained. I want to avoid us bloating functionality and coupling it with other events in the system. After thinking about this approach I think it would make sense to follow the same multiple taps approach that the game currently uses (i.e. double tap to remove the crop) - see comments.

We will still want to go forward with the Basic shovel as well and release this at the same time even if it is a bit of scope creep. It is extremely hard to take functionality away from users so we can't give this functionality (removing crops) with the initial Rusty shovel.

src/features/farming/crops/components/Field.tsx
ShiningMinions: Got it! Can you just confirm my finalized understanding + answer me one question:

> In terms of equipping every player with a Rusty Shovel, I realized this may just be easier that we hardcode this on the API that every farm has the shovel. This will simplify the workflow of an 'airdrop'.

Perfect :+1: - just to confirm, I should not "put the rusty shovel next to the farm sign" - I should just leave the code as-is and have you hardcode the rusty shovel in the API. Correct?

> When a Rusty Shovel is equipped, we can employ a double tap to remove a crop.

Got it :+1: ! Just to be clear, this means we should kill the "warning modal" component altogether, correct?

> We will still want to go forward with the Basic shovel as well and release this at the same time even if it is a bit of scope creep.

Please just confirm me:

- Sell the "Basic Shovel" at the Tools shop on Human Side?
- AssetID 308 ? (rusty shovel is 307)
- Ingredients? Perhaps 5 stone, 10 wood, 3 SFL?

> It is extremely hard to take functionality away from users so we can't give this functionality (removing crops) with the initial Rusty shovel.

The only thing I'm unclear on is what is the functionality difference between the "Rusty Shovel" VS "Basic Shovel"? Are you saying the "Rusty Shovel" will simply sit in player's inventory without any functionality, and only the "Basic Shovel" will allow crop removal?

spencerdezartsmith:
> @spencerdezartsmith new UI for warning modal:
> 
> new-warning-ui

This is great! We're leaning towards using more sentence case so the button would be `Start digging!` but that is not a blocker here.

src/features/farming/crops/components/Field.tsx
adamhannigan: 1. Yes, you don't need to place the rusty shovel anywhere. We will sort that out
2. Yes, let's kill it 🔪 
3. Yes, Sell basic shovel at the tools shop
4. IDs are perfect
5. Let's make it more work for players 💪 10 iron, 20 wood
6. The rusty shovel will 'do nothing' - essentially it's functionality is that it lets you plant/harvest crops. When it gets stolen you can no longer farm

ShiningMinions:
@adamhannigan all changes have been implemented and ready for review! :rocket: 

@spencerdezartsmith , it was decided in above comments to go with a "double-tap" strategy instead of showing the "Warning Modal". The warning modal has been killed - I appreciate the feedback and will help me for future features :+1: .

src/features/farming/crops/components/Soil.tsx
```diff
@@ -47,6 +48,7 @@ export const Soil: React.FC<Props> = ({
   plantedCrop,
   className,
   showCropDetails,
+  isRemoving,
```
adamhannigan, line 51: Not a blocker, but we are beginning to leak the implementation details here.

In this case, the component shouldn't be aware of it's use case (i.e. is removing a crop) - we want UI components to remain decoupled and 'dumb'

In these cases I would suggest just adding the negative progress bar image directly in the Field.tsx component. After 2-3 uses we can then abstract it into a separate file.

adamhannigan (dismissed):
Great work ❤️ Players will love this!

We are working on the API changes and will merge this in once they are ready.

Outstanding contribution!

NoisyBucheron:
Good job @ShiningMinions! It's even better than we've expected, simple and practical! 
Kuebiko owners are going to love it

ShiningMinions:
@NoisyBucheron thanks a bunch! I think the most exciting part to notice about this feature is how many refinements it went through - both from BlueTeam Critics (Pixelized Crib) and GreenTeam (CoreDevs) - and how it resulted at the end. The feature is completely different then when I first proposed it, and it has been refined to pure gold because of everyone's feedback and collaboration. I'm so excited to see how the community reacts to this feature :rocket: .

@adamhannigan I will try to get that code refactored - I might not be able to within the next few days depending when this gets merged. At the very least, consider opening a brief [CHORE] request after merging, and I (or I'm sure even another community dev) would happily clean that up.

adamhannigan:
@ShiningMinions No worries mate.

I will get this today (once API is in) and might add the change while in there 👍

spencerdezartsmith (dismissed):
Excellent work on this @ShiningMinions. This is such a good example of cross collaboration. I love it 🤩

PR #1237, opened 2022-07-15
[FEAT] Land Expansion Tree

adamhannigan:
# Description

This PR adds support for trees with positions on a map. For testing, only 1 of the 5 initial trees has a valid placement on the map.

I have created a new `Tree.tsx` component and matched similar styles/behaviours with the other resources in Land expansion. It is pretty much copy paste with removal of the redundant code

**Includes**

- Remove progress bar
- Fix the shadow on the wood drop (change it to a shadow that also works on top of water)
- Change opacity when chopped
- Place tree in position on map
- Fix image sizing so 1px=1px

Screen Shot 2022-07-15 at 4 02 42 pm


c-py (approved):
LGTM. Was able to chop and get a nice ghostly stump.

PR #1239, opened 2022-07-15
Optimization of synchronous

1176517584:
# Description

Before there is no chicken house, MY seed can be basically planted before it needs to be synchronized, but after there is a chicken house, two eggs will be forced to synchronize, resulting in my seed has not been planted and synchronized, repeated many times, this is very unreasonable, because the price of an egg is not equal to 1SFL.

Fixes #issue

Please change the default price of eggs to 0.2-0.3SFL so that synchronization is not so frequent and seeds can be planted. Of course, this is just a suggestion. We only hope that when we own the chicken coop, we should not ask for synchronization frequently. Please accept it.


veenhdek:
It appears there is no codebase change in this PR, thus I would suggest you close this PR and open this in [discussions](https://github.com/sunflower-land/sunflower-land/discussions) as that is the right arena for this topic

PR #1247, opened 2022-07-19
[FEAT] Land Levels

adamhannigan:
# Description

This PR includes support for the base land structure for levels 1-10.

Based on the `level`, a different base will be rendered

**Water Offsets**

To ensure that water decorations do not clash with the land, they get "offsetted" based on the size of the land.

Screen Shot 2022-07-20 at 7 57 57 am

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How to test?

Put the game into art mode and go into `constants.ts`.

There you can change the `level` on the `INITIAL_FARM` object


spencerdezartsmith (approved):
This is great mate! Its really coming to life. I am wondering should be include in the images the boundary fence that will indicate the position of the next land expansion? ⛵

adamhannigan:
@spencerdezartsmith Yes agreed.

I've put a message in our design chat around this. Right now we need to figure out the different states and how expanding will work.

I.e. Empty state, building state and finished state.

Since we need to dynamically display between building + ready, it may make sense to dynamically display the fencing as well. 

Let's chat about this

PR #1275, opened 2022-07-27
[FIX] Max resource amounts

adamhannigan:
# Description

There was an issue where some people would collect a resource while under the limit which would then take them over the limit that the blockchain supported.

This PR ensures they fall below that limit and do not get transaction errors

### DIFF
--- a/src/features/game/lib/processEvent.ts
+++ b/src/features/game/lib/processEvent.ts
@@ -27,10 +27,10 @@ const maxItems: Inventory = {
   // Stock limits
   ...INITIAL_STOCK,
 
-  Gold: new Decimal("100"),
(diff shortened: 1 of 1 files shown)


Rofrtd (approved):
Nice one mate! :shipit:

veenhdek:
Great job to resolve this issue. Can we alter the error message/alert (line 124) to something like 

"You have reached a security earning cap for an item within a single session. For security reasons, please sync to the blockchain."

adamhannigan:
@veenhdek Yes, you read our minds! It is quite confusing for users at the moment.

@spencerdezartsmith Is making some improvements at the moment to the UI/UX around this.

PR #1277, opened 2022-07-28
[FIX] Need Sync Modal

spencerdezartsmith:
# Description
There has been a lot of confusion around the alert saying `You can only earn 100 SFL in a single session for security reasons. Please sync to the blockchain.`.

Screen Shot 2022-07-28 at 12 03 34 pm

This PR adds a clearer lore based message which alerts the user to the specific resource while also giving them the ability to sync from the modal. I will update the documentation to go along with this work. 

As part of this I have also refactored all the actions to make a deep copy of the state as soon as they're called. During my testing I found that there were some that were modifying the state argument directly. We want to avoid this were possible.

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #issue

## Type of change

# How Has This Been Tested?

Manually

# Checklist:

### DIFF
--- a/src/features/game/Game.tsx
+++ b/src/features/game/Game.tsx
@@ -111,7 +111,6 @@ export const Game: React.FC = () => {
           )}
           {gameState.matches("synced") && <Success />}
           {gameState.matches("syncing") && <Syncing />}
-          {/* Read */}
           {gameState.matches("needSync") && <MaxItems />}
         </Panel>
       </Modal>
--- a/src/features/game/Game.tsx
+++ b/src/features/game/Game.tsx
@@ -34,6 +34,7 @@ import { Refreshing } from "features/auth/components/Refreshing";
 import { GoblinShovel } from "features/farming/crops/components/GoblinShovel";
 import { Announcements } from "features/announcements/Announcement";
 import { Notifications } from "./components/Notifications";
+import { MaxItems } from "./components/MaxItems";
 
 const AUTO_SAVE_INTERVAL = 1000 * 30; // autosave every 30 seconds
 const SHOW_MODAL: Record<StateValues, boolean> = {
@@ -49,6 +50,7 @@ const SHOW_MODAL: Record<StateValues, boolean> = {
   refreshing: true,
   expanded: false,
   expanding: false,
+  needSync: true,
 };
 
 export const Game: React.FC = () => {
@@ -109,6 +111,8 @@ export const Game: React.FC = () => {
           )}
           {gameState.matches("synced") && <Success />}
           {gameState.matches("syncing") && <Syncing />}
+          {/* Read */}
+          {gameState.matches("needSync") && <MaxItems />}
         </Panel>
       </Modal>
       {/* check local storage and show modal if not read */}
--- a/src/features/game/Game.tsx
+++ b/src/features/game/Game.tsx
@@ -34,7 +34,7 @@ import { Refreshing } from "features/auth/components/Refreshing";
 import { GoblinShovel } from "features/farming/crops/components/GoblinShovel";
 import { Announcements } from "features/announcements/Announcement";
 import { Notifications } from "./components/Notifications";
-import { MaxItems } from "./components/MaxItems";
+import { Hoarding } from "./components/Hoarding";
 
 const AUTO_SAVE_INTERVAL = 1000 * 30; // autosave every 30 seconds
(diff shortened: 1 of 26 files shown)


src/features/game/Game.tsx
```diff
@@ -49,6 +50,7 @@ const SHOW_MODAL: Record<StateValues, boolean> = {
   refreshing: true,
   expanded: false,
   expanding: false,
+  needSync: true,
```
adamhannigan, line 53: I'm wondering if we can tie this closer to the actual domain of the game and whether `needSync` is just a technical term for the state?

Could we call this the `hoarding` state?

src/features/game/components/MaxItems.tsx
```diff
@@ -0,0 +1,85 @@
...
+    gameService.send("SYNC", { captcha });
+  };
+
+  const makeTitle = () => {
+    const regex = new RegExp(/^[aeiou]/gi);
+    const startsWithVowel = regex.test(maxedItem);
+    const indefiniteArticle = startsWithVowel ? "an" : "a";
+    const item = maxedItem === "SFL" ? maxedItem : maxedItem.toLowerCase();
+
+    return `Have you become ${indefiniteArticle} ${item} horder?!`;
```
adamhannigan, line 32: Could we simplify the `indefiniteArticle` here and say "Are you hoarding wood?"

src/features/game/events/chop.ts
```diff
@@ -66,17 +67,18 @@ export function chop({
   action,
```
adamhannigan, line 67: It could be useful to mark `state` as a readonly object now to prevent accidental usage in the future.

For example:

```
type Options = {
  state: Readonly<GameState>;
  action: ChopAction;
  createdAt?: number;
};
```

adamhannigan (approved):
LGTM. 

Just a few suggestions around using the concept "hoarding" instead of the technical "needs sync"

I've tested SFL + 5 other resources. Also able to farm, chop, mine etc.

src/features/game/components/MaxItems.tsx
spencerdezartsmith: This would mean we would need to have some pluralise logic that we would need to maintain. This was what we talked about this morning. 

I will change to `Are you an/a {resource} hoarder?`. We can't really escape the grammar check here.

src/features/game/events/chop.ts
spencerdezartsmith: Top suggestion 🥇

src/features/game/components/Hoarding.tsx
```diff
@@ -0,0 +1,85 @@
...
+    await new Promise((res) => setTimeout(res, 1000));
+
+    gameService.send("SYNC", { captcha });
+  };
+
+  const makeTitle = () => {
+    const regex = new RegExp(/^[aeiou]/gi);
+    const startsWithVowel = regex.test(maxedItem);
+    const indefiniteArticle = startsWithVowel ? "an" : "a";
+    const item = maxedItem === "SFL" ? maxedItem : maxedItem.toLowerCase();
```
c-py, line 30: Very nice touch!

src/features/game/components/Hoarding.tsx
```diff
@@ -0,0 +1,85 @@
...
+              {makeTitle()}
+            </span>
+            <img src={maxedItemImage} className="h-12 mt-2 mb-3" />
+            <p className="text-xs sm:text-sm mb-3">
+              {`I don't want to alarm you but we just heard that the Goblins have
+              noticed that you are hording a lot of this resource off chain.`}
+            </p>
+            <p className="text-xs sm:text-sm mb-1">
+              {`Word is that they're planning a raid as we speak so we recommend
+              you secure your progress on chain before continuing.`}
```
c-py, line 53: **suggestion (non-blocking):** This is a little word, perhaps it could be shortened. Here's my suggestion, what do you think?

```suggestion
              {`The Goblins have
              noticed that you are hoarding a lot of this resource off chain.`}
            </p>
            <p className="text-xs sm:text-sm mb-1">
              {`Please sync your progress before continuing, to secure your progress on chain.`}
```

It looks like this
hoard

c-py (approved):
Working well! I tested both Sunflower and SFL - The Modal popped up perfectly.

PR #1300, opened 2022-08-02
[FIX] Goblin Shovel Modal

adamhannigan:
# Description

The Goblin Shovel was not showing because the new `hoarding` action was not bubbling up the event and was throwing a state transition error when the Missing Shovel got thrown


vst212:
I see the problem still exists and it get trouble in the harvest, sure the bug is fixed?

cdxgg:
I see the problem still exists and it get trouble in the harvest, sure the bug is fixed?

cdxgg:
Version 5.7 does not fix this issue

PR #1304, opened 2022-08-03
[FIX] Goblin shovel thief modal

spencerdezartsmith:
# Description

This PR fixes an issue where the goblin modal wasn't showing. 

**Cause**

With our previous fix for the maxItems we needed to `processEvent` first (without a state update) to test whether the event would push maxItems over the threshold and if it didn't then we would `processEvent` again with a state update this time.

The incrementing `harvestCount` was being updated inside of the action which meant our first test `processEvent` was incrementing this count as well therefore for each harvest we were incrementing the `harvestCount` by two even though only one `item.harvested` event fired. 

The `GoblinShovel` was listening for the `item.harvested` event and when it fired it was running some logic to determine whether the shovel was stolen. 

**Example**

For the first steal when the harvest count was 1 before the first threshold the next harvest event runs the action once (maxItems test) which increments to the first threshold and then because we don't go over maxItems it immediately calls the action again which errors out on:

```
 if (isShovelStolen()) {
  throw new Error("Missing shovel!");
}
```
Therefore the `GoblinShovel` is never re-rendered so we don't get the modal showing. 

**Fix**

IMO the incrementing of the `harvestCount` shouldn't be inside the action. This is a technically a "side effect" of the action. I think we should aim to keep the actions pure. I have moved the `addToHarvest` call inside of the `GoblinShovel` component and invoked this when the `item.harvested` event is fired.

Fixes #issue

## Type of change

# Checklist:

### DIFF
--- a/src/features/farming/crops/components/GoblinShovel.tsx
+++ b/src/features/farming/crops/components/GoblinShovel.tsx
@@ -12,7 +12,10 @@ import { isShovelStolen } from "features/game/events/harvest";
 
 import { Button } from "components/ui/Button";
 import { GRID_WIDTH_PX } from "features/game/lib/constants";
-import { recoverShovel } from "features/game/lib/goblinShovelStorage";
+import {
+  addToHarvestCount,
+  recoverShovel,
+} from "features/game/lib/goblinShovelStorage";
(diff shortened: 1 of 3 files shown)


adamhannigan (approved):
LGTM 🚀

PR #1309, opened 2022-08-05
[FEAT] Expansion HUD

adamhannigan:
# Description

This PR introduces a new layout for land expansion HUD with:

- Bumpkin HUD
- Build CTA

It moves the previous 'Menu' actions into a modal and **assumes** autosave works in the background.

Screen Shot 2022-08-05 at 4 19 29 pm
Screen Shot 2022-08-05 at 4 19 37 pm


AL-Simmons92:
Good Stuff!!!!

spencerdezartsmith (approved):
Nice job.. code looks good. I'm not completely sold on the layout though. There are a few issues with it currently. 

- We need consistent margins around the viewport. The bottom cta's and the pp also have no margin on the x axis. We should make this the same as the balance. 
- `Settings` is a strange word for that cta. We can for sure change this though. Also the question mark icon doesn't really make sense here. 
- The title under the cta's should hide on mobile like the inventory cta does.
- I think we have too many different styles... the inventory label is centered against the cta but the bottom two are not. All the cta's have the white border but then the PP has a thin pixel border. 

I think it is fine to work with this now but I think we need to maybe mock this up a little more to find the right combination. I will approve so we can move forward 😄

adamhannigan:
@spencerdezartsmith Thanks for the feedback.

What are your thoughts on the grey button background?

Screen Shot 2022-08-08 at 8 53 01 am

adamhannigan:
Agree on all the feedback, especially around the aligning/wording/icons. 

Will need to find an icon that is better suited for Settings - the classic COG may work.

I will merge in for now so we can iterate on top

PR #1324, opened 2022-08-10
[FEAT] Fertilisers

0xsunflowerstudios:
# Description

This PR includes the first in-game fertiliser 🚀 Rapid Growth.

This was developed using TDD on live stream - check `fertiliseCrop.test.ts` for details.

Screen Shot 2022-08-10 at 3 50 48 pm

**Business Rules**

A rapid growth will halve the overall crop growing time. It can only be applied once per crop

Fixes #issue

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Enter UI only mode (remove `API_URL` from `.env`) and run `yarn dev`

Plant a potato seed and then apply the fertiliser.


c-py (dismissed):
Looking good!

I had a bit of difficulty seeing the fertiliser icon on the potato. The colour blends into the crop a little bit.

icon

Also do you think we should add the fertiliser to the on hover? It looks like we may be able to have multiple fertilisers in the future doing different things.

fert2

src/features/farming/crops/components/Soil.tsx
```diff
@@ -80,6 +81,13 @@ export const Soil: React.FC<Props> = ({
 
     return (
       <div className="relative w-full h-full">
+        {plantedCrop?.fertilisers && (
+          <div className="flex absolute z-10 right-1 -top-2">
```
c-py, line 85: I did a bit of playing, not sure what you think. 
```suggestion
          <div className="flex-col absolute z-10 - right-1 -top-2">
```
It moves the fertiliser out a little bit, and puts them in a column so they dont overlap the crop too much.

fertiliser

Polysemouse:
Great feature!

I agree that the icon should be moved to the on hover (on mobile users can click the crop to view the modal).  Currently the pixel size of the potion is smaller than that of the crop, and it makes the UI feels inconsistent.  Instead we can have a simple plus/boost icon (or any icon) or a highlight or any other simple indicators, and show all the available boosts in the modal.  As c-py has mentioned, there might be multiple boosts in the future

adamhannigan:
@Polysemouse Good call on the pixel size. We need to ensure consistency.

Screen Shot 2022-08-11 at 10 27 18 am


An area of anticipated confusion is when players cannot apply a fertiliser because it has already been applied. The red cross will show if the fertiliser is already applied. Putting fertilisers in the modal hides this information from users and could cause a bit of misunderstanding when the red cross appears

Polysemouse:
> Putting fertilisers in the modal hides this information from users and could cause a bit of misunderstanding when the red cross appears

That makes sense.  Perhaps putting the potion icon outside is the way to go for now.

The UI is looking much better with the pixel size consistency, thanks for updating that!

PR #1327, opened 2022-08-11
[FEAT] Connect land expansion HUD to buildings placement

spencerdezartsmith:
# Description

Co-authored by @c-py 41215134+c-py@users.noreply.github.com

This PR connects the land expansion hud to the building placement functionality. You can now click on the buildings button in the bottom right and select and place the fire pit.

We have placed the placement controller at the bottom of the screen to give clearer feedback to the player that we're in "editing" mode. We also added a "placed" animation on the building.

placement

I also refactored the building modal components. I found the `BuildingDetails` component to be hard to read so I have reworked it into 3 different components ie `Buildings` -> `ModalContent` ->  `ListView` + `DetailView`

**EXCLUDES** 
- Any persistence in regards to building placements. Once you place the building with just disappear 🪄 💨  

I have also extended the splash screen so that it shows while the session is loading as well. Prior to this the genesis land was always rendered while the session loaded. This was not a great experience, it was also the source of an annoying bug that prompted this change.

## Type of change

# How Has This Been Tested?

I have tested locally. You will be able to do the same and look at the loading screen as well as click on the buildings button in the Hud and place a building.

# Checklist:


src/features/farming/hud/components/PlaceableController.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const handleConfirmPlacement = () => {
+    send("PLACE");
+  };
+
+  const handleCancelPlacement = () => {
+    send("CANCEL");
+  };
+
+  return (
+    <div className="fixed bottom-2 left-1/2 -translate-x-1/2">
```
adamhannigan, line 31: Consider adding some higher z-indexes to this for small mobiles.

Screen Shot 2022-08-11 at 1 05 05 pm

adamhannigan: I'm wondering, should a user even be able to click the build hammer button icon?

It may be worth hiding the entire HUD while in edit mode

src/features/game/expansion/Game.tsx
```diff
@@ -80,12 +83,19 @@ export const Game: React.FC = () => {
...
+    gameState.context.sessionId === INITIAL_SESSION;
+
   return (
     <>
       <ToastManager />
 
-      <Modal show={SHOW_MODAL[gameState.value as StateValues]} centered>
-        <Panel className="text-shadow">
+      <Modal
+        show={SHOW_MODAL[gameState.value as StateValues] && !loadingSession}
```
adamhannigan, line 95: Instead of rendering this tree with some conditionals could we short circuit above and return early? It may help make it easier to read

E.g.

```
  const loadingSession =
    gameState.matches("loading") &&
    gameState.context.sessionId === INITIAL_SESSION;

if (loadingSession) {
    return <div className="h-s......
}

return (
   <>
      <ToastManager />
      ...
```

src/features/game/expansion/Game.tsx
```diff
@@ -100,9 +110,36 @@ export const Game: React.FC = () => {
...
       </Modal>
 
-      <Water level={gameState.context.state.expansions.length + 1} />
-      <Land />
-      <Hud />
+      {/* Show splash screen while session is loading  */}
+      {loadingSession ? (
+        <div className="h-screen w-full fixed top-0" style={{ zIndex: 1050 }}>
+          <Splash fadeIn={false} />
+          <Modal show centered backdrop={false}>
```
adamhannigan, line 117: There is a noticeable flash when showing this modal. Not a blocker, but do you have any ideas how we can prevent that?

src/features/island/buildings/components/DetailView.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const DetailView: React.FC<Props> = ({
+  state,
+  building,
+  onBack,
+  onBuild,
+}) => {
+  const { bumpkin, inventory } = state;
+
+  const canBuild = (building: BuildingName) => {
```
adamhannigan, line 32: Should this function be called `cantBuild`

It is returning `return missingIngredients && missingBalance;` so assume the function returns true when you don't have the required ingredients

adamhannigan (approved):
Looks good to me! Just some minor feedback

src/features/game/expansion/Game.tsx
spencerdezartsmith: 100%! This was actually what I had but then got stuck on the dumbest bug that lead to many code changes 😅 .  Will update.

src/features/game/expansion/Game.tsx
spencerdezartsmith: This is happening when the machines switch over. Might have an idea on how to fix this.

spencerdezartsmith: No was unable to fix this. I can circle back at a later stage when I have more time.

PR #1330, opened 2022-08-11
[FIX] Reveal Event

adamhannigan:
# Description

"expansion.revealed" is a UI only event. This PR excludes it from going to the API and causing an error.

## Type of change

# How to test?

1. `yarn dev`
2. Expand your land
3. Wait for the autosave (no error should occur - the event will be filtered out)


Rofrtd (approved):
Working well.. 
LGTM :shipit:

PR #1348, opened 2022-08-18
[FEAT] Community Garden

adamhannigan:
# Description

The purpose of this PR is to introduce a `community` area to the game that can contain community content. The goal is that community developers can iterate quickly and ship features without affecting the core code.

Inside of a community context, balance + items will be loaded using the **personal wallet** instead of the farm wallet. This makes the HUD show the actual balance in their personal wallet.

All new content now lives in the `/src/features/community` directory.

There is no CTA to access the community route. To test this you can load the game then put `http://localhost:3000/#/community-garden` into the URL

**Excludes**

Implementation of community features.

Screen Shot 2022-08-18 at 12 26 43 pm


src/features/community/components/CommunityHUD.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Context } from "../lib/CommunityProvider";
+import { useActor } from "@xstate/react";
+
+export const CommunityHud: React.FC = () => {
+  const { communityService } = useContext(Context);
+  const [state] = useActor(communityService);
+
+  return (
+    <div aria-label="Hud">
+      <Balance balance={state.context.balance} />
```
spencerdezartsmith, line 13: I think this will be very confusing for players. I think we should have a different balance display that includes a metamask logo or something like that. Something that clearly indicates that we're showing a balance that is different for the in game balance.

spencerdezartsmith (approved):
LGTM. Just one comment about the balance. It's not a blocker for now but I think should definitely be addressed before this is released.

src/features/community/components/CommunityHUD.tsx
adamhannigan: Agreed. We can tidy up a few things pre-launch

PR #1394, opened 2022-09-06
[FEAT] Bumpkin URI detection

adamhannigan:
# Description

Use the Bumpkin Token URI for change detection. This is because users can equip/unequip at Bumpkin.io and need to stay in sync.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Run against local API


c-py (approved):
LGTM!

PR #1459, opened 2022-09-30
[FEAT] CloudFlare Captcha

adamhannigan:
# Description

Introduce the CloudFlare Captcha to the Sync action.

This is much more seamless than the Recaptcha and adds a new anti-botting mechanism botters will need to solve 👌 

Screen Shot 2022-09-30 at 10 29 11 am


spencerdezartsmith (approved):
Excited to try it!

PR #1484, opened 2022-10-06
[FEAT] Chicken placement

adamhannigan:
# Description

Place your chickens where you desire!

This PR includes a new `Chicken.tsx` component with duplicated images for backwards compatibility.

The `Placeable` controller has also been updated to accept components, instead of just images. This helps with consistent styling and placement.

**Merge**

Must merge this in at the same time as the API

Screen Shot 2022-10-06 at 4 26 43 pm
Screen Shot 2022-10-06 at 4 26 47 pm
Screen Shot 2022-10-06 at 4 26 50 pm

### How to test?

You will need to run the corresponding API PR to get access to the newest events.

Right now the Chicken House is hardcoded to be cheap and attainable at level 2. We will change this later 👍 

**TODO**

We need a new "placed" effect on an item, since the 'bulge' effect was not compatible with components.


mGranTY:
what happens when you place a Chicken House, place the chickens, then remove the chicken house?

adamhannigan:
@mGranTY Ideally if you remove a Chicken Coop, it would remove X amount of chickens as well 😄

spencerdezartsmith (approved):
Working well for me

PR #1507, opened 2022-10-12
[FEAT] Game Rules

adamhannigan:
# Description

For new players, include message around anti-botting and 1 account per player. This will show every 7 days.

Screen Shot 2022-10-12 at 2 29 12 pm

I also simplified the machine logic with a single 'notifying' state to handle the different message states. There was a bug where some of these states would acknowledge and then go straight into playing state, instead of checking for other messages.


Rofrtd (approved):
Nice one mate.

LGTM. 🚀

mGranTY (approved):

PR #1616, opened 2022-11-03
[FIX] Selling multiple crops

HBandesh:
# Description

The users were not able to sell the crops when the amount was in decimals. 

Fixes #issue

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/features/game/events/landExpansion/sellCrop.ts
+++ b/src/features/game/events/landExpansion/sellCrop.ts
@@ -5,6 +5,7 @@ import cloneDeep from "lodash.clonedeep";
 import { getSellPrice } from "features/game/lib/boosts";
 import { SellableItem } from "../sell";
 import { trackActivity } from "features/game/types/bumpkinActivity";
+import { setPrecision } from "lib/utils/formatNumber";
 
 export type SellCropAction = {
   type: "crop.sold";
@@ -55,7 +56,7 @@ export function sellCrop({ state, action }: Options): GameState {
     balance: game.balance.add(sflEarned),
     inventory: {
       ...game.inventory,
-      [sellables.name]: count.sub(amount),
+      [sellables.name]: setPrecision(count.sub(amount)),
     },
   };
 }
--- a/src/features/island/buildings/components/building/market/Crops.tsx
+++ b/src/features/island/buildings/components/building/market/Crops.tsx
@@ -17,6 +17,7 @@ import { Modal } from "react-bootstrap";
 import { ITEM_DETAILS } from "features/game/types/images";
 import { ToastContext } from "features/game/toast/ToastQueueProvider";
 import { getSellPrice, hasSellBoost } from "features/game/lib/boosts";
+import { setPrecision } from "lib/utils/formatNumber";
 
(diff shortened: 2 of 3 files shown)


c-py (approved):
LGTM!!!

AzaroQL:
Wont we end up in state where 0 is displayed as amount and we wont be able to sell it?
image

Polysemouse:
> Wont we end up in state where 0 is displayed as amount and we wont be able to sell it? image

Yes we would have some crypto, I mean crop, dust left behind in our farms.  I think it's better to tackle the root solution so we don't need a workaround (or implement the workaround until the root solution is found)

HBandesh:
> Wont we end up in state where 0 is displayed as amount and we wont be able to sell it? image

Thanks @AzaroQL for sharing the screenshot. In order to understand the issue better can you please share the steps to reproduce this scenarios. How did you reach at the zero amount? And what message are you getting when you are making that transaction?

HBandesh:
> > Wont we end up in state where 0 is displayed as amount and we wont be able to sell it? image
> 
> Yes we would have some crypto, I mean crop, dust left behind in our farms. I think it's better to tackle the root solution so we don't need a workaround (or implement the workaround until the root solution is found)

@Polysemouse , this is not an issue with the language itself. The issue that is how the computers understand the numbers in 0's and 1's. In order to perform operations, the computers first compress the numbers using IEEE 754 algorithm. This algorithm converts the numbers to 64 bits. This compression algorithm has its own drawbacks. So kind of its not really a fault when we perform 0.1 + 0.2 and we get 0.30000004. It is simply because the algorithm is not designed to represent this number.

Even if we try other libraries to handle numbers in JS, they will ultimately have this problem as they are built on the JS itself.

There is a backend fix to this PR as well, which performs the exact same operation in backend. Over the time all the amounts (product count and not SFL) in the backend would also be replaced by the 4 digits of precision and the issue wont happen.

Polysemouse:
@AzaroQL @HBandesh sounds like a tricky issue with JS numbers.  We can modify the <Box> component so the display is hidden when the item amount is less than 0.0001.  That way we will still have crypto dust in our farms, but it won't be shown in your farms

AzaroQL:
> @AzaroQL @HBandesh sounds like a tricky issue with JS numbers.
> 
> The 0 amount is shown because it is a rounded off number from the actual amount (eg. 0.0000999). The display is currently shown if the amount is greeter than 0, so that's why a 0 is shown.
> 
> We can modify the component so the display is hidden when the item amount is less than 0.0001. That way we will still have crypto dust in our farms, but it won't be shown in your farms

Having invisible crops dust is fine, but I would propose to fix it on BE still. Trim crop values to 4 decimals before sending those to FE, so we wont have any issues down the line.

image

HBandesh:
@AzaroQL @Polysemouse , as stated in my last comment, there is a BE fix against this PR as well. After both are merged, once you perform the first transaction of the crop, the amount will be corrected on BE as well which would be propagated to the blockchain after sync.

Polysemouse:
@HBandesh Looks good, I will test them and see how it goes

PR #1639, opened 2022-11-09
[FEAT] Add mintCollectible api in preparation for Auctioneer

spencerdezartsmith:
# Description

In preparation for the Land Expansion Auctioneer we have updated the mint method on our contract. We will now mint items through the `mintCollectible` method which will be able to accommodate flexible releases i.e. split releases, dutch auctions.

This PR sets up the codebase to call this function on testnet only for now. I have added the new actions as well as the new types for the `Session` contract.

The only way to test this mint for now is through remix.

Screen Shot 2022-11-09 at 5 09 41 pm

https://mumbai.polygonscan.com/tx/0x318b88b6f623d08e1c44044ff0b064603a2d36f3340cfec8f580139515cdcfa2

**NOTE** This updates the Session Contract so when merging we will need to inform the dev community.

# How Has This Been Tested?

# Checklist:


src/lib/blockchain/abis/Session.json
```diff
@@ -119,28 +119,18 @@
       {
         "components": [
           { "internalType": "uint256", "name": "mintId", "type": "uint256" },
-          {
```
adamhannigan, line 122: It gets a bit tricky maintaining 2 versions of the same ABI.

In this case here, it is likely that mainnet will throw an error when using this ABI as it no longer aligns to the actual shape of the data coming back from the blockchain.

We have a couple options here:

1. We maintain 2 Session ABIs in the repo and use based on what network we are on. `SessionMainnet.json` & `SessionTestnet.json`
2. We change the name of the `Recipe` struct in the new contract so that we can keep all of the old ABI structures that are used for minting/recipes on mainnet

src/lib/blockchain/Sessions.ts
```diff
@@ -122,18 +133,6 @@ export class SessionManager {
       // For UI purposes, do not show the wei values
       const ethBasedRecipes = recipesWithIds.map((recipe, i) => ({
         ...recipe,
-
-        tokenAmount: recipe.tokenAmount
```
adamhannigan, line 126: Is this needed for mainnet?

adamhannigan (commented):
Nice flows mate 👌 

Just a couple questions regarding the mainnet compatibility

adamhannigan (dismissed):
LGTM 🚀

PR #1653, opened 2022-11-11
[FIX] Enable Sell 1 Button when Selling Fractional Crop

ShiningMinions:
# Description

When selling a crop less that 1, the "Sell 1" button was disabled. This felt a bit awkward to me, so in this PR I added the ability to click this button which sells the remaining amount of crops less than one. It also renders the text "<1" on the button in the scenario you have less than one to sell.

Finally, the reason I didn't make this text the actual fractional crop amount:  Theoretically it can get to be a really large decimal number (i.e. like `0.123456781234567812`).I thought a simple "less-than" char summarized it just fine, without hte risk of the text overflowing.

BugBash Ticket # 48

Before:
bug

After:
fix

## Type of change

# How Has This Been Tested?

Visually

# Checklist:


adamhannigan (approved):
Agreed. Nice addition 👍

PR #1693, opened 2022-11-18
[FIX] Show Boosted Food Exp

j923sneaks:
# Description

This PR shows the boosted food exp value if Bumpkin obtains Kitchen Hand skill.

Fixes #issue

## Type of change

# How Has This Been Tested?

Manual test - local

Kitchen Hand

https://user-images.githubusercontent.com/89294757/202592216-a689da57-c8e3-443c-b4e9-c71d3c45014b.mp4

None

https://user-images.githubusercontent.com/89294757/202591749-9fe658de-b04b-4c99-b8d8-c83f2ebbf8b0.mp4

# Checklist:

- Any dependent changes have been merged and published in downstream modules


src/features/game/expansion/lib/boosts.ts
```diff
@@ -108,3 +108,24 @@ export const hasBoost = ({ item, collectibles }: HasBoostArgs) => {
...
+  foodExp: number,
+  bumpkin: Bumpkin | undefined
+): number => {
+  let boostedExp = new Decimal(foodExp);
+
+  if (bumpkin?.skills["Kitchen Hand"]) {
+    boostedExp = boostedExp.mul(1.1);
+  }
+
+  return boostedExp.toNumber();
```
mGranTY, line 130: i think its better to return a decimal and work with it later.

src/features/island/bumpkin/components/Feed.tsx
```diff
@@ -80,7 +81,10 @@ export const Feed: React.FC<Props> = ({ food, onClose, onFeed }) => {
                 <div className="flex justify-center flex-wrap items-center">
                   <img src={heart} className="me-2 w-6" />
                   <span className="text-xs text-shadow text-center">
-                    {`${selected.experience} EXP`}
+                    {`${getFoodExpBoost(
```
mGranTY, line 84: can we show the [⚡](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwjZ0YSl7bb7AhUzK7kGHSYuAj8QFnoECAoQAQ&url=https%3A%2F%2Fpsfonttk.com%2Fdecoratetext%2Fraio-para-nick-free-fire.html&usg=AOvVaw0gQ-0BkvEYvWnTbNspDdxY) hasBoost icon?
image

src/features/game/expansion/lib/boosts.ts
j923sneaks: I tried to avoid changing the current `number` type declaration for bumpkin experience as it may affect other components i havent checked as well as backend stuff.

or is your suggestion to return Decimal and just call `toNumber()` after function call on each component that uses it?

mGranTY: > I tried to avoid changing the current `number` type declaration for bumpkin experience as it may affect other components i havent checked as well as backend stuff.
> 
> or is your suggestion to return Decimal and just call `toNumber()` after function call on each component that uses it?

@j923sneaks if all bumpkin exps is a Number then just let code as it is. 

I think everything should be a Decimal, so we can avoid future errors, idk why it its a number, maybe a dev can explain to us.

src/features/island/bumpkin/components/Feed.tsx
j923sneaks: just wanna confirm with @adamhannigan and @spencerdezartsmith if we should continue displaying this icon on every boost. 

At first it was fine because there were few boosts from few sources but now it's kinda hard to maintain and be consistent. Does it really add value as Spencer mentioned here

image

Should we follow through, some screens might get overcrowded like in Feed where exp and time are boosted ie will we put lightning icon on both rows?

image

adamhannigan (approved):
LGTM. We can clean up any remaining lightning bolts in a separate PR 🚀

PR #1721, opened 2022-11-20
[CHORE] increase flag size by 50%

AzaroQL:
# Description

Currently flags look a bit off, too small.

My change alters pixel size of flags, but I think it still looks better than with small flags. Best approach is to redraw flags on a bigger scale, but I doubt someone will do that.

I made intentionally so flag still takes 1 grid cell, but is displayed in 2 cells, that way it can wave above the island if placed on top row. Players can decide where to put the flag and if it interfere with other items on map they can reposition it. But it seems that all interactions work even with odd flag placements.(see the video).

Before:
image

After:
https://user-images.githubusercontent.com/9656961/202896575-1d7adf2e-726a-4cf6-a3f0-09de825c7a7a.mp4

## Type of change

# How Has This Been Tested?

Manual test

# Checklist:

### DIFF
--- a/src/features/island/collectibles/components/Flags.tsx
+++ b/src/features/island/collectibles/components/Flags.tsx
@@ -15,22 +15,16 @@ interface Props {
 
 const FlagsImages: React.FC<Props> = ({ flagName }) => {
   return (
-    <div
-      className="absolute"
+    <img
+      className="absolute max-w-none hover:img-highlight"
       style={{
-        width: `${PIXEL_SCALE * 11}px`,
-        right: `${PIXEL_SCALE * 2.5}px`,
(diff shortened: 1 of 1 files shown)


Polysemouse:
Increasing the flag size would cause a mismatch to the pixel size.  What if the flag is moved upwards a few pixels instead of enlarging it?

Rofrtd:
Thanks @AzaroQL, but as @Polysemouse mentioned this would cause a mismatch to the pixels, and we need to keep them same sizes. And the sizing you added would be covering the above item, for example the Stone rock as per your video, is quite hard to see it behind the flag.

AzaroQL:
Yep, as I mentioned only correct way to address this is to redraw all the flags, but thats an enormous task..

Feel free to close this PR.

PR #1725, opened 2022-11-21
[FIX] Chickens component was calling old actions

AzaroQL:
# Description

Chickens were using old actions which don't take in to account that we are in LE and collectibles should be placed and bumpkin should be in wallet.

## Type of change

# How Has This Been Tested?

Manual test + yarn test

# Checklist:

### DIFF
--- a/src/features/game/events/index.ts
+++ b/src/features/game/events/index.ts
@@ -37,6 +37,10 @@ import {
 } from "./landExpansion/ironMine";
 
 import {
+  collectEgg as LandExpansionCollectEgg,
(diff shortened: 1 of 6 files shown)


spencerdezartsmith (changes requested):
This will require some back end work before we can move forward. I am going to assign "Request Changes" to this just so we know we need to do something. Will get to it asap.

spencerdezartsmith (commented):
Thanks for this 😃 

We chatted about this and have decided to leave this for now. The current action is ok. We will probably leave this refactor until after release.

PR #1743, opened 2022-11-23
[FIX] Block goblin village for LE migrants. Block goblin retreat if level not met.

spencerdezartsmith:
# Description

This PR blocks the old Goblin Village from being accessed by players who have migrated to Land Expansion. This was happening when people were deep linking from other sites/loading url etc. 

Screen Shot 2022-11-23 at 2 28 14 pm

I have also blocked access to Goblin Retreat for players who don't meet the level requirements. Players who were already loaded into LE were able to just swap out the url path and access even if they weren't at Level 5. 

I have also fixed a small bug when getting a connecting error on the home screen. The button was not doing anything because it was always pointing to the gameMachine which wasn't loaded in that instance. I have set it to conditionally use that machine.

Fixes #[issue](https://discord.com/channels/880987707214544966/1043179321159852155/1043179321159852155)

## Type of change

# How Has This Been Tested?

- Test with an account that doesn't have Level 5. Load LE and then swap `/land/:id` with `/retreat/:id`. You should be sent back to where you came from.
- Confirm that Level 5 account can still access the retreat.
- With a migrated account, try to load `..../#/goblins` directly from the url bar and you should not be allowed to access the village.
- Check to make sure an unmigrated account can access old goblin village.

# Checklist:

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -62,8 +62,8 @@ export const Navigation: React.FC = () => {
         <HashRouter>
           <Routes>
             <Route path="/" element={<Humans />} />
-            {/* Forbid entry to Goblin Village when in Visiting State, show Forbidden screen */}
-            {!authState.matches("visiting") ? (
+            {/* Forbid entry to Goblin Village when in Visiting State or when a player has migrated to LE, show Forbidden screen */}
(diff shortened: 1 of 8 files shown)


HBandesh (approved):
Great work! 🥳🥳🥳

PR #1750, opened 2022-11-24
[FIX] Update expansions with server data after autosave

spencerdezartsmith:
# Description
An issue was raised were sometimes people were seeing tree drops of 3. This was happening when a person didn't refresh during the time it took for the tree to replenish and then chopping again. On the client side event we set a hardcoded placeholder value for the next drop from that tree. We set that value to 3. This "should" have been overridden by the real server value on the next autosave however we weren't updating these values 😱 

This PR fixes that issue by updating land expansion resource data in the `updateGame` function.

Fixes #[issue](https://discord.com/channels/880987707214544966/1043856744918564904/1043856744918564904)

## Type of change

# How Has This Been Tested?

Tests

# Checklist:

### DIFF
--- a/src/features/game/lib/transforms.test.ts
+++ b/src/features/game/lib/transforms.test.ts
@@ -1,6 +1,7 @@
 import Decimal from "decimal.js-light";
+import { GameState, LandExpansion } from "../types/game";
 import { TEST_FARM } from "./constants";
-import { getLowestGameState } from "./transforms";
+import { getLowestGameState, updateExpansions } from "./transforms";
 
 describe("transform", () => {
   it("gets the lowest balance from the first object", () => {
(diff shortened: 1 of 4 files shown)


src/features/game/lib/transforms.test.ts
```diff
@@ -58,4 +59,386 @@ describe("transform", () => {
       Axe: new Decimal(90),
     });
   });
+
+  describe.only("updateExpansions", () => {
```
c-py, line 63:
```suggestion
  describe("updateExpansions", () => {
```

c-py (approved):
Works perfectly mate, I temporarily changed the placeholder amount to 99, and could see the state update twice. Once after frontend tree chopped and again once backend gave real value.

c-py (approved):

PR #1737, opened 2022-11-22
[FIX] Stock label position

JacobSobolev:
# Description

1. The stock label isn't centered like the other elements 
2. overlaps with the inner panel.
3. too close to the text under it

| Before  | After |
| ------------- | ------------- |
| image  | image |

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

locally

`yarn test` on the current branch
image

# Checklist:

- [x] Title of the PR is relevant and is prefixed with [FEAT], [CHORE] or [FIX]
- [x] Screenshot if it includes any UI changes
- [x] I have read the contributing guidelines and agree to the T&Cs
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1761, opened 2022-11-25
[CHORE] Label Standardization

hujikol:
# Description

- Add a new Label component for the label format type
  - `default`: silver
  - `success`: green
  - `info`: blue
  - `danger`: red
- Change all labels to using the Label component
- Change rare shop label position based on #1737 
- Change Blacksmith and seed shop out of stock description
- Add a new label on the seed shop when "No Space Left" is Happening.

## Type of change

# How Has This Been Tested?

## UI Inspection

| Number | Before | After |
| ------------- | ------------- | ------------- |
| 1 | image | image |
| 2 | image | image |
| 3 | image | image |
| 4 | image | image |
| 5 | image | image |
| 6 | image | image |
| 7 | image | image |
| 7 | Before | image |
| 8 | image | image |
| 8 | image | image |

## UX Inspection

https://user-images.githubusercontent.com/12151051/204043019-b2ac86f1-465f-4ddf-831a-61d0bb07c8ab.mp4

# Checklist:

- [x] Title of the PR is relevant and is prefixed with [FEAT], [CHORE], or [FIX]
- [x] I have commented on my code, particularly in hard-to-understand areas


Polysemouse:
Seems like there are comple UI warnings in your video.  When you are trying to access variables in the game state, you need to make sure the game state is `initialized` first.

hujikol:
> Seems like there are comple UI warnings in your video. When you are trying to access variables in the game state, you need to make sure the game state is `initialized` first.
> 
> @hujikol You can see how it is done here: https://github.com/sunflower-land/sunflower-land/pull/1759/files#r1032680140

hmm.. it seems I can't find the root cause on what file, are those states always being consumed even when the building is not being placed?

Polysemouse:
In your PR you are trying to access state variables in src/features/island/buildings/components/building/market/Seeds.tsx and src/features/island/buildings/components/building/workBench/components/WorkbenchModal.tsx.  Since these variables are used in the JSX components, they are called every time as long as the component is rendered in the code.  The component might be just initializing while that happens, thus the warnings in the console.

Therefore you need to check if the state machine is initialized before rendering the whole thing, else just return an empty object to avoid calling those state variables

hujikol:
still can't resolve the issue locally, but test it while using the testnet API got nothing in the console

image

and there is me console logging the `!gameService.initialized` from the beginning of the code, always returning `false`, that means it has been initialized from the beginning.. or am I missing something there... 🤔

Polysemouse:
Try something like this for any components that are going to be rendered at the root of the parent component:
```
if (!gameService?.initialized) {
  // Render a view that does not need to access game service variables here
  return <.../>
}

return (
  // Render view with the game service variables
  // eg. If "stock" is obtained from the game service
  <...
    {stock?.someProperty && <.../>
    ...
   .../>
)
```

src/components/ui/Label.tsx
```diff
@@ -4,25 +4,61 @@ import { pixelWhiteBorderStyle } from "features/game/lib/style";
 
 interface Props {
   className?: string;
+  type?: string;
 }
-export const Label: React.FC<Props> = ({ children, className }) => {
+export const Label: React.FC<Props> = ({ children, className, type }) => {
```
spencerdezartsmith, line 9: How about you set a default value for `type` so you don't have to pass the `default` prop?

`React.FC<Props> = ({ children, className, type = "default" })`

spencerdezartsmith (approved):
Love this. Very clean UI. I have left one non blocking comment.

src/components/ui/Label.tsx
hujikol: I have doubts about this decision because I think the default value should be used for items that are used frequently, while we use the "info" (blue) value more than the "default" (silver) itself, or I think we can swap default for the blue one then create another type for the silver one?

PR #1777, opened 2022-11-28
[FIX] Skill points indications

JacobSobolev:
# Description

In the bumpkin model, the green arrow was showing up all the time.
1. now the green level-up arrow shows up only if you have unspent skill points.
2. the alert icon near the skills panel wasn't showing up if you had unspent.

I found out about the second point after I fixed the first one.
now when I think about it there are 2 indications for the same thing.
maybe I should remove one of them?

**update**: only the alert indication is showing if there are unspent skill points

| State | Screen Shot | 
| ------------- | ------------- |
| Before | image | 
| After fix (before using spending the points )| image | 
|  After fix (after spending the points) | image |

## Type of change

# How Has This Been Tested?

1. spending all the points - remove all the indication
2. spending some of the points - the indication still shows up

`yarn test`
image

the main branch had the same result

# Checklist:

### DIFF
--- a/src/features/bumpkins/components/BumpkinModal.tsx
+++ b/src/features/bumpkins/components/BumpkinModal.tsx
@@ -127,15 +127,13 @@ export const BumpkinModal: React.FC<Props> = ({ initialView, onClose }) => {
         <div className="flex-1">
           <div className="mb-3">
             <div className="flex items-center ml-1 my-2">
-              {hasAvaliableSP && (
-                <img
-                  src={levelIcon}
-                  style={{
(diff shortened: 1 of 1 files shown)


Polysemouse:
The green up arrow is the icon for level, I think it should be there all the time.
Though I often get confused with the meaning of the green up arrow.  I believe we need a more suitable icon for level

JacobSobolev:
@Polysemouse I see, thanks for the clarification, I always thought it was an indication of unspent skill points.
So I will remove that condition on the green arrow and keep the alert on the skills

c-py (approved):
Nice touch! Thank you for that!

PR #1803, opened 2022-12-01
[FIX] Fixing snow kingdom illegal access.

HBandesh:
# Description

The players were able to access the snow kingdom without having required level. 

Fixes #issue

## Type of change

# How Has This Been Tested?

# Checklist:


spencerdezartsmith (changes requested):
Nice one. I can still actually load the route it just doesn't show me the map if I don't have the correct level.

I think a better UX would be to redirect back to `/land` if I am not allowed. Otherwise I am just stuck there and need to use the browser back button to get back to land.

HBandesh:
> Nice one. I can still actually load the route it just doesn't show me the map if I don't have the correct level.
> 
> I think a better UX would be to redirect back to `/land` if I am not allowed. Otherwise I am just stuck there and need to use the browser back button to get back to land.
> 
> Screen Shot 2022-12-02 at 12 36 01 pm

Actually, this is what currently happens for other islands like treasure island, stone heaver etc. As far as I remember from my conversation with Adam, it is done to give a reflection that they have landed in the middle of nowhere by accessing that URL. But I understand your point that the user is left with no feedback or action.

I tried force redirecting the user back to land but I guess the better UI would be to show the user and modal give a CTA to travel back home. Which is what I have implemented now. This will work on all the gibberish URL's as well.
image

spencerdezartsmith:
> > Nice one. I can still actually load the route it just doesn't show me the map if I don't have the correct level.
> > I think a better UX would be to redirect back to `/land` if I am not allowed. Otherwise I am just stuck there and need to use the browser back button to get back to land.
> > Screen Shot 2022-12-02 at 12 36 01 pm
> 
> Actually, this is what currently happens for other islands like treasure island, stone heaver etc. As far as I remember from my conversation with Adam, it is done to give a reflection that they have landed in the middle of nowhere by accessing that URL. But I understand your point that the user is left with no feedback or action.
> 
> I tried force redirecting the user back to land but I guess the better UI would be to show the user and modal give a CTA to travel back home. Which is what I have implemented now. This will work on all the gibberish URL's as well. image

Yeah this makes sense. Can we change the wording to just say `You have landed in the middle of nowhere!`? Also can we align the modal padding with the standard. Here is an example. I believe there is no padding around the button but everything else has `p-2`. After that we're good to go 🚀  
Screen Shot 2022-12-05 at 2 27 51 pm

spencerdezartsmith (approved):
Approving for now. Just need to align the styling then you can merge. Awesome job!

PR #1805, opened 2022-12-02
[CHORE] Better Bank Withdrawing Overhaul

### LINKED ISSUE #1355: [CHORE] Localize Bank Withdraw Conditions & Not Withdrawable by Default
## Summary

I am constantly seeing the following bugs appear over and over again:
1. Item shouldn't be withdraw-able, but is.
2. Item should be withdraw-able, but is not.

This is due to the ever-growing list of niche withdraw conditions based on the player's current farm state. Therefore, I recommend we rethink this function. Here is how the function is currently lookng:

`bankUtils.ts`

```ts
export function canWithdraw({ item, game }: CanWithdrawArgs) {
  // Coming soon
  if (isSeed(item)) {
    return false;
  }

  // Coming soon
  if (item in SKILL_TREE) {
    return false;
  }

  // Coming soon
  if (item in FOODS()) {
    return false;
  }

  // Quest item
  if (item in QUEST_ITEMS) {
    return false;
  }

  if (item === "Engine Core" || item == "Observatory") {
    return false;
  }

  if (item in SHOVELS) {
    return false;
  }

  // Make sure no trees are replenishing
  if (
    item === "Woody the Beaver" ||
    item === "Apprentice Beaver" ||
    item === "Foreman Beaver"
  ) {
    return Object.values(game?.trees).every((tree) => canChop(tree));
  }

  if (item === "Kuebiko" && hasSeeds(game.inventory)) {
    return false;
  }

...etc...

}
```
One major reason this bug keeps reoccurring is that **an item is withdraw-able by default**. Instead, we should flip this logic by stating: **an item is NOT withdraw-able, UNLESS explicitly allowed**.

I recommend that instead of one monolithic withdraw function, we somehow store this localized logic on the item level, using some sort of lookup process. Perhaps this would be best extending the "CraftableItem" type in `features/game/types/craftables.ts`, or something like this:

```ts
export interface CraftableItem {
  id?: number;
  name: CraftableName;
  description: string;
  tokenAmount?: Decimal;
  ingredients?: Ingredient[];
  disabled?: boolean;
  requires?: InventoryItemName;
  isPlaceholder?: boolean;

  // NEW FIELD HERE - This means by default, NOTHING is withdrawable UNLESS I EXPLICITLY write it here.
  // This would save a lot of bugs from occurring by default.
  canWithdraw?: boolean | (gameMachine, goblinMachine) => boolean;
}
```
In this way, our `canWithdraw()` function can be much simplified into something like the following:

```ts
export function canWithdraw(item: CraftableItem) {
 const canWithdraw: boolean = typeof item.canWithdraw === 'function' ? item.canWithdraw() : !!item.canWithdraw;
 return canWithdraw;
}
```

## TODO

1. Implement some localized approach for detecting if an item is withdraw-able or not. See above example of an approach which can be taken.
2. Ensure that an item is NOT withdrawable by default UNLESS explicitly allowed.
3. Ensure bank withdraw tests are up to date and no existing tests break.

ShiningMinions:
# Description

This PR overhauls the current bank withdraw logic in the codebase. The goal of this PR is to greatly reduce the likelihood  of bank withdraw bugs from continuously popping-up (as they do now).

Before PR:
In the old code, everything was withdrawable by default (unless explicitly disallowed).

After PR:
In this PR, everything is NOT withdrawable by default (unless explicitly allowed).

Having this logic flipped is a much better approach concerning impact. In old code, if developer forgets to add withdraw condition (which is common), the impact is that the game/economy will be exploited. In new code, if developer forgets to add withdraw condition, the impact is that players simply can't withdraw for a brief moment.

There are several other withdraw logics and/or tests that were missing from the old code. You can see these new rules reflected in new tests. These include (but not limited to):
- Victoria Sisters: Shouldn't withdraw if pumpkins growing
- Rooster: Shouldn't withdraw if any chicken fed
- Carrot sword: Shouldn't withdraw if any crops growing
- Rock Golem: Shouldn't withdraw if any stones replenishing

Fixes #1355 

## Type of change

Chore

# How Has This Been Tested?

Added many new automated tests, and re-using all tests from old code.

# Checklist:


ShiningMinions:
I have 2 asks before we merge this PR:

1. Can we allow the observatory to be withdrawn now? I didn't enable this yet but I think now would be a good time.
2. I'm also going to ask other community members to audit this and ensure I didn't miss anything.

spencerdezartsmith:
> I have 2 asks before we merge this PR:
> 
> 1. Can we allow the observatory to be withdrawn now? I didn't enable this yet but I think now would be a good time.
> 2. I'm also going to ask other community members to audit this and ensure I didn't miss anything.

Yes we can allow the observatory to be withdrawn.

c-py (approved):
Fantastic addition! Thank you so much @ShiningMinions !

PR #1814, opened 2022-12-03
[FIX] Loading Text on Helios Travel

ShiningMinions:
# Description

When clicking the boat on Helios, an autosave gets triggered (which is working as intended). However, the problem is that there is no "Loading..." text blocking the view until autosave completes (like it does on player's island). Without this blocking UI, the player can technically travel faster than the autosave completes, resulting in lost gameState progress from helios.

## Type of change

# How Has This Been Tested?

I purchased white tulips on helios then immediately clicked the boat. I confirmed I received a "Loading..." text, similar to how it works on player's island.

# Checklist:


c-py (approved):
Works great! Thanks for your contribution @ShiningMinions !

PR #1838, opened 2022-12-08
[FIX] Add rusty shovel back to workbench

AzaroQL:
# Description

Added rusty shovel back to workbench, so players who built it dont have to go to Helios.
image
image

CRAFTABLE_TOOLS, WORKBENCH_TOOLS and TREASURE_TOOLS converted to objects to enable correct type checks. Previously we had name field that was not available in Tool interface and we had Sand Shovel in workbench items while it was not in WorkbenchToolName type. Now typescript highlights all the errors in those objects.

## Type of change

# How Has This Been Tested?

Manual test, yarn test.

# Checklist:

### DIFF
--- a/src/features/game/events/landExpansion/craftTool.test.ts
+++ b/src/features/game/events/landExpansion/craftTool.test.ts
@@ -64,7 +64,7 @@ describe("craftTool", () => {
     });
 
     expect(state.balance).toEqual(
-      new Decimal(1).minus(WORKBENCH_TOOLS()["Axe"].sfl)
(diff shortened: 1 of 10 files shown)


adamhannigan:
@AzaroQL I agree, it is a bit of friction needing to go to Helios to buy that Shovel when you want to dig something up.

While we are rearranging these I would also suggest removing the Sand Shovel from that list - lots of players accidentally buy it at the Workbench not realising they can use it until Treasure Island.

As a general rule I feel that tools should be available on the island that they are used on 💯

AzaroQL:
Done, it actually looks much better code wise without treasure tools in workbench.

PR #1878, opened 2022-12-15
[FEAT] Bumpkin Image Builder

adamhannigan:
# Description

Instead of building a Bumpkin on the front-end using small parts, rely on the images that are generated on the backend (how Bumpkin NFTs are built).

This solves a few problems:

- Less flashing Bumpkin parts
- Less image requests
- Support Bumpkin masking (can finally put Hats + Onesies on Bumpkins 🚀 )


src/features/auth/useImagePreloader.tsx
```diff
@@ -26,6 +26,7 @@ import retreat from "assets/land/retreat.webp";
 import treasureIsland from "assets/land/treasure_island.webp";
 import snowKingdom from "assets/land/snow_kingdom.webp";
 import stoneHeaven from "assets/land/stone_haven.webp";
+import silhouette from "assets/bumpkins/silhouette.png";
```
spencerdezartsmith, line 29: I am wondering if this silhouette could be a little more opaque. It was showing for me a lot when loading modals etc and it feels pretty dark 🤔

spencerdezartsmith (approved):
Code looks good and was working. I do think the silhouette is a little too dark but that is non blocking for sure! Excellent improvement

src/features/auth/useImagePreloader.tsx
adamhannigan: Good idea - definitely open to tweaking this 😄

PR #1901, opened 2022-12-19
[FEAT] Remove access to old lands

adamhannigan:
# Description

This PR removes all old access points to Sunflower Land. Now that everyone has migrated, we want a sure way to prevent accessing the old lands.

To prevent sneaky players, all old events have been removed as well

**Includes**

- Remove old routes
- Remove old events
- Fixed references to old events with new events

**Excludes**

This is not an entire clean up of the old code base, just direct references to old events. The remaining UI components will be cleaned up over the following week


src/features/retreat/components/blacksmith/RetreatBlacksmith.tsx
```diff
@@ -44,7 +45,9 @@ export const RetreatBlacksmith: React.FC = () => {
           <Action className="pointer-events-none" text="Craft" icon={icon} />
         </div>
       </div>
-      <ItemsModal isOpen={isOpen} onClose={() => setIsOpen(false)} />
+      <Modal isOpen={isOpen} onClose={() => setIsOpen(false)}>
```
c-py, line 48:
```suggestion
      <Modal centered show={isOpen} onHide={() => setIsOpen(false)}>
```

`isOpen` and `onClose` don't exist on Modal component.

c-py (dismissed):
LGTM! Just a small fix on the craft modal required

PR #1902, opened 2022-12-20
[FIX] Disable button for double mints

adamhannigan:
# Description

Double mints have now been prevented on the API side.

This PR removes the button once a user has already minted an auctioneer item.


spencerdezartsmith (approved):
🚀

PR #1920, opened 2022-12-22
[FEAT] Rebuild state with actions

adamhannigan:
# Description

We are currently seeing an increased amount of API errors coming through for incorrect actions. These should not happen if the UI state is correct.

Our current thoughts are that at some point in the gameplay a user's session is becoming out of sync with that of the server.

This fix ensures we also update the FE with the server state


spencerdezartsmith (approved):
LGTM 🤞🏼

PR #1925, opened 2022-12-22
[FIX] Fallback to RPC

adamhannigan:
# Description

There are currently issues with the Public Polygon RPCs 😢 

1. Split up Blockchain calls and remove un-needed ones
2. On error, fallback to private RPC


spencerdezartsmith (approved):
Working well for me.

PR #1926, opened 2022-12-23
[FEAT] Show Frogs and Seals on Farm and Garden

tiffanydys:
# Description
This feature includes showing of frogs and seals on both main farm and community garden.

## Type of change

# How Has This Been Tested?
`yarn dev`
## On Main Farm:
**FROGS**
image

**SEALS**
Screen Shot 2022-12-23 at 8 22 43 AM
Screen Shot 2022-12-23 at 8 22 38 AM
Screen Shot 2022-12-23 at 8 22 31 AM
Screen Shot 2022-12-23 at 8 22 26 AM
Screen Shot 2022-12-23 at 8 22 19 AM

## On Community Garden:
Screen Shot 2022-12-23 at 8 29 29 AM

# Checklist:


src/features/game/expansion/components/Water.tsx
```diff
@@ -215,6 +199,12 @@ export const Water: React.FC<Props> = ({ level }) => {
           />
         </div>
       </MapPlacement>
+
+      <MapPlacement x={offset} y={-offset} width={6}>
```
adamhannigan, line 203: Now that we are interacting with more external contracts I would feel safer if we created a new `CommunityBoundary` component that could catch any potential errors (such as external API/smart contract errors).

In this case here I would suggest creating a new wrapper boundary, similar to our `ErrorBoundar.tsx`

<CommunityBoundary>
  <MapPlacement x={offset} y={-offset} width={6}>
ProjectDignityFrogs left={0} top={0} />
      </MapPlacement>

      <ProjectDignitySeals isGarden={false} />
</CommunityBoundary>
```


```
// CommunityBoundary.tsx
class CommunityBoundary extends Component<Props, State> {
  public state: State = {
    hasError: false,
  };

  public static getDerivedStateFromError(_: Error): State {
    // Update state so the next render will show the fallback UI.
    return { hasError: true };
  }

  public render() {
    if (this.state.hasError) {
      return null;
    }

    return this.props.children;
  }
}
```

adamhannigan (changes requested):
Nice work and LGTM ✅ 

Just to be cautious, I would like to see some more error boundaries just to protect the main island from erroring out 🙏

src/features/game/expansion/components/Water.tsx
tiffanydys: Hi @adamhannigan great idea to catch these unexpected errors 👌🏻 
I've now added this `CommunityBoundary.tsx` component and imported it to `Water.tsx` 😄 
Cheers!

adamhannigan (approved):
LGTM 🚀 Excited to see my frog and seal again!

PR #1940, opened 2022-12-28
[FEAT] Add beta pass coupon

adamhannigan:
# Description

Include a new coupon that gives players early access to testing feautres.

The purpose is to give moderators and contributors early access so we can get feedback and fix any bugs before launch.

This will be used for the upcoming Treasure Island, Fruit & Magic Beans 🚀


spencerdezartsmith (approved):
🔥

PR #1951, opened 2022-12-29
[FEAT] Jigger detection on withdraw/trade

adamhannigan:
# Description

Jigger has made some significant improvements to their detection system and we are re-enabling them.

The prompt to verify now pops up when withdrawing/trading, instead of being obtrusive in the log in process.

This check happens on:

- Bank
- Hot Air Balloon Listing
- Storage House Delivery

Screen Shot 2022-12-30 at 10 11 54 am
Screen Shot 2022-12-30 at 10 12 02 am

### DIFF
--- a/src/features/goblins/trader/selling/Selling.tsx
+++ b/src/features/goblins/trader/selling/Selling.tsx
@@ -1,12 +1,16 @@
-import React, { useContext } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { useActor } from "@xstate/react";
 
 import alert from "assets/icons/expression_alerted.png";
+
+import * as AuthProvider from "features/auth/lib/Provider";
 
 import { Context } from "features/game/GoblinProvider";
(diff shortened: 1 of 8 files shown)


src/features/goblins/trader/selling/Selling.tsx
```diff
@@ -28,6 +32,48 @@ export const Selling: React.FC = () => {
...
+        setJiggerState({
+          url: check.verificationUrl,
+          status: check.botStatus as JiggerStatus,
+        });
+      }
+      setIsLoading(false);
+    };
+
+    load();
+  }, []);
```
c-py, line 61: **suggestion:** This API request is triggering every time the tab is changed between selling and buying.

I think we can consider caching the response, or moving the check up one level into the `TraderModal`.

c-py (changes requested):
Small suggestion on the trader modal.

c-py (approved):
LGTM!

PR #1969, opened 2023-01-04
[FEAT] Fruit UI animation

HBandesh:
# Description

In this PR, I have added animation for fruit drops. the animation path is similar to what we have in wood/stone drops. There is an added variability of dropping the fruit on the left side or right side.

https://user-images.githubusercontent.com/35486909/210484855-f97b7cde-0132-4409-a101-8e73688d2085.mov

Fixes #issue

## Type of change

# How Has This Been Tested?

# Checklist:


adamhannigan:
Nice work mate, just a few small things.

1. When clicking on harvest, the recovery modal immediately appears. This blocks part of the drop animation. I would suggest not showing this modal right after a harvest.

Screen Shot 2023-01-04 at 4 53 07 pm

2. The animation feels like it jolts extremely fast and goes flying. I would consider reducing the speed to make it feel more realistic

3. Is it possible to make the apple fall closer to the tree? It is almost 3 grid squares away from the middle of the tree when it ends. 1-2 grids would feel a bit better.

Polysemouse:
> Nice work mate, just a few small things.
> 
> 1. When clicking on harvest, the recovery modal immediately appears. This blocks part of the drop animation. I would suggest not showing this modal right after a harvest.
> 
> Screen Shot 2023-01-04 at 4 53 07 pm
> 
> 2. The animation feels like it jolts extremely fast and goes flying. I would consider reducing the speed to make it feel more realistic
> 3. Is it possible to make the apple fall closer to the tree? It is almost 3 grid squares away from the middle of the tree when it ends. 1-2 grids would feel a bit better.

I would suggest using sprites for the drop animation like how it is done for resource nodes / regular trees.  The current animation feels too smooth in the pixel-y world compared to other resource harvesting animations

HBandesh:
> 

The reason we pivoted from that approach is that using sprites/gifs for the animations is not a scalable solution. A lot of assets are downloaded on the page, which will only increase with time as we release more and more features. We can reduce the bytes from the images significantly by converting them to WebP but the same cannot be done for the sprites and gifs. They lose a lot of quality in that process. In the current scenario if I only talk about trees, stones, iron, and gold on the page then for each one of them we have two sprites -> shaking/minting and a second for drops. On average the size of each one of them is around 12kb (8kb for drop sprite and 4kb for shake/mint sprite). So just for these four resources - trees, stone, iron, and gold- we are loading around 48 KB of data which is huge! Now imagine if I take other animations into account. It will easily cross 100kb of data per page.

But with this approach, we have created a component (which will now be used for other drops as well). No sprite and shake animation would be downloaded and it would be done using CSS. This will significantly reduce the load on the page and hence increase user experience. A single component and single CSS file will handle all the animations. 

Definitely while doing this using CSS, it may take some time to perfect it but it would be worth taking that route. I will increase the time of the transaction and smoothen out the offset path further to make it more realistic.

Polysemouse:
@HBandesh You are right, it's easier to scale the game with the new approach of animating drops with css

HBandesh:
@adamhannigan and @Polysemouse . I have created a new offset path for the drops. This will now drop the fruit closers to the tree. I have made some changes in the duration and the keyframes of the animation, and I think it is smoother than before. Please do let me know your thoughts on this. Plus, I have made changes to the popover. It won't appear when the animation is ON.

Polysemouse:
@HBandesh The animation looks better now.

There's one small issue.  The fruit drop animation isn't showing for the last harvest (when `harvestsLeft` is 1)

adamhannigan:
Feels much smoother @HBandesh 👏 

Just one small thing that caught my eye, as soon as I click harvest there is a flash that happens and text that briefly appears.  This happens around 60% of the time.

See attached.


https://user-images.githubusercontent.com/11745561/210657234-25067859-968f-488c-b419-46f8a091190a.mp4
Screen Shot 2023-01-05 at 8 53 55 am


Any idea what could be causing this?

adamhannigan:
Merging in PR now for testnet/beta testing.

We can look into the above issue separately.

PR #1974, opened 2023-01-05
[FEAT] Lady bug and Goblin Blacksmith

adamhannigan:
# Description

This PR introduces support for the Lady Bug and re-opens the Goblin Blacksmith 🚀 

The code for the Blacksmith is very similar to the old `Rare.tsx` component. 

The idea of a `LimitedItem` was a bit confusing so I've removed old references around this.

Screen Shot 2023-01-05 at 1 33 51 pm

**Business rule change**

- No longer any cool downs before a user can place a minted item. Since there is a place item cooldown, we don't need an additional one on the mint
- Goblin Blacksmith Items with not be withdrawable by default - only once the mint ends will we enable these

**Excludes**

- Lady Bug Art + Placing

## How to test?

Run the API locally.

### DIFF
--- a/src/features/retreat/components/blacksmith/RetreatBlacksmith.tsx
+++ b/src/features/retreat/components/blacksmith/RetreatBlacksmith.tsx
@@ -65,6 +65,7 @@ export const RetreatBlacksmith: React.FC<Props> = ({ inventory }) => {
               shoes: "Black Farmer Boots",
             }}
             tabs={[{ name: "Craft", icon }]}
+            onClose={() => setIsOpen(false)}
           >
             <GoblinBlacksmithItems onClose={() => setIsOpen(false)} />
           </CloseButtonPanel>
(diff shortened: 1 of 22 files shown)


Polysemouse:
Would users be able to mint multiples of the same item?

adamhannigan:
@Polysemouse Nope. There is a mint protection that means you can only mint it once 👍

src/features/retreat/components/blacksmith/RetreatBlacksmith.tsx
```diff
@@ -46,9 +53,26 @@ export const RetreatBlacksmith: React.FC = () => {
...
+            bumpkinParts={{
+              body: "Goblin Potion",
+              hair: "Blacksmith Hair",
+              pants: "Brown Suspenders",
+              shirt: "Yellow Farmer Shirt",
+              tool: "Hammer",
+              background: "Farm Background",
+              shoes: "Black Farmer Boots",
+            }}
+            tabs={[{ name: "Craft", icon }]}
```
c-py, line 67:
```suggestion
            tabs={[{ name: "Craft", icon }]}
            onClose={() => setIsOpen(false)}
```

**issue:** The close button isn't working

c-py (changes requested):

PR #1998, opened 2023-01-08
[FIX] Safety check on basket open when visiting other farm

AzaroQL:
# Description

There is an issue when opening inventory on a farm that has no bumpkin.
For example visit farm 77 and click on inventory.

## Type of change

# How Has This Been Tested?

yarn test

# Checklist:

### DIFF
--- a/src/features/game/events/landExpansion/plant.ts
+++ b/src/features/game/events/landExpansion/plant.ts
@@ -77,7 +77,7 @@ export const getCropTime = (
   crop: CropName,
   inventory: Inventory,
   collectibles: Collectibles,
-  bumpkin: Bumpkin
+  bumpkin: Bumpkin | undefined
 ) => {
   let seconds = CROPS()[crop]?.harvestSeconds ?? 0;
 
--- a/src/features/game/events/landExpansion/plant.ts
+++ b/src/features/game/events/landExpansion/plant.ts
@@ -79,8 +79,6 @@ export const getCropTime = (
   collectibles: Collectibles,
   bumpkin: Bumpkin
 ) => {
-  const { skills, equipped } = bumpkin;
-  const { necklace } = equipped;
   let seconds = CROPS()[crop]?.harvestSeconds ?? 0;
 
   if (inventory["Seed Specialist"]?.gte(1)) {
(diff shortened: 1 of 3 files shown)


mGranTY:
I think its better to show the "missing bumpkin" modal when a player has no bumpkin. Idk how some players dont get that modal.

AzaroQL:
> I think its better to show the "missing bumpkin" modal when a player has no bumpkin. Idk how some players dont get that modal.
> 
> For visiting mode, this fix is fine tho.

Yeah, thats only for 'visit farm' case.

src/features/game/events/landExpansion/plant.ts
```diff
@@ -107,8 +101,15 @@ export const getCropTime = (
     seconds = seconds * 0.85;
   }
 
-  if (skills["Cultivator"]) {
-    seconds = seconds * 0.95;
+  if (bumpkin) {
```
spencerdezartsmith, line 104: This conditional check can be removed here now.

src/features/game/events/landExpansion/plant.ts
```diff
@@ -77,10 +77,8 @@ export const getCropTime = (
   crop: CropName,
   inventory: Inventory,
   collectibles: Collectibles,
-  bumpkin: Bumpkin
+  bumpkin: Bumpkin | undefined
```
spencerdezartsmith, line 80: Rather than doing this here, I think we should just do a bumpkin check at the beginning of this function and the `plant` action. If no bumpkin is found then we just return an error.

```
  if (!bumpkin) {
    throw new Error("You do not have a Bumpkin");
  }
```

spencerdezartsmith (changes requested):
Nice catch on this one. I left some comments but basically if someone doesn't have a bumpkin they should immediately be shown the modal saying they don't have a bumpkin. They should move into the `noBumpkin` state inside the machine which allows nothing else to happen. 

This should happen even if visiting. I would prefer fixing that then adding logic to handle cases where no bumpkin is present.

AzaroQL:
I faced quite a few problems in reusing noBumpkins modal for visiting:
1. If I add condition to point to noBumpkin state I end up in wrong location since visiting state is hanlded separately
2. If I add noBumpkin modal to visiting state we face issue with not being able to return to farm or visit other farm

I cant see an easy fix to this, imo my original approach is fine since we dont have a lot of interactions in visiting state, so this check for bumpkin is the only one that we need to do

PR #2032, opened 2023-01-13
[CHORE] Use icons in separate repo

adamhannigan:
# Description

As we scale we need to be proactive about any potential issues that could arise. Despite our 'No Licence', we need to be careful what can be accessed through this repo.

To avoid any legal repercussions, we will place asset pack files in a private repo.

This PR extracts the raw SunnySide Asset files from the asset pack and puts it in a protected location 🔒 

See `src/assets/sunnyside.ts` to view the new protected approach.


c-py (approved):
Huge effort on the grind, especially on the Friday Funday!

Nice work.

PR #2036, opened 2023-01-16
[FIX] Hide placed items in Goblin Retreat

adamhannigan:
# Description

Similar to the Island HUD, filter out any items that are placed.

I've done this on the root level, so these items won't be available anywhere inside of the Retreat - this then automatically applies to the Auctioneer and Blacksmith.

**Backend validation coming in separate PR**

### How to test?

1. Sync your island (so items will exist in Goblin Retreat)
2. Place a collectible (e.g. Brilliant Bear)
3. Visit Goblin Retreat - item should now be hidden at Auctioneer.

### DIFF
--- a/src/features/game/components/Revealing.tsx
+++ b/src/features/game/components/Revealing.tsx
@@ -1,4 +1,3 @@
-import { SUNNYSIDE } from "assets/sunnyside";
 import React from "react";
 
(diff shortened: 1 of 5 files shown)


spencerdezartsmith (approved):
Working well 🐻

PR #2039, opened 2023-01-16
[FEAT] Weekly news announcement

adamhannigan:
# Description

This PR includes support for weekly announcements.

The aim is to handcraft these messages each week in a PR - over time we will discover what works and automate the process.

Screen Shot 2023-01-16 at 3 48 05 pm


hujikol:
could we remove the `underline` class on the `announcement.title` for the consistency, that underline will only be used in links only?

PR #2058, opened 2023-01-18
[FIX] Update ClockIssue Modal

0xSacul:
# Description

Simple addition of an extra text and a button that redirects to our guide to sync the clock of his computer on Freshdesk.

The redirection is done in the same window, editable in `target="_blank"` if needed. Same for the text :) 

image

## Type of change

# How Has This Been Tested?

Locally & `yarn test`
image

# Checklist:

### DIFF
--- a/src/features/game/components/ClockIssue.tsx
+++ b/src/features/game/components/ClockIssue.tsx
@@ -1,15 +1,31 @@
 import React from "react";
 import suspiciousGoblin from "assets/npcs/suspicious_goblin.gif";
+import { Button } from "components/ui/Button";
+const SyncClockGuide =
+  "https://sunflowerland.freshdesk.com/en/support/solutions/articles/101000397200-how-to-sync-my-windows-clock-";
 
 export const ClockIssue = () => {
   return (
     <div className="flex flex-col items-center text-center p-2">
       <span>Clock not in sync</span>
       <img src={suspiciousGoblin} alt="Warning" className="w-16 m-2" />
-      <span className="text-xs">
+      <span className="text-sm mt-2 mb-2">
         Uh oh, it looks like your clock is not in sync with the game. Set date
         and time to automatic to avoid disruptions
       </span>
+      <span className="text-sm mt-2 mb-2">
+        Need help to sync your clock? Have a look at our guide!
+      </span>
+
+      <div className="flex w-full">
+        <Button
+          onClick={() => {
+            window.location.href = SyncClockGuide as string;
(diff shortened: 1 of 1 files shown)


Polysemouse:
Should the button open a new tab instead of navigating away from the game?  If yes, you can refer to how the implementation can be done in https://github.com/sunflower-land/sunflower-land/blob/f911ffd3906a48b63f77ebce7ad49fa8353c0031/src/features/retreat/components/resale/Resale.tsx#L36

0xSacul:
@Polysemouse 

I thought about it as you can see in my PR description.

But I also thought that whatever happens the person will have to refresh the page so to avoid confusion for some new player I didn't include `window.open`, a bit like it is done with the Proof of Humanity the Jigger error you can see below.

https://github.com/sunflower-land/sunflower-land/blob/c9b8dc15a9a6aac61567881c2ec6f2686d800e58/src/features/game/components/Jigger.tsx#L34

But as said, if needed we can use `window.open`

Polysemouse:
@0xSacul That's true.  In that case we can keep your original implementation (` window.location.href = ...`)

c-py:
@0xSacul @Polysemouse Do you think a button or a link (with underline) is more informative for the player?

I think a link indicates the player will be going to a new page and a button indicates something in game.

What do you think?

Polysemouse:
@c-py @0xSacul I am kind of split for this one.  Underline text works better for links, however we have several underline texts in the game that are not links.  Perhaps we need to standardize underline text to reserve for links only so players have an expectation that underline texts are clickable (especially for mobile players)

0xSacul:
@c-py 

I think indeed that an underlined text is a better approach but it already happened that players don't understand that it's a link and don't try to click on it, however a button with the right text is always clicked even if it shouldn't 😅

I think the problem with the underlined text is that it is the same color as the text above it and so players think the underlining is something like a warning.

Waiting for your thoughts on that 👀

c-py:
> @c-py
> 
> I think indeed that an underlined text is a better approach but it already happened that players don't understand that it's a link and don't try to click on it, however a button with the right text is always clicked even if it shouldn't 😅
> 
> I think the problem with the underlined text is that it is the same color as the text above it and so players think the underlining is something like a warning.
> 
> Waiting for your thoughts on that 👀

I agree with this 😆 . Players might not click the underline text.

Let's roll with button then!

c-py (approved):
LGTM!

PR #2072, opened 2023-01-22
[FIX] EGG HOARDING LIMIT FROM 60 TO 100

mGranTY:
# Description

As the title suggest, this PR changed the Hoarder limit from 60 to 100 to match the blockchain sessions values

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/features/game/lib/processEvent.ts
+++ b/src/features/game/lib/processEvent.ts
@@ -21,7 +21,7 @@ const maxItems: Inventory = {
   Blueberry: new Decimal("100"),
 
   Chicken: new Decimal("20"),
-  Egg: new Decimal("60"),
+  Egg: new Decimal("100"),
   "Speed Chicken": new Decimal("5"),
   "Rich Chicken": new Decimal("5"),
   "Fat Chicken": new Decimal("5"),


TheDcol:
Is it possible to increase hoard limit by the amount of hen houses you have installed on land?

Just thinking what will happen when we have 4 hen houses.. someone with all boosts will gather 144 eggs every 34 hours

Right now it's 72 eggs every 34 hours with 2 hen houses

Polysemouse:
Another suggestion is to modify the backend to return the sync limit, so they can be dynamic and don't have to be set on the frontend

mGranTY:
> Another suggestion is to modify the backend to return the sync limit, so they can be dynamic and don't have to be set on the frontend

@Polysemouse we can fetch the limits directly from the blockchain and put them inside the ProcessEvent function. If we want a Dynamic limit we would need to remove the limits from the blockchain and set them on the backend or change the way blockchain limits work inside the session contract.

Polysemouse:
@mGranTY I think the limit on the blockchain is still needed in case someone found a bug and mints billions of iron pickaxes in 1 transaction.  Perhaps the dynamic limit can be implemented in the backend and the blockchain limit serves only as the humanly possible limit

Polysemouse:
@mGranTY personally I think we should at least double the egg limit (to 120) temporary as we have double the number of hen houses

mGranTY:
> @mGranTY personally I think we should at least double the egg limit (to 120) temporary as we have double the number of hen houses

@Polysemouse i agree with you, this PR is just to make the numbers match, if a core devs appear here, please change the limit to 120 😂

spencerdezartsmith (approved):
💯

PR #2098, opened 2023-01-26
[FIX] Add price per unit when selling resources

DeMouraZ:
# Description

Although my main fix is adding price per unit, both are thinking on newer players making fewer mistakes in the game.

- Adding Price per Unit so players know for how much they are selling and prevents mistakes that hurt a lot sometimes, even for certain items helps us alot from using calculator 😆

- Setting only 2 decimals for selling and buying Idle. While doing the fix above someone told me that some new users didn't knew when going to #selling-ranch that 1.002 wood is not one thousand and two woods and could be scammed. For less problems and complains in the future and because we all want new players to have an experience that make them stay.

Price Per Unit Example
priceperunit

Difference with amount of resources when buying visiting the land
Before:

After:

When a player wants to sell them 
Before:

After:

Sorry for the lack of quality of the images 😒

## Type of change

# How Has This Been Tested?

Manually on local.

# Checklist:

### DIFF
--- a/src/features/goblins/trader/buying/components/Idle.tsx
+++ b/src/features/goblins/trader/buying/components/Idle.tsx
@@ -70,7 +70,10 @@ export const Idle: React.FC<IdleProps> = ({
           const listing = farmSlot.listing;
           const listingId = listing.id;
           const resourceName = KNOWN_ITEMS[listing.resourceId];
-          const resourceAmount = listing.resourceAmount;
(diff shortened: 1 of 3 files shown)


c-py:
Thanks for the suggestion! After giving it some thought, we've decided not to merge this change at the moment as it makes the trader UI a bit more complex and harder to read. Maybe there's a different way to approach this that we can chat about!

PR #2114, opened 2023-01-29
[FEAT] Bulk Sand Shovel

adamhannigan:
# Description

This PR updates the UI to support bulk purchases of Sand Shovels.

A user can only buy a minimum of 5 Sand Shovel - Since roughly 2/3 holes are empty, we want to make sure they have enough shovels to find an item

Screen Shot 2023-01-30 at 10 44 54 am

### DIFF
--- a/src/features/game/events/landExpansion/craftTool.ts
+++ b/src/features/game/events/landExpansion/craftTool.ts
@@ -74,7 +74,11 @@ export function craftTool({ state, action }: Options) {
 
   const oldAmount = stateCopy.inventory[action.tool] || new Decimal(0);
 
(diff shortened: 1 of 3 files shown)


Polysemouse:
What do you think of we rename the craft button to "Craft 5"?

adamhannigan:
@Polysemouse This was the initial idea, but it caused confusion how many resources were needed.

For instance, tt wasn't clear if the price per shovel would be 2 Wood, 10 wood or 50 wood.

We decided to change the title to `5 x` to make it clearer

c-py (dismissed):
LGTM!

Just need to add the `amount` into `trackActivity` call

c-py (approved):
LGTMd

PR #2119, opened 2023-01-30
[FIX] Treasure Island Metadata

adamhannigan:
# Description

Ran `yarn metadata` to produce missing metadata

Screen Shot 2023-01-30 at 5 24 32 pm


c-py (approved):
LGTM decimals all looking good.

PR #2122, opened 2023-01-30
[POC] Lottery

adamhannigan:
This PR includes an experiment of a lottery system. It is not the decided approach, a draft PR was created just for testing. There are lots of ideas, but without experimenting any there won't be any progress

**Goal**

- Support all timezones (24 hours)
- Less RPC bugs & high gas fees
- Promote equality (not people with best RPC connections)
- Encourage new players and give them a chance - if the game only favours OG players, no new players will come and all assets devalue. It is in everyone's best interest to create a fair and engaging game for all players 

**Description**

When an auction opens, players have 24 hours to cast their bids.

Resources/SFL are escrowed (off chain) until the auction is over - players can not access these items. This guarantees that a player is actually intending to mint the item.

A player can select how many lottery tickets they wish to use - if there is more supply than participants, then players who used the most lottery tickets win. This ensures loyal players have a higher chance of minting these items.

After 24 hours, players come back to check the results of the lottery.

Losers get their refund (except their lottery tickets), winners can mint the item.


maxisthemax:
this is what i suggested last time in discord.

DaanMich:
Just read this and thought I'd give you guys some feedback/remarks:

1. Seems like this is a lottery system which pushes players to have more than multiple accounts to have a chance at winning a lottery ticket to mint an NFT. How will this be dealt with? 

2. This will be a big nono for older players (including me) who know how to set manual gas/custom RPC. The randomised aspect does ensure new players to get a chance at minting an nft but I doubt this is the best solution seeing this will leave a sour taste for lots of OG's. Secondly I think this also does not promote the learning experience from a blockchain point of view. I'd rather see all players educated on how they can set custom gas/custom RPC's seeing that was/is the goal of the game. 

3. The thin spreading of boost nfts could also have something to do with the recent run on Wendy (and thus many people missing out). Imo minting cost + supply should be looked at. In addition to that I think small boost nfts + high supply could also help solve this problem partly. More high supply mints with a smaller boost should keep the newer players busy.

**Conclusion**: I think this feature will have a lot of backlash. Pushing players to multiaccount, luck based mints (even more so than now) and not many boost nfts are a few things that come to mind. Educating the community would be a better option imo. 

Just my thoughts on this. Hope you consider all the elements!

AzaroQL:
I like the idea, since current approach does not guarantee a mint due to errors and high load issues, even for experienced OG users who know how to set the gas.

So we actually change from frustrating lottery of who gets the error or who mints the item, to more fair lottery where you have no need to wake up at night, no need to stress over refreshing the page multiple times and no need to pay high gas fees.

Supply and price of items should also be reconsidered to reduce the demand in some way(looking at cheap low supply Wendy)
 
Leftover items can be left open for crafting for few more days(or burned right away)

Polysemouse:
I like the idea, however it isn't really a blind auction if we support different value bids (because player can see what other players deposited off chain).

iwasthesword:
> 2\. Players will always find something to complain about and may question the validity of the random winners. More work will need to be done to verify we used a truly random seed shuffle (e.g. using Transaction Hash at a given block)

Maybe [chainlink VRF](https://docs.chain.link/vrf/v2/introduction/) could be used? Then have the tx in the UI so anyone can verify.

SFL-Aeon:
I don't like this system as users can create new accounts attempt to place a bid and then take over the entire supply or most of it.

I don't believe a lottery system will solve players being happy about the mints because then you have people who will complain about those able to mint and just flipping the nfts for profits.

The lottery system doesn't necessarily cover the few main issues i believe which are as follows:

The supply is never enough for the current player base.
The cost to mint is always to low in order to entice players to mint which shouldn't be the case.
Users are upset because of the following two and not able to mint so preventing them from minting will ultimately end with them upset about that.

I don't believe the costs should be ridiculous but they should be realistic with the current supply of resources and the buff they are offering. The wendy nymph mint could've easily asked for 2-3K wood and it would've minted out just as quick as it did but with more burn of current supplies as well as give its actual value to the nft for the cost.

WTFpacman:
Highly against it alot of people like sfl cuz it doesnt have that much RNG. The RNG with the expands was already quite bad and didnt ended well as we have seen.

brian-man:
> players have 24 hours to cast their bids

I don't see why this significant change is necessary.  Why not keep the 2x 30min windows of the current model?

drumdrumdrum3337:
+1 for the idea. As Azaro already mentioned, current mints are basically lotteries anyway, but with tons of added frustration. Experienced players won't like it since they know how to set up MM for optimal results and get an edge over less experienced players. But what is more important is that the current system is a point of frustration for the new players, and I'd say we need to take care of them so that game could grow.

brian-man:
It's a completely different model, but I'll offer it up for consideration anyway....

Let's assume `Collectible Bear` is in high demand.  The current listed cost is 1 green bear (I forget his name and there's no tooltip) and 25 gold.  Possible supply is listed as 150x2, so 300.  Let's focus on just the gold part and also for the moment ignore the timezone stuff.

Suppose demand was 5000 at that price.  With the current model <= 300 people will cheer and/or breathe a sigh of relief while 4700 will be at best frustrated.  The root issue IMO is that the demand/supply relationship is a step function rather than a cureve.  Meaning it's like an offer on an exchange that's priced too low rather than a liquidity pool where the pricing is more dynamic/smooth.

Via both on-chain and (presumably) off-chain analytics, the devs have access to player ingredient supply information and production information.  With that the pricing can be tuned accordingly.  What they don't know is interest/demand beyond just "can afford".

Proposal for new auctioneer mints:
1. Decide on the boost.
2. Decide on the supply.
3. Set the initial price such that only 10% of the supply can be bought immediately by players -- meaning if you have 1000 supply, then only 100 players can actually afford to buy the item immediately.
4. Lower the cost by 1% daily.

The whales can decide if they want to spend "launch price" or "gamble" on waiting a day or more while more competitors surface.

Mbst36:
I like Azaro's system with one more layer of protection.

Shorten the timeframe to receive a lottery ticket to ~1 minute and add a captcha or other anti-bottling device to receive the lottery ticket.

This would heavily decrease the amount of multi account users getting access to the mint

andronicu3:
I don't like that idea at all of random winners, that would take away so much from the game for me as I'm not able to better my chance to mint.. With all that time i spend playing i would see that I'm able to mint and for that it would be best for low,mid,high boost nft were the really rare item have a lot higher recuerements to mint those items or something others has mention here 

I feel there have to be some reward for those who spend a lot of time in the game, raffle could lead to only people that don't care about the game and just want to cash out as much possible wins all and that with multiple farms...

It just feels so wrong with with a rng raffle if anything it has to be waighted raffle in some sort but straight rng raffle is a no no
it opens up for so much for bad actors with multiple farms and what not and lead to those who really put in time and effort feels left out and without any competition or chance to better theirs chances...
I know that current system isn't perfect and there needs to be something done about it but i haven't missed out yet just bcoz i have a chance to better my chances and would that go away to just about anyone can mint rare items without putting in any effort or money more than get a ticket to get those items feels (i can't even describe how it feels)

There has to be a way to get those nft through game play and with the resources we collect and with a possibility to better our chance to mint...
So either with high priced to low priced boost items or through game play like fruit nft, just something else than rng..
I wish i had a solution but from what i feel rng raffle feels so wrong

Aquila199:
1. This system encourages multi accounting and it will be abused but people that just want to sell the item
2. a bid system has issues due to block chain being able to read the values also makes the game 100% whale only 

out of all the idea i think this is one of the worst directions due to abuse issues and whale focused

i have suggested it already please consider more grind focused mints like fruit where most players have a shot and requires people to play the actual game not buy supplies, mint then sell

kegw7689:
Leaving it to a lottery system would be anticlimactic in my opinion. Wouldn't the problem with the errors be solved if there was a quest involved in the mint? For example keep the price of 25 gold and 1 brilliant bear and add a quest where we need to collect tickets by harvesting crops. That way there won't be a lot of people that collect the needed tickets at the same time to get an error when clicking the mint button

Kay1797:
> It's a completely different model, but I'll offer it up for consideration anyway....
> 
> 
> 
> Let's assume `Collectible Bear` is in high demand.  The current listed cost is 1 green bear (I forget his name and there's no tooltip `*`) and 25 gold.  Possible supply is listed as 150x2, so 300.  Let's focus on just the gold part and also for the moment ignore the timezone stuff.
> 
> 
> 
> Suppose demand was 5000 at that price.  With the current model <= 300 people will cheer and/or breathe a sigh of relief while 4700 will be at best frustrated.  The root issue IMO is that the demand/supply relationship is a step function rather than a curve.  Meaning it's like an offer on an exchange that's priced too low rather than a liquidity pool where the pricing is more dynamic/smooth.
> 
> 
> 
> Via both on-chain and (presumably) off-chain analytics, the devs have access to player ingredient supply information and production information.  With that the pricing can be tuned accordingly.  What they don't know is interest/demand beyond just "can afford".
> 
> 
> 
> Proposal for new auctioneer mints:
> 
> 1. Decide on the boost.
> 
> 2. Decide on the supply.
> 
> 3. Set the initial price such that only 10% of the supply can be bought immediately by players -- meaning if you have 1000 supply, then only 100 players can actually afford to buy the item immediately.
> 
> 4. Lower the cost by 1% daily.
> 
> 
> 
> The whales can decide if they want to spend "launch price" or "gamble" on waiting a day or more while more competitors surface.
> 
> 
> 
> `*` Apparently it's Brilliant Bear.

with this, does the daily amount burn?
Example: If the 12 is minted from the first 100, will 88 of the supply be burned , or just added to the next day's supply?

either way, I love this idea over a lottery system.

vitt0:
> Leaving it to a lottery system would be anticlimactic in my opinion. Wouldn't the problem with the errors be solved if there was a quest involved in the mint? For example keep the price of 25 gold and 1 brilliant bear and add a quest where we need to collect tickets by harvesting crops. That way there won't be a lot of people that collect the needed tickets at the same time to get an error when clicking the mint button

Addig something like this makes a lot more sense Imo, altrough I don't think is needed for all the auctions, Imo the decorative NFT don't have a high demand, wiht previous auctions we didn't have problems, even more that most of them didn't sold out.

In addition to this I think it would be better to split the auctions is 3 waves, each of them 8hs after the other, to help in splitting the attemps at the same time. Plus we could do like for example each time that there is a auctions that requires tickects the travel salesman comes to a small island with 3 a tiers quest, each of them give 1 tickect to attemp to 1 of the wave of the auction and the progression to the next tier requires a harder quest, Ideally this quest would be related to whatever the NFT boost is, like for example Wendy's quest could have be like: 

Tier 1 - Chop 250 trees
Tier 2 - Chop 750 trees
Tier 3 - Chop 1250 trees

(Or something around this line)

So as I say before maybe decorative NFT don't need quest and only do them for NFT with boost, unless they have a small supply like fore example Collectible Bear (althourgh with the expensive price that it has I think the demand is not going to be that high)

brian-man:
> with this, does the daily amount burn? Example: If the 12 is minted from the first 100, will 88 of the supply be burned , or just added to the next day's supply?
> 
> either way, I love this idea over a lottery system.
My mental model (for your example) is:
- Mintable supply of 100 on day 1
- 12 mint on day 1
- Mintable supply of 88 on day 2
etc.

andronicu3:
Leaving it to rng doesn't feel like a solution at all it would ruin the game as a game to just become a lottery where you don't know anything and have nothing to fight for... 
Seeing some people write that if you really want it nft and if you don't win the lottery, you just have to buy it.. 
It's not that simple, it's not enough that you don't win, but there is no direct competition or something that creates some exitment until mint and then end up on the market for several hundred matic and bought up by those who have it to be sold even more expensive... 
If you then have bad rng and don't have several hundred matic, there is no chance of ever being able to get rare items... 
There needs to be a system where both those with money and those who invest a lot of time should be able to get the opportunity to mint... 

The way fruit nft works now is actually a good system, but it doesn't have to be that way,, there may be several smaller competitions or quests you need to get through for to finally be able to get that ultra rare nft....
During the time when you are fighting your way there may be less boost items and decoratives as it has been a few times.. Then those who can't bear or don't want to spend time on the game can buy their nft which are then available for sale right from the start and so can the real players fight for a chance to be able to mint.. 

I never felt anything was wrong with the game and always accepted everything and stood on the other side of the people who complain about everything.... 
I wouldn't be annoyed if i miss a mint like the last few have been but handing those bad mints to rng where i have minimal to no chance to influence if i get nft or not and without any exitment makes me really "annoyed".... (sorry to say it like that wish i could come up with better word)

Have rare mints to rng is not a way to go and in the end only those with a lot of money can buy and the losers didn't even get a chance to influence their chance or felt any competition,, even a bad mints like it was lately you still felt exitment and that you could influence their chance...
Raffle is gambling and if it would be with valued stake makes it even more gambling and gambling only attracts bad actors...

adamhannigan:
Wasn't expecting so much traction on this experiment 😅 

There are a lot of great comments but I won't be able to address individually.

We are mixing up a few suggestions that are not exactly related to the lottery, this can be added additionally but I will not them here:

- Multiaccounting is addressed with detection + Jigger.
- Increasing supply, changing prices + boost is not related to this and can be done additionally

Here are some notes:

1. Multi-accounting is treated separately with detection + Jigger. However lets explore how it would now work:

This new approach now deters multi-accounters as they must 'escrow' (lock) resources on multiple accounts. Previously a multi-accounter could have multiple devices open and click mint at the same time. If they only wanted to spend resources on 1 item, they would just choose whichever device has a successful transaction and mint. Now, they must be willing to commit all of the resources across all of their devices.

2. The current system is already a lottery that just gives errors to the losers. These Blockchain errors are unavoidable since it takes (15-20 secs) to determine the supply on chain. Mostly whales/OG are favoured by the current system as they have learnt how to game it and set the best conditions to mint (pay for a private RPC, crazy high gas prices etc.). Whilst it is great for OG players, the game will crumble if we don't cater to new players as well which overall destroys the value of items that existing players have.

"So we actually change from frustrating lottery of who gets the error or who mints the item, to more fair lottery where you have no need to wake up at night, no need to stress over refreshing the page multiple times and no need to pay high gas fees." - Yes, this is the idea. Players are less frustrated as they face less errors and do not need to mint at inconvenient times.

3. In this proposed system, resources are 'escrowed' off chain - meaning that it is not visible what a player bids. However, most auctions would require the same bid from all users. Only on occassion would we do a 'blind auction'

4. Having a lottery system means we can add 'weighted' randomness. Since we know the desired supply and total amount of people trying to mint, we can filter out players to flip items. In our current approach we don't know how many people will mint, so we can't filter transactions. However, we don't want to influence a RNG, it would raise too many questions - if we want to combat people flipping we can just implement a new game rule - For example, if you withdraw an auction item (within 4 weeks), you cannot access the next auction.

5. We want to make this accessible for players in all timezones. Currently the 2 drops happen during working hours in Australia. This ensures someone is always available to debug and help users if needed. We are hesitant to do drops at 3am in case something happens 😢 

6. Deciding the boost post-mint is an idea we tossed up, but will lead to disatisified players if too many people want to mint. For example, if 5000 people minted a Wendy the buff would be `~0.02-0.05 wood`

7. The quest approach has been considered (and done before in old Goblin Village - require a certain flag or skill). However, these are hard to perform as everyone needs an equal starting point. Our best mints are for SFTs where the resources do not exist yet - see the current Fruit SFTs. However, once players have these resources (e.g. Lottery tickets) it just leads to a mad rush to mint at the same time

adamhannigan:
These comments sparked a few ideas 👏  

"How do we create a system that favours the loyal players?"

Imagine the suggested scenario where players can find 'Lottery Tickets' by doing daily actions - harvesting crops, chopping trees etc.

When an auction arises, players all pay the same amount of resources (e.g. 500 Wood, 10 Gold) but must choose a hidden amount of 'Lottery Tickets' that they want to use.

A loyal player would be able to allocate 50 Lottery Tickets, whereas someone not playing much would only allocate 20 Lottery Tickets.

When the winners are drawn, those with more Lottery Tickets have an increased chance. If you lose, your resources are refunded but not your Lottery Tickets. 

This creates a really engaging game theory aspect - since the bids are blind, a player must strategically determine how many lottery tickets they should burn for the auction. OG players must watch player trends and try predict the minimum amount of Lottery Tickets they want to use on the auction

SFL-Aeon:
I think the ticket idea is better than purely RNG based. This would allow players to choose what mints they want to go for and how badly they want this. 

It also allows for the more active and loyal players to mint the item over someone looking just to flip the item. Those who actively play have intentions to continue playing. I think this rewards those players and not focus on those just simply minting an item to flip it.

I would like to think though that the "lottery tickets" would not be tradable similar to war bonds and other event items. This would prevent people from selling them for profit and allowing whales to just purchase the mint access.

xSubZero:
I don't like this system, as mentioned above, the game is good for not having too much RNG and this proposal is to insert more, outside that opens loopholes for multi account and would continue to generate negative reactions in the community.

I believe it would be better to think of another way to get around the current mint problem.

Telknub:
> Imagine the suggested scenario where players can find 'Lottery Tickets' by doing daily actions - harvesting crops, chopping trees etc.

In my opinion this would be the best approach and the most fair one for everyone. If you lose in this, there won't be any frustration towards the game because of server issues or rpc issues, only towards the player that placed a low amount of lottery tickets.

andronicu3:
Latest ideas is definitly better than rng bcoz there must be some system so those how put a lot in the game both money and time gets a better chance than those who only grows wheat and witdraws all sfl and nft they earn in game and not reinvest... Those people maybe even normaly stay out of mints and didn't care now got a chance to easely take part...

It feels like you already in some way decided for to go for an raffle system then it's important it's benifit right people and close out bad people with bad intension so the idea with tickets you have to grind for is a way and it doesn't feels as bad as completely rng...
 In some way it will alway be somewhat of rng even with normal mint but the bigger chance you have to better your your luck 
lesser it will feel like a raffle...
I belive with some more brainstorming with right people you can come up with an idea that works for everyone exept bad people and there all can feel equal bad or happy
The part with a game rule can feel a bit strange maybe let them withdraw how they want without punishment for i feel that opens up for bad actions with multiple account just to take part in next mint rather than  benifit in raffle..
It could be a reward system fot those players that keep it in game so that can incurage others to keep it game or something else, punishment have never been a good solution...
I hope you will end up with something good and if you brainstorm with right people i'm sure you can come up with an idea that
work but for now it is a lot better idea and on the right way..
I'm not a brainstormy type of person but i will share how i feel and for now i feel calmer than before
..

Polysemouse:
> Deciding the boost post-mint is an idea we tossed up, but will lead to disatisified players if too many people want to mint. For example, if 5000 people minted a Wendy the buff would be ~0.02-0.05 wood

0.02-0.05 wood is still not a bad boost considering that boosts will accumulate with more SFTs being released in the future.  I would like to see what other players think about that

IvanSkazkin:
> 

I love collecting tickets. But I don't like that the tickets are not refundable if you lose. Let's say I play a lot. And I put a lot of tickets. Reading the chat, I'll probably find out that someone has won from someone who has fewer tickets and this will be disappointing.

You can do it. 30-50 first minutes can be obtained without a lottery. That is, it is necessary, for example, to collect 100-300 firewood and another 800 sunflowers + spend 50 blits. And you can mint the item. Then people will fight to get 100% nft. Otherwise, if there is only a lottery system with no possible reward. It will be sad. We must give the most active players a chance.

After 30-50 minutes of the most active players, people can participate in the lottery. This feature will also increase the cost of nft. For example, the t2 Beaver will become more expensive, because it reduces the time it takes to harvest a tree.

adamhannigan:
Agreed, whoever places the most lottery tickets should get the items.

In an ideal scenario, there should be roughly 20% people that miss out on items (with the correct prices and supply). A 4/5 success rate would lessen the frustration of missing out and/or losing lottery tickets.

I believe burning the lottery tickets regardless of a win/lose is essential gameplay. 'Risk' is one of funnest gameplay elements and incorporating it here feels like a great use case.

celpower:
Be careful what you wish for, RNG is all fun and games until it feels like pure gambling and the players are on the lose end CONSTANTLY. Million on Mars loves the gambling/RNG aspect, and there are a lot of players trying to leave the game and sell the NFTs but hardly anyone interested in buying. If I feel like this game is going in this direction as well, I would jump ship before it actually hits the iceberg.

WTFpacman:
> Be careful what you wish for, RNG is all fun and games until it feels like pure gambling and the players are on the lose end CONSTANTLY. Million on Mars loves the gambling/RNG aspect, and there are a lot of players trying to leave the game and sell the NFTs but hardly anyone interested in buying. If I feel like this game is going in this direction as well, I would jump ship before it actually hits the iceberg.

I agree so much to this Mom basicly killed their whole Playerbase with too much too bad RNG

AzaroQL:
As it was stated by Adam they are reviewing lottery version with tickets, means there is low amount of rng, only for low bidders.

Example:
1k supply, 900 players bid 10 tickets, 200 players bid 5 tickets
everyone who bid 10 tickets and only 100 of players who bid 5 tickets will get the sft

Tickets are should be obtainable through gameplay and you can decide on how much tickets you are willing to spend on each SFT.

Telknub:
If anyone here is interested and did not see it yet, I opened a discussion on #2130 on an idea on how to improve Adam's idea to reduce rng and make it more fair for everyone. Please read it and give feedback if you have any.

NotKeen:
I like the idea with the lottery tickets very much. I have always thought that those players who actually play the game should have better chances to mint items more than whales (or even some OGs who check their farm once a day, no offense :D ).

Two points of possible refinement in my eyes:
1. Consider making lottery tickets tradable - I know this kinda makes them yet another resource, but I think it would be good giving hard-working farmers the choice to use or sell their hard-earned tickets :-)
2. Do not burn the lottery tickets of the losers. This just feels wrong / very frustrating imo, sorry to say. You already lose by not having been able to mint AND also lose tickets you staked? I think that's bad. Also, if we don't burn them, then players can accumulate tickets on every lost mint and can eventually have enough to successfully mint one day. Something nice to look forward to instead of getting the double punch.

IvanSkazkin:
How about giving away tickets for cooking?
For example, 10% chance to get a ticket for cooking food. It will make people play a lot

Telknub:
@NotKeen tradeable tickets would basically throw this idea in the trash since whales would always have an advantage this way.

And I think burning the tickets is good because its a risk reward thing. If they are not burned and then returned to the players they would have a massive advantage next mint, also making the idea not work as intended, that is, having every mint being more balanced. Yes, players can like not mint 2 items and have tons of tickets for the third one, but this goes by the strategy of the player.

PR #2125, opened 2023-01-30
Fix spelling of studios

brian-man:


adamhannigan (approved):
Nice atomic commits 👍

PR #2161, opened 2023-02-02
[FIX] Fixes in bumpkin level logic

AzaroQL:
# Description

- Bug: Badge for next level with skill point was displaying incorrect level in some cases.
- Update: Display 'unspent skill points' badge and 'next level with skill point' badge together.(quality of life change, its always useful to know when you will get your next skill point)

image
image

## Type of change

# How Has This Been Tested?

Manual test + yarn test

# Checklist:


src/lib/utils/flipObject.ts
```diff
@@ -1,18 +0,0 @@
-export type AnyKey = keyof any;
-
-/**
- *  Flips your object so that values are keys and keys are values
- * @param object
- * @returns object flipped { v: k }
- */
-export function flipObject<K extends AnyKey, V extends AnyKey>(
```
AzaroQL, line 8: This was working incorrectly since we had duplicate values.

src/features/game/types/craftables.ts
```diff
@@ -991,6 +991,7 @@ export const ANIMALS: () => Record<Animal, CraftableItem> = () => ({
 /**
  * getKeys is a ref to Object.keys, but the return is typed literally.
  */
+// TODO: Dangerous function that returns array of strings, but acts like returns array of T
```
AzaroQL, line 994: TS wont catch errors if T is number how it was in levels. For example if we take the key(level) from such object - {30: 22}, and further in code we will decide to get level+1 we will get 301, instead of 31

src/features/game/lib/level.ts
```diff
@@ -1,49 +1,4 @@
-import { flipObject } from "lib/utils/flipObject";
-import { getKeys } from "../types/craftables";
-
-export type BumpkinLevel =
```
AzaroQL, line 4: If this type was purely used to keep LEVEL_EXPERIENCE and SKILL_POINTS keys exactly the same, I can create a test that will check those objects. Or maybe it makes sense to refactor those objects in to array of {level: number, exp: number, skillPoints: number} ?

spencerdezartsmith: IMO, this type is useful because it adds context to a primitive type. Changing the type to `{ [key: string]: number }` removes all context and is also confusing for some people. Even though all keys are strings under the hood, we have the ability to see these strings represented as numbers in our IDE's so it's not super clear in this instance.

The refactor to a type object might be a better solution in this case.

AzaroQL: I didnt find any place in code where we would need that tho. Those two objects are not referenced by key anywhere. But we had a lot of 'as BumpkinLevel' which hides any number under that type( for example '5993 as BumpkinLevel' is completely valid)

AzaroQL:
Removed part with refactoring.

src/features/game/lib/level.ts
```diff
@@ -151,19 +151,30 @@ export const findLevelRequiredForNextSkillPoint = (
...
-  if (Number(level) >= MAX_BUMPKIN_LEVEL) return undefined;
+  if (level >= MAX_BUMPKIN_LEVEL) {
+    return;
+  }
 
-  const currentSkillPoints = SKILL_POINTS[level];
-  const skillPointToLevels = flipObject(SKILL_POINTS);
+  let nextLevelWithSkillPoint = (level + 1) as BumpkinLevel;
+  // Skip levels that give no skill points
+  while (
```
spencerdezartsmith, line 160: This is all working well but this part of the code is not very easy to read. Can you think of a way to refactor this to be a little easier to understand?

spencerdezartsmith (changes requested):
This is all working well but some of the code was a little difficult to read. If you can just rework this piece of code to be more readable then I am happy to merge. Even if just assigning conditions to clearly named variables would go a long way. 

For example, naming this something to clearly indicate the condition you're checking against: 

```
SKILL_POINTS[nextLevelWithSkillPoint] && SKILL_POINTS[level] === SKILL_POINTS[nextLevelWithSkillPoint]
```

src/features/game/lib/level.ts
brian-man: How about this, @spencerdezartsmith?

```
```

spencerdezartsmith (approved):
LGTM

src/features/game/lib/level.ts
AzaroQL: Replaced 'while' with 'for in' loop.

brian-man: @AzaroQL My versions quite a bit simpler.  Dunno if you saw it.

AzaroQL: > 

Both methods looks good to me, yours is less lines tho.

AzaroQL:
Is there anything else that stops this PR from merging?

johannesunana:
Why has this yet to be merged into main if this has been approved already?

PR #2165, opened 2023-02-03
[FEAT] Seasonal Tickets

adamhannigan:
# Description

This PR takes player feedback from #2122 and #2130 to implement an auction system that is fairer and rewards good players.

We are planning to experiment this approach for 3 auctioneer drops - if it is successful, we will keep it. Otherwise revert.

**Goal**

- Support all timezones (24 hours)
- Less RPC bugs & high gas fees
- Promote equality (not people with best RPC connections)

**Mechanic**

When an auction opens, players have 24 hours to cast their bids.

Resources/SFL are escrowed (off chain) until the auction is over - players can not access these items. This guarantees that a player is actually intending to mint the item.

A player can select how many auction tickets they wish to use - if there is more participants than supply, then players who used the most auction tickets win. This ensures loyal players have a higher chance of minting these items. A player **cannot** change their bid.

After 24 hours, players come back to check the results of the auction.

Losers get their refund (except their auction tickets), winners can mint the item.

**Off chain actions**

The following actions occur off-chain and do not require a transaction:

- Bids (these need to be private)
- Escrowing resources
- Refunds

This means the resources are not actually burnt until the item is minted

**Flow**

_New Item_

Screen Shot 2023-02-03 at 3 19 30 pm

_Choose auction tickets_

Screen Shot 2023-02-03 at 3 19 50 pm

_Bidded_

Screen Shot 2023-02-03 at 3 23 53 pm

_Successful_

Screen Shot 2023-02-03 at 3 45 19 pm

_Unsuccessful_
Screen Shot 2023-02-03 at 3 31 50 pm

_Refunded_

Screen Shot 2023-02-03 at 3 32 51 pm

**Excludes**

1. Earning Auction Tickets in Grub Shop


brian-man:
It might be obvious but I only skimmed:
Is "casting a bid" off-chain or does it require a blockchain transaction?

Polysemouse:
Can users change their number of ticket they cast after placing a bid?

adamhannigan:
@brian-man @Polysemouse Great questions - I've just updated the description to be clearer 👍

brian-man:
Will there be after-auction statistics or discord posts giving players a feel for what bids were made? Or will it be completely blind auction to auction?

adamhannigan:
@brian-man It is hard to predict what user behaviours/strategies will arise from this. We will display the minimum requirements - not sure what extend we want to expose more results.

Once we provide too much information, the curiosity and decision making lessens for a player which takes away from the gameplay of risk

Telknub:
Really good! Thanks for taking the feedback ♥️

takanoki0723:
I suggest that those who participated in the lottery and were not selected return half of the lottery tickets, so that they will not be so disappointed and have more chances in the next lottery draw.

FernandoAvellar:
In my opinion loose all lotterys ticket will only add frustration to the player. Imagine how many effort to get this tickets and loose all trying to get an NFT will add a lot of frustration in the process. If this tickets been returned the player can at least try to win the next audiction and it will be a better experience.

adamhannigan:
Some great discussion. Thanks everyone 🤗  @takanoki0723 @FernandoAvellar 

This decision was made so we can gamify the auction feature and make it a gameplay element, instead of a chance mechanic. We needed to incorporate 'risk' in the players decision to encourage them to make the right decision. The best challenges in games are ones where there is an inherent risk involved - when players succeed the challenge, they feel much more rewarded.

An example of this is RPG games where you lose money or equipment when your character dies. This is a frustration in the short term, but once the player finally succeed it makes the challenge much more rewarding.

alankof10:
I loved
I hope it takes place

Polysemouse:
The number of withheld lottery tickets should not be visible in the API or else people can scan through all the farms and potentially see how many tickets each player deposited

henrymendezz:
I don't like this system, because I don't like luck. I prefer the effort, the sft system is very good, because it rewards the effort!

Telknub:
@henrymendezz but this system is not luck based ....
Its point is to remove all luck and make it fair for everyone

henrymendezz:
> @henrymendezz but this system is not luck based .... Its point is to remove all luck and make it fair for everyone

Lottery: public game of chance. Google definition. Greetings!

adamhannigan:
@henrymendezz That is some good feedback.

As @Telknub mentioned this is not luck based. But you are right, the word "Lottery" means luck.

We will remove that wording to prevent confusion -Changed to "Auction Tickets"

rasheed19871:
> In my opinion loose all lotterys ticket will only add frustration to the player. Imagine how many effort to get this tickets and loose all trying to get an NFT will add a lot of frustration in the process. If this tickets been returned the player can at least try to win the next audiction and it will be a better experience.

I totally agree, so we will not lose all our ticket

AzaroQL:
I agree that tickets should be burned for all parties, that way players wont bid on items that they dont want.
Its easy to place low bid on every mint thus inflating entry point for no reason if you know you wont lose anything.

PR #2167, opened 2023-02-03
[FEAT] Show more details about skills and achievements in bumpkin modal

varrt:
# Description

This PR shows more details about skills and achievements in bumpkim modal. Also The Legacy Skills has different color of title than other skills.

On mobile tooltips are disabled.

modal

## Type of change

# How Has This Been Tested?

Localy show bumpkin modal.

# Checklist:

### DIFF
--- a/src/features/bumpkins/components/AchievementBadge.tsx
+++ b/src/features/bumpkins/components/AchievementBadge.tsx
@@ -1,0 +1,45 @@
+import React, { useState } from "react";
+import { ITEM_DETAILS } from "features/game/types/images";
+import { InnerPanel } from "components/ui/Panel";
+import { AchievementName } from "features/game/types/achievements";
+import { Box } from "components/ui/Box";
+import { detectMobile } from "lib/utils/hooks/useIsMobile";
+import { PIXEL_SCALE } from "features/game/lib/constants";
+
+export const AchievementBadge: React.FC<{
+  achievement: AchievementName;
+}> = ({ achievement }) => {
+  const image = ITEM_DETAILS[achievement].image;
+  const name = achievement;
+  const description = ITEM_DETAILS[achievement].description;
+
+  const [showTooltip, setShowTooltip] = useState(false);
+  const isMobile = detectMobile();
+
+  return (
+    <div key={achievement} className="relative">
+      <div
+        className="flex justify-center m-1"
+        style={{
+          height: `${PIXEL_SCALE * 19}px`,
+          width: `${PIXEL_SCALE * 18}px`,
+          margin: `${PIXEL_SCALE * 1.5}px`,
+        }}
+        onMouseEnter={() => setShowTooltip(true)}
+        onMouseLeave={() => setShowTooltip(false)}
+      >
+        <Box image={image} />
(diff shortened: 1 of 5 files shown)


Polysemouse:
We already have a <Box/> component to show items in a box.  Perhaps you can reuse that component instead of creating a new one

varrt:
@Polysemouse I used Box component for display skills and achievements. I added size props for Box component where "medium" is default value and it's border with like before. Albo you can set size on "small" to display smaller box border

hujikol:
I think we should not add the corner border to be permanently displayed, as it is intended to be used in the hover select state

src/features/game/lib/landData.ts
```diff
@@ -243,9 +243,22 @@ const INITIAL_BUMPKIN: Bumpkin = {
     tool: "Farmer Pitchfork",
     background: "Farm Background",
   },
-  skills: {},
```
hujikol, line 246: also please omit the landData.ts on your commit as it will affect others.

Polysemouse:
> I think we should not add the corner border to be permanently displayed, as it is intended to be used in the hover select state

In that case we don't need the smaller scale select indicator for boxes, and we can maintain consistent pixel size (aka revert all changes for the Box component)

varrt:
I also reverted changes from Box.ts and set display border only for hover Box.

Polysemouse:
@varrt That looks better and the code is much cleaner!

There are some places (for skill and achievement badge) you can further improve it:

- You don't have to pass in the `isHoverBadge` to the `isSelected` prop for `Box`, because the `Box` already have it's own hover logic
- You don't have to set the width and height of the parent div (the one wiht INNER_CANVAS_WIDTH and INNER_CANVAS_HEIGHT because that the size of the div will follow the size of the `Box`.  You can also set the classname="flex justify-center" to remove the style={{}} prop completely to further simplify the code (try removing the `relative` from the classname, it's probably not needed)
- The `text-white` for skill/achievement description for the classname is not needed because white text is already enabled system wide.
- I would probably add `pointer-events-none` to the innnerpanel (hover panel) so the panel won't be blocking click events for other boxes

varrt:
@Polysemouse I used your advice. Only one thing i can't do it's remove setting width and height for my div. In Box component there are margins and i can't change it without changing Box component.

If I don't set this i have this effect:

wrong

And I want to get this effect:

good

I think it is looking better

Polysemouse:
@varrt That's right, that should be fine with the extra divs then

varrt:
@Polysemouse You mean to write one more div and move style for width and height to this div?

Polysemouse:
@varrt I think the current number of divs are good enough.  It shouldn't need additional divs on top of that

varrt:
@Polysemouse Ok, i see. So is there something I can do else?

Polysemouse:
@varrt I think you are all good!

adamhannigan:
@varrt Thanks for the contribution.

The reason this has been taking so long was that the core team was reworking the modal with Multiple Bumpkins and still has changes planned. Since these will conflict with your changes we will close this PR

PR #2183, opened 2023-02-06
[FIX] Treasure Hole Recovery

adamhannigan:
# Description

Previous calculations were relying on the day of the week to check if they can dig again.

Unfortunately this caused an error if people waited 1 week between digging treasures.


spencerdezartsmith (approved):
LGTM

spencerdezartsmith (approved):
LGTM

Rofrtd (approved):

PR #2219, opened 2023-02-10
[FEAT] Revert PWA

adamhannigan:
# Description

We need to fix some caching issues - right now revert so we can release Block Bucks


spencerdezartsmith (dismissed):

spencerdezartsmith (dismissed):

Rofrtd (dismissed):

Rofrtd (approved):
🚀

c-py (changes requested):
❌

PR #2235, opened 2023-02-13
[FEAT] Daily Rewards (On Chain DAU)

adamhannigan:
# Description

This feature introduces the concepts of "on-chain" daily rewards.

The purpose is to increase the transactions submitted by users to **increase our on chain daily active user metrics**. This metric is used specifically by platforms like DappRadar and is a useful way to climb the charts and gain publicity/gain new users.

**Unlock & Open**

To receive a Daily Reward, players must first submit an on-chain transaction. We have created an on-chain function which requires minimal gas - Roughly $0.005

We wanted to keep this as lightweight as possible to save user's gas fees and encourage them to submit a transaction.

When the transaction is submitted, the treasure chest becomes 'unlocked'. Once a chest is 'unlocked' a user can then 'open' the chest and receive a reward from our backend.

**Can players bypass the smart contract?**

To keep the smart contracts gas efficient we have not included any on-chain verification. Our aim was to keep these contracts gas efficient.

Our UI enforces players submit a transaction first - there is no business logic to prevent actors from accessing the smart contract directly, however the API enforces cooldowns + standard API botting techniques.

**Schedule**

The daily reward schedule is based on UTC timing, meaning it will become available for everyone at the same time 🚀 

## Flow

Screen Shot 2023-02-13 at 4 53 00 pm
Screen Shot 2023-02-13 at 4 53 06 pm
Screen Shot 2023-02-13 at 4 53 21 pm
Screen Shot 2023-02-13 at 4 54 57 pm
Screen Shot 2023-02-13 at 4 57 17 pm
Screen Shot 2023-02-13 at 4 57 25 pm


Polysemouse:
This is a good way to increase the DAU count, however it doesn't increase the "freshness" of the blockchain data.  Instead it would be worthwhile to try minimizing the data needed for each sync so the gas fees can be reduced.

Side note: the regular (18hrs 2mins) format is used for the blue label usually than the short one (18h 2m).

adamhannigan:
@Polysemouse Thanks for your feedback 🤗 This specifically is solving a different problem to the gas fees/storing on chain data.

We will address that separately 💪 we are already exploring ways to minimise this fee.

spencerdezartsmith:
I like it. 

I would love to get the idea of where the button should lie in the modal sorted. I should be like the second screenshot here.

Basically when adding any content to a panel the content and button would be siblings.

```
<Panel>
    <>
        <Content/>
        <Button/>
    </>
</Panel>
```

Screen Shot 2023-02-14 at 12 11 31 pm

Screen Shot 2023-02-14 at 12 18 02 pm

🦅 👁️ 😄

spencerdezartsmith (dismissed):
LGTM

PR #2237, opened 2023-02-13
[FIX] Add Treasure island item limits

mGranTY:
# Description

Treasure Island items does not have a limit set, so when players dig and get their 10th item AS-01 error appears.

BTW commons items limit should be increased, Craig should be 30 per sync.

## Type of change

# How Has This Been Tested?

localhost with landData of 11 Craig

# Checklist:


adamhannigan (approved):
LGTM

PR #2290, opened 2023-02-27
[FIX] Enable SFT withdrawals

adamhannigan:
# Description

Enable SFTs for withdrawal.

- Maneki
- Heart of Davy Jones
- Valentines Beat

**API schema**

Removed the checks on the API level - these are now only type checks, the withdraw file holds all of the logic for whether an item can be withdrawn. Too many bugs were getting caught here 😢


c-py (approved):
LGTM

PR #2308, opened 2023-03-05
[FEAT] Hayseed Hank

adamhannigan:
# Description

HaySeed Hank, the daily chore master!

Provide players with a range of quests they must complete to receive Solar Flare Tickets 🚀 

The quests act as a sort of onboarding and get progressively harder 💪 

Screen Shot 2023-03-06 at 9 02 32 am

## Business Rules

1. The quest starts when a player clicks "Start Quest" - it does not include previous progress.
2. A player cannot swap Bumpkins during a quest - they must start the same quest again (Modal warning is shown).
3. Even if a Bumpkin is swapped, the quest remains the same

### DIFF
--- a/src/features/game/events/index.ts
+++ b/src/features/game/events/index.ts
@@ -119,6 +119,11 @@ import {
 import { sellTreasure, SellTreasureAction } from "./landExpansion/treasureSold";
 import { restock, RestockAction } from "./landExpansion/restock";
 import { sellGarbage, SellGarbageAction } from "./landExpansion/garbageSold";
+import { startChore, StartChoreAction } from "./landExpansion/startChore";
+import {
+  completeChore,
+  CompleteChoreAction,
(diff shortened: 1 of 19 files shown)


4dr3naline:
when did it start?

Polysemouse:
Would players know what tasks they will get on the next day?  I.e. Players might miss out a farming task if they planted kale and have to wait for 1.5 days to harvest them

adamhannigan:
@4dr3naline On Wednesday this will become available 👍

adamhannigan:
@Polysemouse Every player faces the same order of tasks, so they can find out about it through other players.

But ideally we want this to be a surprise for players - when the quest pops up is when they should start working towards it 😄

popalexandrudan:
@adamhannigan will the daily quest be repeatable ? i'm afaraid that if it will, it my cause a lot of burn outs ^^

PR #2310, opened 2023-03-06
[FEAT] Track wallet types

adamhannigan:
# Description

Gain insights to what wallets players are using. This will help prioritise development efforts to supporting smooth wallet experiences

### DIFF
--- a/src/features/auth/lib/authMachine.ts
+++ b/src/features/auth/lib/authMachine.ts
@@ -198,12 +198,12 @@ export const authMachine = createMachine<
             cond: (context) => context.wallet === "METAMASK",
           },
           {
-            target: "connectingToWalletConnect",
-            cond: (context) => context.wallet === "WALLET_CONNECT",
-          },
-          {
             target: "connectingToSequence",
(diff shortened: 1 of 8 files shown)


src/features/auth/lib/authMachine.ts
```diff
@@ -197,14 +197,14 @@ export const authMachine = createMachine<
...
-            target: "connectingToWalletConnect",
-            cond: (context) => context.wallet === "WALLET_CONNECT",
-          },
           {
             target: "connectingToSequence",
             cond: (context) => context.wallet === "SEQUENCE",
           },
+          {
+            target: "connectingToWalletConnect",
+            cond: (context) => !!context.wallet,
```
spencerdezartsmith, line 206: Can you explain this change?

spencerdezartsmith (approved):
This looks good. Just one question about the auth change.

src/features/auth/lib/authMachine.ts
adamhannigan: @spencerdezartsmith Yes that is right - Wallet Connect will provide whichever wallet is chosen 😄

PR #2344, opened 2023-03-15
[FEAT] Node Rebalancing

adamhannigan:
# Description

WIP branch for balancing and placing resources


hujikol:
isn't this PR becoming too big?
I think we can split it into, a plot node, a mineral node, and another one for trees and fruit.

adamhannigan:
@hujikol The issue here is that we need to do the rebalance in one step (airdrop) so everyone is up to date with the right resources. This required a large refactor of all our backend types that our front-end now needs to match.

Unfortunately we cannot do one resource at a time as people would expand lands and continue to get unbalanced 😢

PR #2356, opened 2023-03-20
[FIX] Select seeds/tools in a more consistent way

AzaroQL:
# Description

Some players use market as a way to select seeds for planting, since only buy action was selecting seeds it was required for player to buy one seed before proceeding with planting. Also it causes confusion when player has one seed selected, goes through seeds in market and while doing so buys few seeds up to a limit.

I made it more consistent by setting 'shortcutItem' on seed select, not on buy action. That way nothing changes for players who select seeds from inventory and prevent miss planted seeds for players who use market.

Applied same approach to tools, just in case.

Also removed two duplicated lib files. Probably those were left out from old farm since there is nothing in parent folders.

## Type of change

# How Has This Been Tested?

Manual test

# Checklist:


AzaroQL:
@spencerdezartsmith, done.

spencerdezartsmith (approved):
Love it! 🚀

PR #2497, opened 2023-04-27
[FEAT] Dawn Breaker Decorations

adamhannigan:
# Description

The following items have been added to the list of supported decorations. They don't need to be disabled since they require items that cannot be attained until the Season starts 🚀 

Screen Shot 2023-04-27 at 10 54 27 am

Screen Shot 2023-04-27 at 1 33 40 pm

1. This PR adds support for the Seasonal Decorations + 2 bonus normal tombstone decorations.
2. The landscaping machine now supports setting a `maximum` amount of items to place - now you can also buy multiple chickens at once 🚀 

_Note: Open Sea images are in progress_


c-py (approved):
LGTM!

Tested:
1. Buying from Helios
2. Placing from chest
3. Buying & placing from design mode

Note: Helios salesperson doesn't sell the limited time decorations - only the permanent ones.

PR #2502, opened 2023-04-28
[CHORE] Grub Shop Ticket Message

adamhannigan:
# Description

Adds a temporary message explaining what will happen with the tickets. The UI isn't great, but it is only temporary and should serve the purpose.

Screen Shot 2023-04-28 at 1 53 11 pm


spencerdezartsmith (approved):
LGTM 👏🏼

PR #2507, opened 2023-04-29
[FEAT] Add Discard tab to Garbage Collector in Helios

brian-man:
# Description

Periodically players request that something be discardable.  An easy example that comes up periodically is the desire to discards seeds -- especially since the introducing of the daily chest -- so that Kuebiko is more easily withdrawable.

This change introduces the Discard tab to the Garbage Collector.  The initial discardable items are seeds, but it's trivial to add any other items to the discardable list.

## Type of change

# How Has This Been Tested?

Tested with local game using purchase and sell of sunflower, potato, and pumpkin seeds.

# Checklist:

# Screenshots

image

### DIFF
--- a/src/features/game/types/garbage.ts
+++ b/src/features/game/types/garbage.ts
@@ -60,3 +60,36 @@ export const GARBAGE: Record<GarbageName, Garbage> = {
     sellPrice: marketRate(1),
   },
 };
+
+export type DiscardName =
+  | "Sunflower Seed"
+  | "Potato Seed"
+  | "Pumpkin Seed"
+  | "Carrot Seed"
+  | "Cabbage Seed"
+  | "Beetroot Seed"
+  | "Cauliflower Seed"
+  | "Parsnip Seed"
+  | "Eggplant Seed"
+  | "Radish Seed"
+  | "Wheat Seed"
+  | "Apple Seed"
+  | "Blueberry Seed"
+  | "Orange Seed";
+
+export const DISCARD: DiscardName[] = [
+  "Sunflower Seed",
+  "Potato Seed",
+  "Pumpkin Seed",
+  "Carrot Seed",
+  "Cabbage Seed",
+  "Beetroot Seed",
+  "Cauliflower Seed",
+  "Parsnip Seed",
+  "Eggplant Seed",
+  "Radish Seed",
+  "Wheat Seed",
+  "Apple Seed",
+  "Blueberry Seed",
+  "Orange Seed",
+];
--- a/src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
+++ b/src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
@@ -1,0 +1,77 @@
+import React, { useContext, useState } from "react";
+import { useActor } from "@xstate/react";
+
+import { Box } from "components/ui/Box";
+import { Button } from "components/ui/Button";
+
+import { Context } from "features/game/GameProvider";
+import { ITEM_DETAILS } from "features/game/types/images";
+import { Decimal } from "decimal.js-light";
+
+import { DISCARD, DiscardName } from "features/game/types/garbage";
+import { SplitScreenView } from "components/ui/SplitScreenView";
+import { ShopSellDetails } from "components/ui/layouts/ShopSellDetails";
+
+export const GarbageDiscard: React.FC = () => {
+  const [selectedName, setSelectedName] = useState<DiscardName>(DISCARD[0]);
+
+  const { gameService } = useContext(Context);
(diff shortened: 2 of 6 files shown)


ShiningMinions:
Solves https://github.com/sunflower-land/sunflower-land/discussions/2486 , thank you!

src/features/game/types/garbage.ts
```diff
@@ -60,3 +60,36 @@ export const GARBAGE: Record<GarbageName, Garbage> = {
     sellPrice: marketRate(1),
   },
 };
+
+export type DiscardName =
```
spencerdezartsmith, line 64: Are these new types necessary? You could potentially use the `isSeed` type guard in your action instead. Would prefer to not have another type that duplicates items.

```
  const isSeed = (selected: InventoryItemName): selected is SeedName =>
    isFruitSeed(selected) || selected in CROP_SEEDS();
```

src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
```diff
@@ -0,0 +1,77 @@
...
+    {
+      context: { state },
+    },
+  ] = useActor(gameService);
+
+  const inventory = state.inventory;
+
+  const amount = inventory[selectedName] || new Decimal(0);
+
+  const discard = (amount = 1) => {
```
spencerdezartsmith, line 29: Is this default assignment necessary?

src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
```diff
@@ -0,0 +1,77 @@
...
+          }}
+          properties={{
+            sfl: new Decimal(0),
+          }}
+          actionView={Action()}
+        />
+      }
+      content={
+        <>
+          {DISCARD.map((name: DiscardName) => (
```
spencerdezartsmith, line 64: Here you could either use the seeds, fruit seeds constants and grab their keys or you could build a `DISCARD` constant by doing the same. It just helps us not have to manually update code in multiple places.

spencerdezartsmith (changes requested):
Nice initiative. Just a couple comments about reusing already established types and constants. 

This will require core dev work on the backend before it can be merged.

src/features/game/types/garbage.ts
brian-man: The intent was/is to have an explicit list -- like for the Sell tab -- of what is discardable.  Further, the intent (as noted in the PR description) is to provide a framework for the easy addition of other items to the discardable list.

src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
brian-man: Modeled on what the Sell tab does.  Would you like to remove the ` = 1` from both of them?

src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
brian-man: See previous reply regarding dedicated names for discardable items.

As for initially populating DISCARD via the seed constants, I'm not opposed to that.

spencerdezartsmith: Yes, I would prefer that. Maintaining multiple lists always leads to more work and things being forgotten 🙏🏼

ShiningMinions:
@brian-man FYI if you need help with this one I'm available to collab.

src/features/game/types/garbage.ts
brian-man: @Romy#7694 
How do I do this part...

```
import { CROP_SEEDS } from "features/game/types/crops";

export type DiscardName = <what goes here>;
```

I need `DiscardName` for the type for `DiscardGarbageAction`s 2nd parameter.

src/features/helios/components/garbageCollector/components/GarbageDiscard.tsx
brian-man: Updated

c-py:
@brian-man Thanks for your contribution and raising this PR! I regret to say, I don't think we will go with this right now.

The reason is that there is a cost of engineering time to maintain this API action - and without the use cases available just yet doesn't seem to be worth it. I understand for Kuebiko holder's it can be frustrating to burn though seeds to withdraw, but we don't want to maintain features for a small numbers of players in a niche circumstance.

Feel free to ping me on #coders-chat if you would like to discuss further or if there are use cases I am missing.

brian-man:
@c-py Bummer.  People have been requesting this for months for seeds and other things.

PR #2524, opened 2023-05-02
[FEAT] Buy more block bucks + gas optimisations

adamhannigan:
# Description

Network is struggling right now. The purpose of this PR is to help provide relief for gas costs.

Block Bucks now go through the new purchase flow, which is more optimised storing data on chain.

Screen Shot 2023-05-03 at 10 42 16 am


Polysemouse:
Would that mean player will hit the hoarding limit more often?

brian-man:
Before:
5 is more expensive per BB then 1
New:
15 is more expensive per BB then 5

It really seems predatory that the team is following the model of "pay us more or pay the blockchain more".  It didn't really sit well with me originally and this doubles down on it.

adamhannigan:
@Polysemouse The player will hit the hoarding limit in the same amount of time as before. When they do hit the hoarding limit, "Storing Progress On Chain" will be a reduced gas cost.

@brian-man That is fair feedback - The project does need to make revenue and all in app purchases are optional. We are still working on these prices. Gas should be the same, regardless of how many Bucks you buy so we can look at discounted prices.

brian-man:
And now the latest raises the price of 1 BB as well?  Sigh.

Old
```
 1BB $0.10/ea.
 5BB $0.15/ea.
```

New
```
 1BB $0.25/ea.
 5BB $0.198/ea.
10BB $0.175/ea.
20BB $0.1495/ea.
```

1BB went up 150%
5BB went up 32%

All (1, 5, 10, 20) are more expensive than 1BB used to be.

Worse - The commit that changed this is called 'lower prices`.

I seriously don't understand.

adamhannigan:
Hey @brian-man Thanks for the feedback. It looks like your calculations are not including the entire fee, just the raw wei value of the transaction.

Here are the estimated values inclusive of Block Chain fees.

`Current - Gas prices ~0.5 MATIC`

 1BB $0.10 + $0.5 = $0.60 each
 5BB $0.75 + $0.5 = $1.25/ 5 = $0.25 each

`New - Gas Price ~0.1 MATIC`

 1BB = $0.25 + $0.1 = $0.35 each
 5BB = $0.198 + $0.1 = $1.198/5 = $0.2396 each

A request from the community was to buy more in bulk, thus reducing gas fees. In this example, we will look at the difference for a player to acquire 20 Block Bucks.

`Current - Gas Price ~0.5 MATIC`

5 Block Bucks = $.125
5 Block Bucks = $1.25
5 Block Bucks = $1.25
5 Block Bucks = $.125
Total = $5

This averages out to $0.25 per Block Buck

`New - Gas Price ~0.1 MATIC`

20 Block Bucks = $2.99 + $0.1 = $3.09. / 20 = $0.1545 total

Let me know if this makes sense 👍

brian-man:
> Hey @brian-man Thanks for the feedback. It looks like your calculations are not including the entire fee, just the raw wei value of the transaction.

This sounds exactly like I originally pointed out as troubling:
> It really seems predatory that the team is following the model of "pay us more or pay the blockchain more".

Again, the latest does the following:
1. increases the fee for 1BB by 150%
2. sets the fee for 5, 10, and 20 BB to higher than 1BB currently costs
3. so everything is more expensive _in game fees_ that before this PR
4. adding the gas costs for the transactions on top of that just makes everything even worse, not better

Rofrtd (approved):
LGTM.

adamhannigan:
@brian-man I hear your frustration. To be completely honest, if the gas prices weren't so high we would have initially priced Block Bucks higher.

This is our main source of revenue for the project that keeps us building.

Right now there is an opportunity for players to pay less and for the project to increase its funding. We hope you understand 🙏

elias7905:
I remember @adamhannigan mentioned on stream that a possible solution to the gas issue is that block bucks will not force a sync, hence reducing gas fees. Is that still gonna be the case here?
With the idea of Silo coming out and that players can choose what resources they want to sync I think this model will be more pleasant as there isn't a forced sync anymore.

adamhannigan:
@elias7905 Yes, that is part of this change.

Buying a block buck no longer requires a sync on chain - gas prices are expected t drop significantly 🙏 

We've definitely learnt that players are preferring an opt in method and deciding which data they want to secure, instead of being forced to store everything 😄

TheDcol:
If buying a blockbuck no longer syncs, does this mean that until silo is released players can only use the store on chain option via settings to sync their resources?

poteitouweb3:
> @elias7905 Yes, that is part of this change.
> 
> Buying a block buck no longer requires a sync on chain - gas prices are expected t drop significantly 🙏
> 
> We've definitely learnt that players are preferring an opt in method and deciding which data they want to secure, instead of being forced to store everything 😄

Adam, if you will allow me to contribute...

I understand that Sunflower Land's participation in terms of block chain activity has its importance for the promotion and positioning of the project, but I understand that the web3 audience is still extremely limited and already has an audience with a very clear and definite bias that "web3 games are play to earn and I must have financial returns". 

What I mean is that we should continue our focus on reaching a web2 audience and introduce them directly to play-to-own as a technological advancement and not as a financial product (as we already set out to be/do). In this regard, focusing on providing data publicly via APIs and encouraging community tools as well as reducing fees for this sounds much more beneficial to players than block chain data. Consequently for us and for new players the game being more off-chain will give us bigger leaps. We are a web3 project with nfts, blockchain and everything else but custody is our enemy at this point.

RafaelM011:
> 

Your main focus is that everything is "more expensive" and that 1,5,10,20 BB is also "worse" than before considering Price/BB before and after...but I completely disaggre with your point of view.

Let's be honest, would you buy 20 BB using the 1BB option? if that is the case, no matter if you do it with old/new price, is it always going to be cheaper buying 20BB at once... so I see no point on your arguments tbh since it is definitely not "more expensive".

If for some reason a player can't afford 5/10/20 BB at once, I think the best would be to find their way from 1 to 5, 5 to 10 and finally 10 to 20...that way they will be saving long term

SaYaNiTo:
Only two things to say:
First, in any way you ll turn it , i ll never be ok to raise block bucks price just because of gas fees ... First you make us pay more to have more bb, then you put out more bb with an increased price. Polygon saturation is used to make price higher, well just assume that you need more cash in your company than using weird arguments. First time in my entire life that i saw a product being more expensive the more you buy some.
Second point is about hoarder limit. I can be ok to buy more block bucks (even if price is a small financial effort). But why buying 5 10 or 20 bb if i m hoardering at 40% of the tickets i bought ? I got a green amulet and it is so annoying ... That s why i m a bit angry about those block bucks. A thing that should help players brings me much problems than it helps

PR #2570, opened 2023-05-11
[FEAT] Deliveries

adamhannigan:
# Description

This PR introduces the new delivery mechanic, which aims to solve the following issues in the Grub Shop:

- More balanced orders, appropriate to a players skill level
- More variety of orders
- Engaging onboarding gameplay and daily tasks.

We will plan to start Beta Testing for this next week.

https://docs.sunflower-land.com/player-guides/deliveries

## How to test?

Click on the order board and view the orders to complete.


src/features/island/delivery/components/Orders.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const select = (id: string) => {
+    onSelect(id);
+  };
+
+  if (orders.length === 0) {
+    return <p>No orders available</p>;
+  }
+
+  const canFulfill = hasRequirements(previewOrder as Order);
```
c-py, line 73:
```suggestion
  const canFulfill = !!previewOrder && hasRequirements(previewOrder);
```

**suggestion:** It may be better to show when the next order is upcoming rather than the No orders available screen.

screenshot

c-py (changes requested):
Look awesome!

I have annotated a few issues encountered in-line. Also a minor bug with the hayseed hank tick, but this one is non blocking 😁:
hank the tank

src/features/island/delivery/components/Orders.tsx
adamhannigan: Good catch. This is a fallback scenario, realistically we should always have an order pending/preparing.

I have improved the UI though so it looks nicer 👍

c-py: Yup! `orders.length` can be 0 if there are pending orders - as the value of `orders` filters pending further up in the function.

PR #2595, opened 2023-05-30
[FEAT] Phaser Enhancements (Maybe)

c-py:
# Description

Some phaser enhancements. I've annotated in line what everything does. Keen to start a discussion on the changes.

### DIFF
--- a/src/features/world/Phaser.tsx
+++ b/src/features/world/Phaser.tsx
@@ -28,7 +28,6 @@ export const PhaserComponent: React.FC = () => {
   const { gameService } = useContext(Context);
   const [gameState] = useActor(gameService);
 
-  console.log({ b: gameState.context.state.bumpkin });
   const roomService = useInterpret(roomMachine, {
     context: {
       jwt: authState.context.user.rawToken,
@@ -38,9 +37,11 @@ export const PhaserComponent: React.FC = () => {
   }) as unknown as MachineInterpreter;
 
   const roomState = useSelector(roomService, _roomState);
-  const messages = useSelector(
-    roomService,
-    (state: MachineState) => state.context.messages
+  const messages = useSelector(roomService, (state: MachineState) =>
+    state.context.rooms[state.context.roomId]?.state.messages.map((m) => ({
+      sessionId: m.sessionId ?? "",
+      text: m.text,
+    }))
   );
 
   useEffect(() => {
@@ -86,8 +87,7 @@ export const PhaserComponent: React.FC = () => {
       parent: "game-content",
     });
 
-    // Manually inject onto scenes so we have access to it
-    (game as any).roomService = roomService;
+    game.registry.set("roomService", roomService);
   }, []);
 
   return (
@@ -99,7 +99,7 @@ export const PhaserComponent: React.FC = () => {
         onMessage={(m) =>
           roomService.send("SEND_CHAT_MESSAGE", { text: m.text ?? "?" })
         }
-        messages={messages}
+        messages={messages ?? []}
       />
       <NPCModals />
       <InteractableModals />
--- a/src/features/world/roomMachine.ts
+++ b/src/features/world/roomMachine.ts
@@ -104,6 +104,7 @@ export const roomMachine = createMachine<ChatContext, RoomEvent, RoomState>({
     rooms: {
       plaza: undefined,
       auction_house: undefined,
+      clothes_shop: undefined,
     },
     // TEMP FIELD - server will set this
     bumpkin: INITIAL_BUMPKIN,
--- a/src/features/world/roomMachine.ts
+++ b/src/features/world/roomMachine.ts
@@ -8,27 +8,24 @@ import { Bumpkin } from "features/game/types/game";
 import { INITIAL_BUMPKIN } from "features/game/lib/constants";
 import { BumpkinParts } from "lib/utils/tokenUriBuilder";
 
-type RoomSchema = any;
+export type Rooms = {
+  plaza: Room<PlazaRoomState> | undefined;
+  auction_house: Room<PlazaRoomState> | undefined;
+  clothes_shop: Room<PlazaRoomState> | undefined;
+};
+export type RoomId = keyof Rooms;
 
-type RoomId = "plaza" | "auction_house";
 export interface ChatContext {
   jwt: string;
   farmId: number;
   bumpkin: Bumpkin;
-  room?: Room<RoomSchema>;
+  rooms: Rooms;
   roomId: RoomId;
-  messages: { sessionId: string; text: string }[];
-  players: Record<
-    string,
-    {
-      x: number;
-      y: number;
-    }
-  >;
+  client?: Client;
 }
 
 export type RoomState = {
-  value: "idle" | "initialising" | "ready" | "error";
+  value: "initialising" | "joinRoom" | "ready" | "error";
   context: ChatContext;
 };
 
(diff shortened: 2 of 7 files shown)


src/features/world/Phaser.tsx
```diff
@@ -86,8 +87,7 @@ export const PhaserComponent: React.FC = () => {
       parent: "game-content",
     });
 
-    // Manually inject onto scenes so we have access to it
-    (game as any).roomService = roomService;
+    game.registry.set("roomService", roomService);
```
c-py, line 90: Small change to use the built in phaser `registry` for this global object.

src/features/world/roomMachine.ts
```diff
@@ -8,27 +8,24 @@ import { Bumpkin } from "features/game/types/game";
...
+  clothes_shop: Room<PlazaRoomState> | undefined;
+};
+export type RoomId = keyof Rooms;
 
-type RoomId = "plaza" | "auction_house";
 export interface ChatContext {
   jwt: string;
   farmId: number;
   bumpkin: Bumpkin;
-  room?: Room<RoomSchema>;
```
c-py, line 18: I have changed `room` to `rooms` to track state across multiple scenes independently.

src/features/world/roomMachine.ts
```diff
@@ -8,27 +8,24 @@ import { Bumpkin } from "features/game/types/game";
...
+  rooms: Rooms;
   roomId: RoomId;
-  messages: { sessionId: string; text: string }[];
-  players: Record<
-    string,
-    {
-      x: number;
-      y: number;
-    }
-  >;
```
c-py, line 27: From what I can tell, this information is available directly in the room state under `rooms`. I think this can be deduplicated.

src/features/world/roomMachine.ts
```diff
@@ -89,63 +88,76 @@ export type MachineInterpreter = Interpreter<
...
+  any,
+  any
 >;
 
 /**
  * Machine which handles room events
  */
 export const roomMachine = createMachine<ChatContext, RoomEvent, RoomState>({
-  initial: "idle",
+  initial: "initialising",
```
c-py, line 99: I've updated the state machine so that initialisation only sets up a single client to the backend. 

Subsequent calls to change room will not reinitialize another client.

src/features/world/roomMachine.ts
```diff
@@ -164,116 +177,67 @@ export const roomMachine = createMachine<ChatContext, RoomEvent, RoomState>({
...
-          actions: assign({
-            players: (context, event) => {
-              const newPlayers = context.players;
-
-              delete newPlayers[event.sessionId];
-
-              return newPlayers;
-            },
-          }),
-        },
```
c-py, line 276: The context this was updating has been removed, so for now these events don't need processing directly by the machine.

src/features/world/scenes/BaseScene.ts
```diff
@@ -9,24 +9,98 @@ import { INITIAL_BUMPKIN } from "features/game/lib/constants";
...
 
 export type NPCBumpkin = {
   x: number;
   y: number;
   npc: NPCName;
 };
 
-export class BaseScene extends Phaser.Scene {
+export abstract class BaseScene extends Phaser.Scene {
+  abstract roomId: RoomId;
```
c-py, line 34: This is a reference to the current room. Inheriting classes are required to implement it.

src/features/world/scenes/BaseScene.ts
```diff
@@ -9,24 +9,98 @@ import { INITIAL_BUMPKIN } from "features/game/lib/constants";
...
+          x: 300,
+          y: 300,
+          isCurrentPlayer: true,
+          clothing: INITIAL_BUMPKIN.equipped,
+        });
+
+        this.readonly = true;
+      }
+    };
+  }
```
c-py, line 102: I've defined these listeners in the constructor so they don't get reinitialised when `create()` runs for the second or third times.

src/features/world/scenes/BaseScene.ts
```diff
@@ -163,66 +232,15 @@ export class BaseScene extends Phaser.Scene {
...
-
-    // this.createPlayer({
-    //   x: 300,
-    //   y: 300,
-    //   isCurrentPlayer: true,
-    // });
+    this.roomService.off(this.eventListener);
+    this.roomService.off(this.transitionListener);
+    this.roomService.onEvent(this.eventListener);
+    this.roomService.onTransition(this.transitionListener);
```
c-py, line 238: This removes listeners before adding new listeners. This is to attempt to prevent double listeners getting attached to the machine. This was causing double events.

src/features/world/scenes/BaseScene.ts
```diff
@@ -163,66 +232,15 @@ export class BaseScene extends Phaser.Scene {
...
 
     // Connect to Room
-    this.roomService.send("CONNECT");
+    this.roomService.send("CHANGE_ROOM", {
+      roomId: this.roomId,
+    });
```
c-py, line 243: The room change is performed in the `create` method. This has been done to make sure the listeners are attached before a room switch occurs.

src/features/world/scenes/BaseScene.ts
```diff
@@ -265,12 +283,7 @@ export class BaseScene extends Phaser.Scene {
             this.cameras.main.on(
               "camerafadeoutcomplete",
               () => {
-                console.log("Fade complete");
-                this.roomService.send("CHANGE_ROOM", {
-                  roomId: warpTo,
-                });
-
-                this.game.scene.switch(this.scene.key, warpTo);
+                this.scene.start(warpTo);
```
c-py, line 286: `this.scene.start` shuts down the previous scene, and starts a new scene. I think previously the plaza was hidden underneath the auction house, and was still receiving events. I believe this change stops both scenes rendering at the same time.

src/features/world/scenes/BaseScene.ts
```diff
@@ -9,24 +9,98 @@ import { INITIAL_BUMPKIN } from "features/game/lib/constants";
...
+          this.playerEntities[sessionId]
+        ) {
+          this.playerEntities[sessionId].speak(text);
+        }
+      }
+
+      if (event.type === "PLAYER_JOINED") {
+        const { sessionId, x, y, clothing, roomId } = event as PlayerJoined;
+
+        if (roomId !== this.roomId) return;
```
c-py, line 59: Events now receive a room ID. This allows scenes to filter events that are not for them.

The alternative to this is to deregister the listeners, but I think allowing filtering is a bit more flexible.

src/features/world/roomMachine.ts
adamhannigan: This feels much better than trying to fire internal machine events and storing it.

src/features/world/roomMachine.ts
adamhannigan: Love this!

src/features/world/scenes/BaseScene.ts
adamhannigan: Since we deregistered the listeners above in what scenario does this get called?

adamhannigan (approved):
LGTM. Just a few non blocking questions

src/features/world/scenes/BaseScene.ts
c-py: The listener doesn't get de-registered when the scene is shutdown. That deregistration call only happens during `create` - which would be called when the scene is restarted. It's only really there to ensure `onEvent` doesn't get attached twice.

PR #2601, opened 2023-05-31
[FEAT] More Phaser Fixes

c-py:
Fixes commented in line

### DIFF
--- a/src/features/world/scenes/BaseScene.ts
+++ b/src/features/world/scenes/BaseScene.ts
@@ -62,7 +62,6 @@ export abstract class BaseScene extends Phaser.Scene {
 
         if (!room) return;
 
-        console.log("Player joined", sessionId, x, y, clothing);
         const player = this.createPlayer({
           x,
           y,
@@ -77,11 +76,7 @@ export abstract class BaseScene extends Phaser.Scene {
 
         if (roomId !== this.roomId) return;
 
-        const entity = this.playerEntities[sessionId];
-        if (entity) {
-          entity.destroy();
-          delete this.playerEntities[sessionId];
-        }
+        this.destroyPlayer(sessionId);
       }
     };
 
@@ -115,6 +110,8 @@ export abstract class BaseScene extends Phaser.Scene {
   customColliders?: Phaser.GameObjects.Group;
 
   cursorKeys: Phaser.Types.Input.Keyboard.CursorKeys | undefined;
+  wasd: Phaser.Types.Input.Keyboard.CursorKeys | undefined;
+
   inputPayload = {
     left: false,
     right: false,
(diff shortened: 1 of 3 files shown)


src/features/world/scenes/BaseScene.ts
```diff
@@ -229,7 +226,18 @@ export abstract class BaseScene extends Phaser.Scene {
...
+      this.cursorKeys = this.input.keyboard.createCursorKeys();
+      this.wasd = this.input.keyboard.addKeys({
+        up: Phaser.Input.Keyboard.KeyCodes.W,
+        down: Phaser.Input.Keyboard.KeyCodes.S,
+        left: Phaser.Input.Keyboard.KeyCodes.A,
+        right: Phaser.Input.Keyboard.KeyCodes.D,
+        space: Phaser.Input.Keyboard.KeyCodes.SPACE,
+        shift: Phaser.Input.Keyboard.KeyCodes.SHIFT,
+      }) as Phaser.Types.Input.Keyboard.CursorKeys;
+    }
```
c-py, line 239: WASD support, it feels so natural to use WASD.

src/features/world/scenes/BaseScene.ts
```diff
@@ -403,38 +419,42 @@ export abstract class BaseScene extends Phaser.Scene {
...
-      console.log(sessionId);
       const entity = this.playerEntities[sessionId];
 
-      const position = room.state.players.get(sessionId);
+      // Skip if the player hasn't been set up yet
+      if (!entity.active) return;
 
-      if (!position) {
-        return;
-      }
```
c-py, line 420: This was a bug causing to the render loop if any position was undefined. This occurs when we have stale Bumpkin Containers in the list. I have updated this loop to rely on the game state rather than local state.

src/features/world/scenes/BaseScene.ts
```diff
@@ -403,38 +419,42 @@ export abstract class BaseScene extends Phaser.Scene {
...
+    Object.keys(this.playerEntities).forEach((sessionId) => {
+      if (!room.state.players.get(sessionId)) {
+        this.destroyPlayer(sessionId);
+      }
+    });
 
-      if (sessionId === room.sessionId) continue;
+    // Render current players
+    room?.state.players.forEach((player, sessionId) => {
+      if (sessionId === room.sessionId) return;
```
c-py, line 434: This loop now relies on game state instead of local state. it seems like this state can get a little bit corrupted when restarting scenes. 

It also deletes any stale players from `playerEntities`.

src/features/world/scenes/BaseScene.ts
```diff
@@ -353,10 +359,20 @@ export abstract class BaseScene extends Phaser.Scene {
...
+    // to the client
+    if (
+      player?.x !== this.currentPlayer.x ||
+      player?.y !== this.currentPlayer.y
+    ) {
+      this.roomService.send("SEND_POSITION", {
+        x: this.currentPlayer.x,
+        y: this.currentPlayer.y,
+      });
+    }
```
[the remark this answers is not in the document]
adamhannigan: Nice, any thoughts on some other debounced/timed behaviour with this?

E.g Only sending your position every 0.1 secs.

adamhannigan (approved):
They will put a statue of Craig in the Plaza. The almighty creator of Bumpkin World ✅

PR #2628, opened 2023-06-15
[FEAT] Golden Snout

adamhannigan:
# Description

Sir Goldensnout will be an elusive item that will help launch the Auction system. It produces manure to improve surrounding crops.

Area of Impact Item (10 surrounding crops will receive +0.5 yield)


c-py (approved):
LGTM! Might just need to add this into `removeables.ts` to restrict digging when crops are placed.

PR #2631, opened 2023-06-16
[FIX] Prevent selected collectible from appearing under others

c-py:
Requested by @Polysemouse 

# Before

before

# After

after


Polysemouse:
@c-py thanks!  This makes it much easier to move around chickens/crops in packed areas

adamhannigan (approved):

PR #2637, opened 2023-06-19
[FIX] Bypass cache on error when fetching bumpkins

c-py:
# Description

This change adds a second attempt to fetch a bumpkin URL if there is an error. This is a safety precaution in case incorrect CORS headers get cached on the infrastructure layer.

Under normal circumstances this should never really happen, but unfortunately we have a bunch of assets already incorrectly cached.

Below is a screenshot of this in action:

1. First we request the images and get a CORS error
2. We attempt to fetch the image by putting a query string on the asset
3. This is successful so this is the image we use in bumpkins. Note that this second request is served from the disk cache.

screenshot

# How Has This Been Tested?

localhost

# Checklist:

### DIFF
--- a/src/features/bumpkins/actions/buildNPCSheets.ts
+++ b/src/features/bumpkins/actions/buildNPCSheets.ts
@@ -88,6 +88,7 @@ export async function buildNPCSheets(request: Request): Promise<Response> {
       },
     };
   } catch {
+    // Since these are not real NFTs, prepend fake ID and version
     const validName = `0_v1_${tokenUri}`;
     const sheets = await buildNPCSheetsRequest(validName);
     return {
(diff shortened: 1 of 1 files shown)


src/features/bumpkins/actions/buildNPCSheets.ts
```diff
@@ -44,39 +43,55 @@ const URL =
...
+const getImage = async (url: string) => {
+  let image: string;
 
-  await Promise.all(
-    SHEET_TYPES.map(async (sheetType) => {
-      const url = `${URL}/${sheetType}/${tokenUri}.webp`;
+  try {
+    image = await new Promise((resolve, reject) => {
       const img = new Image();
+      img.crossOrigin = "anonymous";
```
c-py, line 52: This is important so that the XHR request sends CORS headers.
Otherwise it seems we might get an incorrect cached image in our disk cache. Not sure exactly why that happens.

adamhannigan (approved):
LGTM. Nice work ✅

PR #2644, opened 2023-06-22
[FIX] Add missing Hoot boost

brian-man:
# Description

Adds yield boost for Hoot collectible.

## Type of change


c-py (approved):
LGTM, Adam has added this into the backend as well!

PR #2664, opened 2023-07-01
[FIX] Update tool SFL cost tests for halvings

brian-man:
# Description

Two of the axe purchase tests had hardcoded SFL cost values that will not scale properly with future halvings.

## Type of change


adamhannigan:
As per announcement these tools will no longer be needed. A huge thanks for the initiative 🤗

PR #2694, opened 2023-07-12
[FIX] OKX refresh

adamhannigan:
# Description

A few culprits which may of been causing a screen refresh:

- Clicker detection
- CHAIN_CHANGED being called even when chain is the same


c-py (approved):
LGTM!

PR #2699, opened 2023-07-17
[FIX] more lint issues

brian-man:
# Description

Addresses more lint issues.

## Type of change

### DIFF
--- a/src/features/auth/components/CreateFarm.tsx
+++ b/src/features/auth/components/CreateFarm.tsx
@@ -58,7 +58,7 @@ export const CreateFarm: React.FC = () => {
   );
 
   const isLoading = createFarmState.matches("loading");
-  const hasEnoughMatic = createFarmState.matches("hasEnoughMatic");
+  //const hasEnoughMatic = createFarmState.matches("hasEnoughMatic");
 
   if (isLoading) {
     return (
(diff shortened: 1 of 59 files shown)


c-py (changes requested):
Awesome thanks @brian-man !

Would it be possible to undo the changes of `const send = useActor(authService)[1];` back to `const [,send] = useActor(authService);`

I understand there was a chat in the coders chat about this. It's just a preference to use the array destructing format as that's what all the rest of the code does. Both approaches are valid.

---

Also, just for reference, although `action` and `createdAt` are not used in the frontend, these are still required by the backend. If an `action` or `createdAt` is not passed to to the gameMachine, it will fail with an HTTP 5XX error when the save goes to the backend. It's fine to omit them from the frontend actions, but they will still need to be sent to the backend. They currently are still being sent so no changes needed here, this message is just for information.

brian-man:
> Also, just for reference, although `action` and `createdAt` are not used in the frontend, these are still required by the backend. If an `action` or `createdAt` is not passed to to the gameMachine, it will fail with an HTTP 5XX error when the save goes to the backend. It's fine to omit them from the frontend actions, but they will still need to be sent to the backend. They currently are still being sent so no changes needed here, this message is just for information.

Thanks for the heads up.

brian-man:
> Would it be possible to undo the changes of `const send = useActor(authService)[1];` back to `const [,send] = useActor(authService);`

Will do.

> I understand there was a chat in the coders chat about this. It's just a preference to use the array destructing format as that's what all the rest of the code does. Both approaches are valid.

If the "unreferenced _" is considered valid, then we should disable the warning complaining about it.

adamhannigan:
Closing for inactivty. Fixing the lints is a noble activity - right now the code base has drifted a lot since initial opening.

PR #2704, opened 2023-07-18
[CHORE] Web3 Connect Strategy Factory

ShiningMinions:
# Description

- Refactors authMachine to use strategy factory pattern in attempt to make adding/removing web3 wallet support much more scale-able and maintainable in the long run.
- When adding a new supported web3 provider in the future, simply add new entry to "Web3SupportedProviders" type, then follow typescript errors to easily implement new strategy. No modifications needed to authMachine.ts.

# Out of Scope

- There are several other pre-existing issues I found when refactoring this logic. The goal of this PR is to simply moving the already-existing logic to a new pattern. This PR is not going to fix these pre-existing issues (but a future PR can):

1. WalletConnect not testable locally due to missing testnet projectId.
2. PhantomWallet is hi-jacking sequence login when sequence option is pressed (possibly other wallets do this too).
3. (related to previous) Many connect strategies are looking for `window.ethereum` object when they don't have to be. Recommend a more explicit approach for detecting providers to reduce collisions. For example: "Connect to Metamask" should not work for every provider that injects `window.ethereum`, and should only work explicitly for metamask wallets.
4. (related to previous) I recommend adding a state in authMachine "wrongChain", which presents users from proceeding until they are on correct chain, and coupled with a "switch chain" button in UI. This will easily allow users to change their chain when their wallet makes it difficult for them to do so (i.e. Phantom Wallet).

## Type of change

Chore

# How Has This Been Tested?
- Locally. This was difficult for me to test, so I recommend coreDev double-checks and test.
- Could not test Walletconnect locally as I don't have the projectID for testnet.
- Did not test OKX.

# Checklist:


c-py (approved):
Awesome work @ShiningMinions ! 👍

Will merge this now then rebase the crypto.com changes onto this

PR #2709, opened 2023-07-20
[FIX] Add intermediate step for Sequence

c-py:
This adds back the intermediate step for sequence.

This is required so that a pop up has a better chance of showing up in chrome.

### DIFF
--- a/src/lib/blockchain/wallet.ts
+++ b/src/lib/blockchain/wallet.ts
@@ -211,7 +211,7 @@ export class Wallet {
           symbol: "MATIC",
           decimals: 18,
         },
-        rpcUrls: ["https://matic-mumbai.chainstacklabs.com"],
+        rpcUrls: ["https://rpc.ankr.com/polygon_mumbai"],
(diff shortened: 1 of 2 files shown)


src/lib/blockchain/wallet.ts
```diff
@@ -211,7 +211,7 @@ export class Wallet {
           symbol: "MATIC",
           decimals: 18,
         },
-        rpcUrls: ["https://matic-mumbai.chainstacklabs.com"],
+        rpcUrls: ["https://rpc.ankr.com/polygon_mumbai"],
```
c-py, line 214: Chainstacklabs seems to be down at the moment

adamhannigan, line 214: It would be nice if we had 3-4 fallback URLs to cycle through ;)

adamhannigan (approved):
LGTM ✅

src/lib/blockchain/wallet.ts
c-py: This is a good idea 👍

PR #2719, opened 2023-07-23
Draft/exciting blackburn

JCtapuk:
fix format lint


brian-man:
@JCtapuk It looks like this is closed but if the shortcut stuff gets added to chickens, please make sure you exclude it for people that have Gold Egg placed.  See my other PR that addressed the Foreman Beaver concerns with the shortcut for axes.

PR #2723, opened 2023-07-25
Enable withdraw for Eggplant in front-end

brian-man:


adamhannigan:
@brian-man Thanks. I've just included this in the API.

Once that is live we can merge this in 🙏

PR #2724, opened 2023-07-25
Change Testnet release label to commit hash

brian-man:
# Description

The `VITE_RELEASE_VERSION` for Testnet value of `refs/heads/main` is not very helpful.

This change uses the SHA hash of the commit instead.  With this hash, developers and testers can immediately know which front-end code is being used for testnet.


adamhannigan (approved):
Great suggestion 👏

brian-man:
image

PR #2732, opened 2023-07-26
[FIX] Checking the number of axes in the pop-up

JCtapuk:
When pointing scolded in the presence of an axe is now fixed. Thanks to found by @c-py 
Connected with the old PR https://github.com/sunflower-land/sunflower-land/pull/2719


c-py (approved):
LGTM! Thanks @JCtapuk !

PR #2767, opened 2023-08-02
Enable withdraw for DB TI boost collectibles in front-end

brian-man:
# Description

Enables withdraw of Foliant and Genie Lamp in front-end.


spencerdezartsmith (approved):
Have added changes to the BE. Will merge when it has been merged.

PR #2778, opened 2023-08-06
[FIX] Add Dawn Breaker Ticket to Garbage Collector NPC

brian-man:


spencerdezartsmith (approved):
Thanks! Added to BE. Will merge once that is released 🙏🏼

PR #2794, opened 2023-08-09
[FEAT] Delivery cycles

adamhannigan:
# Description

Based on player feedback, we are adjusting the deliveries to cycle based on the UTC date (similar to chores). Players will have the entire day (UTC) to complete the delivery. Once complete, it will stay in a 'complete' state until the next day.

This PR will introduce the following improvements:

- Deliveries reset for all players at same time (consistent)
- Players no longer get duplicate NPCs (e.g. multiple Tywins)

These changes aim to simplify the system & make it more predictable for players. This will lead to more stability and consistency in the feature. This will be useful in launching new NPCs + Content for players.

Screenshot 2023-08-09 at 3 33 48 pm
Screenshot 2023-08-09 at 3 33 53 pm
Screenshot 2023-08-09 at 3 46 53 pm

## Rollout

Dates to be decided, currently placeholder dates.

### DIFF
--- a/src/features/game/events/landExpansion/deliver.ts
+++ b/src/features/game/events/landExpansion/deliver.ts
@@ -24,6 +24,31 @@ type Options = {
   action: DeliverOrderAction;
 };
 
+export const BETA_DELIVERY_END_DATE = new Date("08-14-2023");
+export const DELIVERY_END_DATE = new Date("08-08-2023");
+export function canGenerateDeliveries({
+  game,
+  now,
+}: {
+  game: GameState;
+  now: number;
+}) {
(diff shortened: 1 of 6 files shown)


spencerdezartsmith (commented):
I think the UI looks 🤩

brian-man:
If the delivery is for Kale and the player does not have Kale then, even if they "see" the delivery "immediately" at reset they have no chance to complete it.  Am I missing something?

adamhannigan:
@brian-man Thanks for checking out the PR.

Only orders that are completed will be changed automatically at midnight.

Otherwise, once you complete the order it will then load the next one instantly (for that day)

PR #2809, opened 2023-08-14
[FIX] Baba Ganoush

adamhannigan:
# Description

The spelling change also changed the unique string ID of the resource. This was causing inconsistencies with the API.

To change the name, we would need a data migration first


spencerdezartsmith (approved):
LGTM

PR #2817, opened 2023-08-16
[FIX] Switch network on crypto com

adamhannigan:
# Description

By default we were using `window.ethereum` to switch networks. If you had both MetaMask and CryptoCom wallet, it would always open Metamask.

This fix ensures we use the provider that was selected.

Also updated link to one they provided.


spencerdezartsmith (approved):
Works well 🚀

PR #2907, opened 2023-09-18
[FIX] Gold Egg has no wheat requirement and need for shortcut

brian-man:
When the Gold Egg is "built" on a farm, both the inventory check for wheat and the auto-shortcutting of wheat should be bypassed.

This is similar to the Foreman Beaver fix when auto-shortcutting axes was implemented.


spencerdezartsmith (approved):
🦅 👁️

PR #2927, opened 2023-09-27
Disable on-load buds advertisement

brian-man:


c-py (approved):
Great work from a certified contributor ☕

PR #2936, opened 2023-09-30
[FIX] Locking for fruits is too restrictive

brian-man:
# Description

The `areFruitsGrowing` and `areAnyFruitsGrowing` functions were incorrect for fruits patches that are ready to be harvested.  As a result, any fruit boost was remaining farm-locked until all patches were razed down to empty soil.

This change updates those functions so that boost items -- importantly buds -- don't get locked on farms for 3+ days.
 
## Type of change


elias7905:
image
Ability to remove bud when fruit is ready to be harvested check ✅

spencerdezartsmith (approved):
❤️ 🍓

PR #2943, opened 2023-10-03
[FIX] Locking for crops is too restrictive

brian-man:
# Description

The functions used for locking based on crops were requiring empty plots -- i.e. not allowing "ready to be harvested" plots.

This change updates those functions so that boost items -- importantly buds -- don't get excessively locked on farms.

## Type of change


elias7905:
Before
image

After
image

spencerdezartsmith (approved):
👏🏼 👏🏼 🌻

PR #2959, opened 2023-10-09
[FIX] Trigger auto-save when showing TravelModal

brian-man:


elias7905:
@adamhannigan I was testing for @brian-man and we couldn't seem to get the autosave to work when we clicked on the world button, do you know where we went wrong?

src/features/island/hud/components/deliveries/TravelModal.tsx
```diff
@@ -29,6 +29,7 @@ export const TravelModal: React.FC<Props> = ({ isOpen, onClose }) => {
         centered
         show={isOpen}
         onHide={onClose}
+        OnShow={() => gameService.send("SAVE")}
```
elias7905, line 32: It was a capitalisation issue
"OnShow" change to "onShow"

brian-man: Weird.  I quadruple checked that.  Fixing now.

adamhannigan (approved):
Nice addition 👏 

For context, we have an upcoming task that is dedicated to World Modal UX + Fixes 😄 We will be shipping a lot more improvements

PR #2970, opened 2023-10-12
[FIX] Cannot load farm if fish not in ITEM_DETAILS

brian-man:
# Description

My testnet farm won't load.

I suspect it's because the `undefined` here:
```
        "fishing": {
            "wharf": {
                "castedAt": 1697134118506,
                "caught": {
                    "undefined": 1
                },
                "chum": "Potato"
            }
        },
```

But I also saw a similar error message (but not blocking farm load) after catching `Fish B`.

This change attempts to make the code generally more resilient so that farm loading isn't broken.

## Type of change


adamhannigan (approved):
Nice improvement to catch these errors 👏 

I will investigate what happened on the API side 👀

PR #2979, opened 2023-10-15
[FIX] Withdraw requirements for Angel/Devil Wings

brian-man:
# Description

Angel Wings and Devil Wings really should follow the same withdraw rules as Green Amulet.

This change updates Angel and Devil Wings to match.

## Type of change


spencerdezartsmith (approved):
Nice one

PR #2983, opened 2023-10-16
Remove unused FERTILISE_FRUIT_ERRORS.FRUIT_EXISTS

brian-man:
# Description

I was about to update the text to fix the grammar, but noticed that the value is no longer used.


adamhannigan (approved):
:shipit:

PR #2984, opened 2023-10-17
[FIX] Remove removal restrictions for composters

brian-man:
# Description

An alpha version of composters had AOE impacts on surrounding plots.  When the AOE impact was removed, the removal restrictions were not lifted/removed.

This change lifts the restrictions.

## Type of change


spencerdezartsmith (approved):
Nice one

PR #2988, opened 2023-10-17
[FIX] [TESTNET] Ignore buds with id > 5000

brian-man:
# Description

On Testnet, buds above 5000 were minted but metadata does not exist for them.  This breaks the deposit UI.

## Type of change


adamhannigan (approved):
LGTM

PR #2997, opened 2023-10-19
[FEAT] Support dynamic inputs

adamhannigan:
# Description

Worms are picky eaters and don't like eating the same produce twice!

This PR adds dynamic composter requirements each time.


src/features/game/events/landExpansion/collectCompost.test.ts
```diff
@@ -82,16 +82,16 @@ describe("collectComposterProduce", () => {
     ).toThrow("Compost is not ready");
   });
 
-  it("removes the Compost from the building", () => {
-    const basicComposter: PlacedItem = {
+  it("removes the Produce from the building", () => {
```
brian-man, line 85: Romy and I had this discussion a few times.

"Produce" means "crops brought to the grocery store".  The composters _consume_ that rather than create it.

The previous naming was better.

PR #3007, opened 2023-10-19
[FIX] Error when attempting to apply Fruitful Blend on Fruit Trees that are already grown

elias7905:
# Description

- Added Error check for when Fruit Tree is harvested at least once

Fixes Error when attempting to apply Fruitful Blend on Fruit Trees that are already grown as reported [here](<https://discord.com/channels/880987707214544966/927213157905223730/1164563775664423043>)

## Type of change

# How Has This Been Tested?

# Checklist:


adamhannigan:
@elias7905 We give a lot of flexibility when users can apply the fertilisers. In this case I don't think we should block someone if they wanted to apply the fertiliser after harvesting fruit. It isn't optimal, but they can still do it if they please.

The restriction on the backend is preventing players from applying fertilisers if a fruit is ready to be harvested. It wouldn't make much sense if the fertiliser was applied to a ready to be collected piece of fruit.

elias7905:
> @elias7905 We give a lot of flexibility when users can apply the fertilisers. In this case I don't think we should block someone if they wanted to apply the fertiliser after harvesting fruit. It isn't optimal, but they can still do it if they please.
> 
> The restriction on the backend is preventing players from applying fertilisers if a fruit is ready to be harvested. It wouldn't make much sense if the fertiliser was applied to a ready to be collected piece of fruit.
> 
> From a gameplay experience POV, what do you think makes sense here?

@adamhannigan I agree that we should be allowed to apply fertilisers whenever we like regardless of which stage the fruit tree is at (growing/fully grown). However, there's currently a bug where if the fruit tree is in a fully grown state, and the user tries to apply the fertiliser on that fruit tree, the game throws an error at the user when it tries to save. 
image

If the intention is to let the player apply the fertiliser regardless of growth state, then there might be an error in the backend code that's preventing the player from applying the fertiliser to the fruit tree in fully grown state

adamhannigan:
Thanks @elias7905 for the clarification.

The issue has been spotted in the Backend and a fix is going live 👍 I will close this PR 😄

PR #3045, opened 2023-10-27
[FEAT] Catch the Kraken

adamhannigan:
# Description

This PR includes the fish mini-game + Kraken Tentacle exchanging. When a difficult fish is faced, the mini-game will pop up. **The Kraken will now appear through the normal fishing mechanic** 

The Kraken Tentacle is the resource you can catch while fishing. I have changed the other collectible from "Kraken Tentacle" -> "Kraken Head"

Each week player's can earn 8 Kraken Tentacles and deliver them to Shelly in exchange for Mermaid Scales.

Visit the Kraken at the beach to discover what food it is craving.

Screenshot 2023-10-30 at 11 21 34 am
Screenshot 2023-10-30 at 11 20 53 am

## How to test?

1. Run locally and hardcode the fish you want to catch. Head to Reelin roy and cast your rod

    if (!CONFIG.API_URL) {
      setTimeout(() => {
        fishing.wharf = { castedAt: 10000, caught: { "Kraken Tentacle": 1 } };
      }, 1000);
    }

2. Run against API and try catch a fish. During Witches Eve, we will automatically trigger the challenge when someone catches an Anchovy, Red Snapper or Tuna.


src/features/island/fisherman/FishingChallenge.tsx
```diff
@@ -0,0 +1,339 @@
...
+          <img
+            src={ringDot}
+            style={{
+              position: "absolute",
+              width: "24px",
+              top: "1px",
+              left: "2px",
+              rotate: "137deg",
+            }}
+          />
```
c-py, line 302:
```suggestion
          <img
            src={ringDot}
            style={{
              position: "absolute",
              width: "24px",
              top: "1px",
              left: "-1px",
              rotate: "137deg",
            }}
          />
```

This appears to be off center from the parent div - which gives a bias when the green bar is rotating counter clockwise

bias

src/features/island/fisherman/FishingChallenge.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+  const barHeight = 10;
+
+  const reel = () => {
+    const greenBarAngle = extractAngle(greenBarProps.transform.get());
+    const tentacleAngle = extractAngle(tentacleProps.transform.get());
+
+    const degreeDifference = Math.abs(
+      normaliseAngle(greenBarAngle) - normaliseAngle(tentacleAngle)
+    );
```
c-py, line 158:
```suggestion
    const degreeDifference = Math.abs(
      ((normaliseAngle(greenBarAngle) - normaliseAngle(tentacleAngle) + 180) %
        360) -
        180
    );
```

I think this may return failed results around zero degrees. For example if the fish is at `5` deg and the green bar is at `358` degrees, the difference will be `353`

src/features/island/fisherman/FishingChallenge.tsx
```diff
@@ -0,0 +1,339 @@
...
+      if (misses >= 2) {
+        onMiss();
+      }
+
+      changeDirection(castTo);
+      setAttempts((prev) => [...prev, "miss"]);
+    },
+    config: {
+      duration: getSpeed(attempts.length),
+    },
```
c-py, line 135:
```suggestion
    config: {
      duration: 5000,
      mass:0,
      tension:0,
      friction:0
      
    },
```

Consider un-setting the default spring configuration on the green bar

c-py (changes requested):
LGTM - just some feedback that may need to be addressed

PR #3053, opened 2023-10-29
[FIX] Add missing bait to wharf

brian-man:


adamhannigan:
Thanks @brian-man. This will need some API additions.

What is the main use case for storing the bait that was used?

brian-man:
> Thanks @brian-man. This will need some API additions.
> 
> What is the main use case for storing the bait that was used?

Same reason to include the chum info:
(1) display in the UI for better UX
(2) support experimentation

For the latter, motivated people will remember / write-down anyway so this just reinforces that experimentation is intentionally a goal and part of the design of the fishing feature.

adamhannigan:
@brian-man Good point - I've added this to our backend tasks 👍

brian-man:
> @brian-man Good point - I've added this to our backend tasks 👍

@adamhannigan 
How's the timing for this looking?

PR #3054, opened 2023-10-29
[FIX] LiveProgressBar percentage attribute

brian-man:
The percentage attribute was being calculated incorrectly resulting in composters showing an "always empty" progress bar.


adamhannigan (approved):
Good catch! 😱

PR #3072, opened 2023-10-31
[FIX] Order CHUM_AMOUNTS and CHUM_DETAILS

brian-man:
# Description

- Updates CHUM_AMOUNTS to match normal crop ordering.
- Updates CHUM_DETAILS order to match CHUM_AMOUNTS.


adamhannigan (approved):
Thanks! I knew something was off 🙏

PR #3094, opened 2023-11-02
[FIX] Sort order of items for plaza trading

brian-man:
This change fixes the resources in the plaza trade UI to match the order in the Goblin Retreat Trade Balloon.


adamhannigan (dismissed):
Nice idea 👏  @brian-man Just a small conflict with a recent commit

adamhannigan (approved):

PR #3097, opened 2023-11-03
[FIX] Milestone 'Claim reward' button is disabled on mainnet

brian-man:
# Description

Fishing milestone rewards were disabled on mainnet as extra protection during beta period, but enablement was overlooked on feature launch.

## Type of change


adamhannigan (approved):
😱 Thanks!

PR #3099, opened 2023-11-05
Enforce bumpkin level requirement for buildings

brian-man:
# Description

After installing a bumpkin, expanding, and placing a building, players could freely remove the bumpkin (and replace it with a lower-level bumpkin) while continuing to use the building.

This change now enforces the bumpkin level requirements for buildings.

## Type of change

### DIFF
--- a/src/features/island/buildings/components/building/kitchen/Kitchen.tsx
+++ b/src/features/island/buildings/components/building/kitchen/Kitchen.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from "react";
+
 import classNames from "classnames";
 
 import kitchen from "assets/buildings/kitchen.png";
@@ -72,7 +73,7 @@ export const Kitchen: React.FC<Props> = ({
 
   return (
     <>
-      <BuildingImageWrapper onClick={handleClick} ready={ready}>
+      <BuildingImageWrapper name="Kitchen" onClick={handleClick} ready={ready}>
         <img
           src={kitchen}
           className={classNames("absolute pointer-events-none bottom-0", {
--- a/src/features/game/expansion/lib/buildingRequirements.ts
+++ b/src/features/game/expansion/lib/buildingRequirements.ts
@@ -1,0 +1,26 @@
+import { BuildingName, BUILDINGS } from "features/game/types/buildings";
+import {
+  EXPANSION_REQUIREMENTS,
+  Land,
+} from "features/game/expansion/lib/expansionRequirements";
+import { BumpkinLevel } from "features/game/lib/level";
+
+export function getBuildingBumpkinLevelRequired(
+  name: BuildingName,
+  index?: number
+): number {
+  let requiredExpansionLevel = 1;
+  const blueprint = BUILDINGS()[name];
+  if (blueprint) requiredExpansionLevel = blueprint[index ?? 0].unlocksAtLevel;
+  return (
+    EXPANSION_REQUIREMENTS[requiredExpansionLevel as Land]?.bumpkinLevel ?? 1
+  );
+}
+
+export function isBuildingEnabled(
+  bumpkinLevel: BumpkinLevel,
+  name: BuildingName,
+  index?: number
+): boolean {
+  return bumpkinLevel >= getBuildingBumpkinLevelRequired(name, index);
+}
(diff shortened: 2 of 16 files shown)


src/features/island/buildings/components/building/kitchen/Kitchen.tsx
```diff
@@ -30,6 +34,13 @@ export const Kitchen: React.FC<Props> = ({
 }) => {
   const [showModal, setShowModal] = useState(false);
 
+  const { gameService } = useContext(Context);
+  const [gameState] = useActor(gameService);
```
adamhannigan, line 38: For optimisation, are we able to use `useSelector` here?

It may be a too much to cause a re-render whenever any game state changes, instead we can listen and react only to Bumpkin experience change.

adamhannigan (commented):
Awesome initiative!

I can definitely see this being used on every building. Just a small bit of feedback around the xState hooks to increase performance 👏

src/features/island/buildings/components/building/kitchen/Kitchen.tsx
adamhannigan: I will include an example of how we can achieve this 👍 One minute 🤓

adamhannigan:
```
const _level = (state: MachineState) =>
  getBumpkinLevel(state.context.state.bumpkin?.experience ?? 0);

export const Kitchen: React.FC<Props> = ({
  buildingId,
  crafting,
  idle,
  ready,
  name,
  craftingService,
  isBuilt,
  onRemove,
}) => {
  const [showModal, setShowModal] = useState(false);

  const { gameService } = useContext(Context);

  const bumpkinLevel = useSelector(gameService, _level);
  const is_kitchen_enabled = isBuildingEnabled(bumpkinLevel, "Kitchen");
  ```

adamhannigan:
One extra thing, it felt a bit strange that all of a sudden the Kitchen was disabled.

Are we able to disable a basic popup that explains you need to level up your Bumpkin?

brian-man:
> One extra thing, it felt a bit strange that all of a sudden the Kitchen was disabled.
> 
> Are we able to disable a basic popup that explains you need to level up your Bumpkin?

That could be done by passing a `disabledOnClick`, right?

src/features/island/buildings/components/building/kitchen/Kitchen.tsx
brian-man: > For optimisation, are we able to use `useSelector` here?
> 
> It may be a too much to cause a re-render whenever any game state changes, instead we can listen and react only to Bumpkin level changes.

Updated

adamhannigan:
@brian-man I had in mind a UI modal that would pop up with the following text.

"Your Bumpkin needs to be level X to access this building"

This modal could open on the `onClick` handler

brian-man:
> @brian-man I had in mind a UI modal that would pop up with the following text.
> 
> "Your Bumpkin needs to be level X to access this building"
> 
> This modal could open on the `onClick` handler

@adamhannigan Similar to what I used (a few hours ago):
> {name} requires bumpkin level {bumpkinLevelRequired} to use.

adamhannigan:
@brian-man Apologies, was testing a stale PR 😓 

When I run the following code I am still able to open the Kitchen, even though I am level 2.

Any ideas what is happening here?

Screenshot 2023-11-08 at 12 34 21 pm

brian-man:
> When I run the following code I am still able to open the Kitchen, even though I am level 2.
> 
> Any ideas what is happening here?
@adamhannigan
Derp.  I missed updating the onClick attribute/property.

I've added two commits:
1. The first fixes this issue.
2. The second adds a `console.log` which shows that the handleDisabledOnClick is getting called.  But it's not showing the expected UI. Ideas?  I'm guessing I'm missing some syntactic sugar.

adamhannigan:
Thanks for the updates @brian-man 

I found the issue. We were returning a JSX element into a function that wasn't used for rendering.

I've updated the code to show a Modal and show the modal 👍 

Screenshot 2023-11-09 at 11 33 38 am

With these latest updates, I will suggest with the team that we change all of the Basic Land Level Requirements back to Bumpkin Level Requirements.

For instance, when building we will only rely on the Bumpkin level, not the expansion level. This will make the UX much simpler for new users 👏 Well done!

adamhannigan (approved):
Great work!

PR #3128, opened 2023-11-09
Enforce bumpkin level requirement for Water Well

brian-man:
# Description

After installing a bumpkin, and placing one or more water wells, players could freely remove the bumpkin (and replace it with a lower-level bumpkin) while continuing to use the benefit of the well(s).

This change now enforces the bumpkin level requirements for Water Well buildings.

## Type of change


src/features/island/plots/Plot.tsx
```diff
@@ -67,6 +67,7 @@ export const Plot: React.FC<Props> = ({ id }) => {
     plotIndex: id,
     crops: crops,
     buildings: buildings,
+    bumpkin: bumpkin,
```
spencerdezartsmith, line 70: Just a note that when assigning a value to a key with the same name you can just use the key. 
```suggestion
    bumpkin
```

spencerdezartsmith (approved):
Nice one

PR #3129, opened 2023-11-09
Enforce bumpkin level requirements for composters

brian-man:
# Description

After installing a bumpkin, and placing one or more composters, players could freely remove the bumpkin (and replace it with a lower-level bumpkin) while continuing to use the composter(s).

This change now enforces the bumpkin level requirements for composters.

## Type of change


spencerdezartsmith (approved):
Works well

PR #3142, opened 2023-11-13
Change AuctionCountdown text to include the item name

brian-man:


adamhannigan (commented):
Nice addition 👏 

I would suggest we keep the word "Auction" in there though. For a new player, it would be quite confusing to see this Countdown Appear (e.g. "Knowledge Crab").

They may think it is a special event 😅

brian-man:
> I would suggest we keep the word "Auction" in there though. For a new player, it would be quite confusing to see this Countdown Appear (e.g. "Knowledge Crab").

@adamhannigan Done.

adamhannigan:
I just tested mobile mode and we will be cutting it pretty close with the full width with some item names.

Screenshot 2023-11-16 at 2 57 59 pm

I'm not sure what the best UI is, but we may need to stack the word "Auction" on top of or under the label to ensure the width does not cover the settings button

spencerdezartsmith:
This is a tough one as some names can be quite long. What if it just said Auction is live and you could click on that and it would open a model with information about the item?

spencerdezartsmith:
Also, what happens if you visit someone and an auction is live? This is the same position as the visit banner goes. Might be worth checking. 

Also just a note on developing UI, it makes the most sense to work on it only on a mobile screen. Once you have it working well on mobile then you can handle desktop. Its difficult to work in the other direction. We want mobile UI to be the main focus moving forward. 🙏🏼

adamhannigan:
Screenshot 2023-11-20 at 1 34 09 pm

@spencerdezartsmith @brian-man For simplicity, I've opted with @brian-man approach, but added a text overflow and set a max width to ensure it won't overflow on mobile devices.

PR #3149, opened 2023-11-16
Simplify and sort canWithdrawBoostedWearable()

brian-man:
# Description

Code cleanup for canWithdrawBoostedWearable():
- Combine Infernal Pitchfork with GA+Wings case
- Reorder: all crops, crops by base duration, all fruits, specific fruits

The latter might help prevent cases like the former in the future.


adamhannigan (approved):
LGTM! :shipit:

PR #3150, opened 2023-11-16
[FEAT] Discord Gift

adamhannigan:
# Description

Players that join our Discord have much higher engagement and conversion rates. This PR introduces a new NPC that encourages players to join the Discord channel for a free gift (non-tradeable).

The boat drifts nearby the island and pauses. New players should experience this at around the 30-40 seconds mark.

Players will receive a:
- Companion Cap
- 5 Axes

**Existing players can also claim this reward**

Screenshot 2023-11-16 at 4 41 13 pm
Screenshot 2023-11-16 at 4 41 16 pm
Screenshot 2023-11-16 at 4 41 20 pm


spencerdezartsmith (commented):
Nice initiative Adam! I am going to be annoying though and point out some inconsistencies in the UI. Its coming from a place of love <3.

I think we should be consistent with our padding and alignment. I have pointed out some discrepancies. I think it might be cleaner to have the number of the reward item in the label just to have it be cleaner. 

The `Box` component has a lot of additional padding that is there to accommodate the select box. For these types of display boxes you can use the `SimpleBox` component that I made for the codex which doesn't do this. 

I think pushing for these small consistencies is important because when there is one or two it looks like not a big deal but they can easily grow and end up leaving the game looking unpolished/unprofessional. 

Feedback

src/features/game/expansion/components/DiscordBoat.tsx
```diff
@@ -0,0 +1,220 @@
...
+  BONUSES["discord-signup"].isClaimed(state.context.state);
+
+const _hasAccess = (state: MachineState) =>
+  hasFeatureAccess(state.context.state, "DISCORD_BONUS");
+
+export const DiscordBoat: React.FC = () => {
+  const [showModal, setShowModal] = useState(false);
+
+  const { authService } = useContext(Auth.Context);
+  const [authState] = useActor(authService);
```
spencerdezartsmith, line 154: You could also use `useSelector` on the authService for the `usesr`. Not crucial given the auth machine doesn't change too much.

adamhannigan:
Thanks mate @spencerdezartsmith I will address the padding/UI/airdrop separately to this one 👍

PR #3153, opened 2023-11-16
[FIX] Remove/Withdraw requirements for Alba, Ancient Rod, and Walrus

brian-man:


src/features/game/types/removeables.ts
```diff
@@ -275,6 +276,14 @@ export const REMOVAL_RESTRICTIONS: Partial<
   "Heart of Davy Jones": (game) => areAnyTreasureHolesDug(game),
   "Maneki Neko": (game) => hasShakenManeki(game),
   "Time Warp Totem": (_: GameState) => [true, "In use"],
+
+  // Fishing Boosts
+  Alba: (game) => {
```
spencerdezartsmith, line 281: We should always use the sstrict equality ( === ) operator in any kind of conditional logic because the way javascript performs type coercion in cases where you don't expect.

Also, I don't really understand this remove condition? Alba is only in use if you haven't used it for the day?

spencerdezartsmith (changes requested):
Nice work. Just need to change the operands and help me understand one of the conditions and then we will be good to go.

src/features/game/types/removeables.ts
brian-man: @elias7905 Proposed that "fished today" should be the criteria for locking.  I'm open to other ideas, if that doesn't make sense to you.

brian-man: > We should always use the sstrict equality ( === ) operator

Where is this referring to?  Thanks.

brian-man: > Alba is only in use if you haven't used it for the day?

Ah, I see what you're saying now.  Typo: should have been `!==`.

spencerdezartsmith (approved):
All good. I think that condition is ok for now.

PR #3158, opened 2023-11-17
[FEAT] Move chum selection button in Fishing UI

brian-man:
# Description

In discord, user Tumbo reported
> I flubbed a couple of taps and fished without chum.

and recommended
> more separation between the chum “+” and “Cast” buttons

Good idea, IMO.

image

### DIFF
--- a/src/features/island/fisherman/FishermanModal.tsx
+++ b/src/features/island/fisherman/FishermanModal.tsx
@@ -194,16 +194,27 @@ const BaitSelection: React.FC<{
         </div>
       </div>
       <div>
-        <div className="flex flex-wrap">
-          {BAIT.map((name) => (
-            <Box
-              image={ITEM_DETAILS[name].image}
-              isSelected={bait === name}
-              count={state.inventory[name]}
-              onClick={() => setBait(name)}
-              key={name}
-            />
-          ))}
+        <div className="flex items-center justify-between flex-wrap">
+          <div className="flex flex-wrap">
+            {BAIT.map((name) => (
+              <Box
+                image={ITEM_DETAILS[name].image}
+                isSelected={bait === name}
+                count={state.inventory[name]}
+                onClick={() => setBait(name)}
+                key={name}
+              />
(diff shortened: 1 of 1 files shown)


src/features/island/fisherman/FishermanModal.tsx
```diff
@@ -251,15 +262,6 @@ const BaitSelection: React.FC<{
               Attract fish by throwing chum into the water.
             </p>
           </div>
-          <Button
```
spencerdezartsmith, line 254:
```
          <Button
            disabled={fishingLimitReached}
            className="h-11 w-12 min-w-max"
            onClick={() => setShowChum(true)}
          >
            <div className="flex items-center">
              <img src={plus} className="w-6 mt-1" />
            </div>
          </Button>
```

spencerdezartsmith (changes requested):
I can see the issue that the player mentioned, however moving the button away from the help text doesn't make sense. I think it makes more sense to just increase the size of the button in place which forces more space between the chum and cast buttons. 

I have added some code in the comments.

Screenshot 2023-11-20 at 9 54 49 am

adamhannigan:
Where have we settled with this initiative?

I agree with @spencerdezartsmith that the button needs to me mapped (in proximity) closely to the helper text about chum.

Screenshot 2023-11-29 at 11 18 05 am

spencerdezartsmith:
> Where have we settled with this initiative?
> 
> I agree with @spencerdezartsmith that the button needs to me mapped (in proximity) closely to the helper text about chum.
> 
> I do see how fat fingers can accidentally press it - we may need to figure out an alternative UI to support this
> 
> Screenshot 2023-11-29 at 11 18 05 am

I haven't re run this but if the button is just a bit bigger in the position I mentioned, then I am happy to get this in.

adamhannigan:
Closing for inactivity - it appears a consensus was not reached.

I think we need to revisit this with a new lens - perhaps some iconography for the chum action?

PR #3160, opened 2023-11-17
[FEAT] Disable 'Seed not selected' with selected Shovel

brian-man:
# Description

When harvesting only (meaning no planting wanted) of crops, the new "Seed not selected" prompt is pure annoyance.

While the feature has value for new players, it's unfortunate that it breaks the workflow for experienced players.

This change does the following:
1. When in inventory, add Shovel to the Tools section of the Basket.
2. When Shovel is the selected item (as indicated via the first entry in the shortcuts MRU list), clicking an empty crop plot is a non-action as it was previously.

## Type of change

### DIFF
--- a/src/features/game/types/tools.ts
+++ b/src/features/game/types/tools.ts
@@ -5,6 +5,8 @@
 import Decimal from "decimal.js-light";
 import { GameState, Inventory } from "./game";
 
+export type UnlimitedUseToolName = "Shovel";
+
 export type WorkbenchToolName =
   | "Axe"
   | "Pickaxe"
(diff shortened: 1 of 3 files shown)


adamhannigan:
Thanks for the PR @brian-man 

The initial use case that this aims to explore is that when players have the Shovel, they keep trying to plant on the plots. Unfortunately if the modal never showed while Shovel was selected, new players may get stuck.

The selection of the shovel may become a bit tedious for those who would like to do their seed selection this way as well. 

I am going to play around with a few more mechanics to make it more seamless and will keep you updated. Something like a double click would trigger the modal.

adamhannigan:
I have temporarily disabled this logic until we can figure out a more sustainable path.

brian-man:
Closing for now.

PR #3161, opened 2023-11-17
[FIX] Kraken Head description refers to tentacle

brian-man:


c-py (approved):
Awesome, thanks @brian-man !

PR #3162, opened 2023-11-17
[FEAT] Exclude harvested item from StopTheGoblins challenge

brian-man:
# Description

Exclude the harvested item (crop or wood) from the possible images to avoid cognitive dissonance for players.

# Testing

On local, forced `StopTheGoblins` challenge for both moonseeker and non-moonseeker, confirming that the moonseeker flavor excludes Sunflower when Sunflower is being harvested.

### DIFF
--- a/src/features/game/expansion/components/resources/tree/Tree.tsx
+++ b/src/features/game/expansion/components/resources/tree/Tree.tsx
@@ -180,6 +180,7 @@ export const Tree: React.FC<Props> = ({ id }) => {
 
       {/* Chest reward */}
       <ChestReward
+        collectedItem={"Wood"}
         reward={reward}
         onCollected={onCollectChest}
         onOpen={() =>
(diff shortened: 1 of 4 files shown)


src/features/game/expansion/components/resources/tree/Tree.tsx
```diff
@@ -180,6 +180,7 @@ export const Tree: React.FC<Props> = ({ id }) => {
 
       {/* Chest reward */}
       <ChestReward
+        collectedItem={"Wood"}
```
spencerdezartsmith, line 183: Just FYI, No need to wrap to wrap strings in `{}`.
```suggestion
        collectedItem="Wood"
```

spencerdezartsmith (approved):
I am fine with this change but I would love to understand why you would describe this issue as one of cognitive dissonance? Just generally curious. I would imagine its more so a case of autopilot confusion (not an official term 😄 ) 🤷🏼‍♂️

PR #3169, opened 2023-11-19
[FEAT] Add animations enable/disable to SubSettings

brian-man:
# Description

Adds setting to disable/enable animations to the secondary settings menu.

Initial implementation only impacts the firework animation.

# Testing

On local, set all 3 initial Sunflower plantings to 11 yield and confirmed that the enabled and disabled animation values worked as expected.

image


c-py (approved):
A good start towards allowing animations to be disabled, thanks @brian-man !

PR #3173, opened 2023-11-19
[FIX] Update compost limits in front-end

brian-man:
# Description

The back-end limits for compost (in the game contract) are set to 500.

The front-end was not enforcing (any value), causing players to run into unusual "can't collect from composter" issues.

## Type of change


spencerdezartsmith (approved):
LGTM

PR #3178, opened 2023-11-20
[FIX] Stable (and faster) sorting of buildings

brian-man:


brian-man:
While the (likely very slight) performance improvement is potentially worthwhile, the unstable sorting issue I wasn't able to repro (yet) to verify the fix.  Hence still in Draft.

PR #3206, opened 2023-11-28
[FEAT] Pete Helpers

adamhannigan:
# Description

This PR adds helper speech from Pumpkin Pete to help onboard new players. The aim is to improve players going through the first 5-6 steps of the game.

Screenshot 2023-11-28 at 11 01 19 am

**How to test?**

Run the game locally and go through the initial steps of the game.


spencerdezartsmith (approved):
Worked well for the most part. I think the part where I could get lost is the step between selling and buying. I think there needs to be something more explicit to go back to the buy tab. Maybe the floating pointer can pop up here:
Screenshot 2023-11-28 at 11 11 26 am

Also just a couple other things:

If you click a different plot you kinda get pointers all over. Not sure if its an issue or not.
Screenshot 2023-11-28 at 11 12 32 am

When I got asked to craft a scarecrow I didn't get a pointer to the workbench. I think it might be useful. 
Screenshot 2023-11-28 at 11 13 07 am

Also, I should probably have a pointer to the fire pit when I am directed to cook food. I think we should be consistent with using the pointer for each step. I also didn't find any real value with the coin icon. It just felt cluttered. I think the Pete hint and pointer are enough.

All non blocking comments.

PR #3210, opened 2023-11-29
[FEAT] Disable Animations impacts Betty in the Market

brian-man:
# Description

Enable/Disable Animations setting now toggles the animation of Betty in the Market.

## Type of change

# How Has This Been Tested?

On local:
1. move the home island so that the Market is on the right side of the browser window
2. open `...`/`Settings`
3. toggle Enable/Disable Animations, observing Betty

image

### DIFF
--- a/src/features/island/buildings/components/building/market/Market.tsx
+++ b/src/features/island/buildings/components/building/market/Market.tsx
@@ -37,10 +37,27 @@ const hasBoughtCropsBefore = (bumpkin?: Bumpkin) => {
   );
 };
 
+export const paintCanvas = (imgElement: HTMLImageElement) => {
+  const canvasElement = imgElement.previousElementSibling as HTMLCanvasElement;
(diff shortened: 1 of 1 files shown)


adamhannigan:
Interesting proposal @brian-man 

A few things I would like to dive deeper into:

- Can we benchmark performance on this approach?
- Can we simplify the usage? It isn't the clearest for someone reading our code base what is happening here. There may be some optimised pre-built React GIF libraries we could use here.

If we can satisfy an increase in overall performance + simple usage I am happy to go forward with this approach with GIFs.

Something worth mentioning as well is we are planning on removing the NPCs from the shops on your main island (they will be returning to the plaza to live).

adamhannigan:
@brian-man As discussed in past Discord discussion, we will be opting from removing the NPC bumpkins from the main land. This helps the lore of being on a 'deserted' island + helps with performance 👍

PR #3221, opened 2023-12-05
[FIX] Sequence Icon not available in some regions

c-py:
# Description

Sequence Logo is not available in some regions. This PR base64 encodes it so it can be served in line.


adamhannigan (approved):
LGTM

PR #3224, opened 2023-12-08
[FEAT] Link wallet flow

adamhannigan:
# Description

This flow allows Google SSO players to link their wallet. This will enable them to access Web3 features such as Auctions, Withdrawing + much more.


elias7905:
Hi @adamhannigan would like to enquire, will existing players be able to link their wallet to Google SSO so that they can play from Google rather than logging in from their wallet?

adamhannigan:
@elias7905 I would love this. The team have spoken about it but won't be something we release in the initial iteration.

elias7905:
> @elias7905 I would love this. The team has spoken about it but won't be something we release in the initial iteration.
> 
> 

@adamhannigan I can see a situation where an existing player would attempt to log in with their Google account and accidentally create a new account. Anything to prevent this from happening on launch?

adamhannigan:
@elias7905 Yes that is a valid use case.

If players click 'Google', they will be presented with a new screen - "Looks like you don't have an account, create an account...."

We anticipate this will deter people accidentally creating multiple accounts. The same is if you accidentally switch to a different MetaMask account, you will now be prompted with this new UI screen.

Given we are close to the holidays we likely won't release this until after christmas. I will try prepare a video and share in mods chat to view these use cases

adamhannigan:
@elias7905 https://vimeo.com/894334512?share=copy

elias7905:
> @elias7905 https://vimeo.com/894334512?share=copy

@adamhannigan video's really clear and addresses most of my queries regarding this feature. Thank you for your work!

PR #3229, opened 2023-12-10
[FIX] Negative Tree Times

adamhannigan:
# Description

This PR addresses an incorrect boost for tree times. It was not following the same pattern for every other speed boost in the game (multiplicative) and could lead to negative times.

This could be exploited for instantly ready trees for someone with Foreman + Time Warp.


c-py (approved):
LGTM!

PR #3232, opened 2023-12-11
[FEAT] Snow respects Disable Animations setting

brian-man:


adamhannigan (approved):
LGTM 👍 

Strangely the snow is a pure CSS animation and respects the accelerated/optimised animation properties. 

I believe they may be experiencing lag from something else on the Christmas map.

PR #3235, opened 2023-12-12
[FIX] Level Up unlocks

brian-man:
# Description

With Node Rebalancing, the meaning of `unlocksAtLevel` was changed but the Level Up UI wasn't updated to match.

## Type of change


adamhannigan (approved):
LGTM

PR #3246, opened 2023-12-19
[FEAT] Add bait to garbage collector

brian-man:
# Description

Some players only want to use their composters for fertilizer.  For such players, the hoarder limits for bait is cumbersome (and expensive).

This change adds all three composter bait types to the garbage collector.

## Type of change


adamhannigan (approved):
Looks good to me 👏 🚀 Backend updated ✅

PR #3252, opened 2023-12-20
[FIX] Christmas candy finish

adamhannigan:
# Description

After 12 days, players should not be able to pick up candy


spencerdezartsmith (approved):
LGTM

PR #3254, opened 2023-12-21
[FIX] Reveal Land

adamhannigan:
# Description

When a new player expands, our local `land.revealed` event adds the natural resources and creates random IDs for them. On Autosave, our backend then creates new IDs for these resources.

If a player quickly interacts with a resource after expanding, an error will be thrown.

**Short Term Fix**

This is not a permanent fix, but will help alleviate most of the problems:

1. Save as soon as they expand
2. Show a pop congratulations modal - by the time they close it the game should be updated.


spencerdezartsmith (approved):
LGTM

PR #3255, opened 2023-12-21
[FIX] canWithdrawBoostedWearable() is too restrictive

brian-man:
# Description

Wearable withdraw restrictions are "older" and thus too restrictive.  More specifically, they needed to be updated to allow for removal when affected nodes (plots, chickens, etc.) are ready for harvesting/collection.

## Type of change


adamhannigan (approved):
Nice addition 👍 Will definitely help with confusion

PR #3257, opened 2023-12-23
[FIX] Elf Suit withdrawable

brian-man:


elias7905:
I was thinking, should we change the date to 31st instead? Cuz that's the date stated in the docs

adamhannigan (approved):
LGTM

PR #3259, opened 2023-12-23
[FEAT] Sunflower Express setting

brian-man:
# Description

More experienced players want to interact with NPCs (especially for chores and deliveries) without having to spend precious time or risk RSI (repetitive strain injury).

This change adds a new "Sunflower Express" setting to expand the interactive range for NPCs.

## Type of change

# How Has This Been Tested?

With Testnet farm in the ChristmasScene.

Setting UI
image

Disabled
image

Enabled
image

### DIFF
--- a/src/features/farming/hud/lib/express.ts
+++ b/src/features/farming/hud/lib/express.ts
@@ -1,0 +1,14 @@
+/**
+ * Cache show timers setting in local storage so we can remember next time we open the HUD.
+ */
+const LOCAL_STORAGE_KEY = "settings.express";
+
+export function cacheExpressSetting(show: boolean) {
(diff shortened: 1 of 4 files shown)


adamhannigan (commented):
This is an interesting idea @brian-man 👏 

It seems like some players are annoyed having to visit the plaza, walking around a lot complete their deliveries. It is time consuming and requires a lot of WASD movement & clicks.

For this sort of friction, we will need to dive deeper into a more sustainable approach to making the feature more accessible and address the root problem these players face around deliveries.

This will likely lean to some option to completing deliveries from your main island. On our end, we are still encouraging players to visit the plaza.

Perhaps going to the plaza will be optional, but players must pay a 'delivery' fee otherwise.

I will close this PR for now but leave the discussion open 👍

PR #3273, opened 2024-01-06
[FIX] Respect mute for walkAudioController in scenes

brian-man:


c-py (approved):
LGTM! Nice addition @brian-man !

I did notice if I was already in the plaza the steps sound continue to play after mute. But I think this can be merged even without the audio toggle functionality

PR #3277, opened 2024-01-08
[FIX] Add Banana Time Reduction Boosts to front-end

elias7905:
# Description

The Banana time reduction boosts for Nana and Banana Onesie are currently working in the backend but not reflected in the front-end

`fruitPlanted.ts` was updated on the front end for the boosts when the banana plant is planted but not `fruitHarvested.ts` for when the banana is harvested and goes into cooldown

Fixes #issue
Banana timer after harvest doesn't show time reduction until save

## Type of change

# How Has This Been Tested?

- Tested on local Host with Nana, Banana Onesie and Squirrel Monkey as well as an ready to harvest banana tree (and orange tree) since the issue was happening to the timers on harvest. Not when first planted. 
- Tested with Nana and Banana Onesie separately then together to make sure their boost number works separately
- Tested Nana and Banana Onesie together to make sure that the multiplicative boost works
- Tested Squirrel Monkey since there has been a change in how the code is presented

# Checklist:


c-py (approved):
@elias7905 I have updated the PR to reuse the code from `fruitPlanted` to try and remove duplicated code between the two events - let me know if I missed something!

Thank you for raising the fix!

PR #3279, opened 2024-01-09
[FIX] Trade Notifications

adamhannigan:
# Description

Since a wallet is no longer connected, goblin trade notifications cannot be shown on the main island.

Instead, when accessing Goblin Retreat, display any notifications and let the player refresh.

Screenshot 2024-01-10 at 8 49 40 am

## How to test?

Purchase a trade from a farm. Access the farm and see the notification


c-py (approved):
LGTM!

PR #3281, opened 2024-01-09
[FIX] Plaza Disconnects

c-py:
This PR attempts to fix random plaza disconnects that are currently occurring every few days.


adamhannigan (approved):
LGTM

PR #3313, opened 2024-01-27
[FIX] Bert's Obsession for mislabelled wearables

brian-man:
# Description

Bert's Obsession crashes players this week because a wearable is mislabelled as a collectible.
```
"bertObsession": {
    "type": "collectible",
    "name": "Lion Dance Mask",
    ...
}
```

This change makes the front-end more resilient to this kind of back-end issue.

Long-term all lookups of properties for collectibles and wearables should use a common function that double-checks to ensure the type is correct before accessing data structures.

## Type of change


Rofrtd (approved):
Thanks for the fix

PR #3314, opened 2024-01-27
[FIX] Carrot Amulet equipped on farmhand

brian-man:
# Description

When equipping Carrot Amulet on a farmhand, the front-end is now reflecting the boost while the back-end does apply the boost.

## Type of change


adamhannigan (approved):
Good spot 👍 

I've confirmed the backend has the correct logic

PR #3323, opened 2024-01-29
[FEAT] Add mutant chickens as chum to front-end

brian-man:
# Description

Front-end implementation for Blastaman's proposal.

## Type of change

### DIFF
--- a/src/features/game/lib/landDataStatic.ts
+++ b/src/features/game/lib/landDataStatic.ts
@@ -364,6 +364,9 @@ export const STATIC_OFFLINE_FARM: GameState = {
     },
   },
   inventory: {
+    "Rich Chicken": new Decimal(1),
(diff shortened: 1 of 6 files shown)


adamhannigan (dismissed):
Backend is finally live!

Great initiative 👌 We've decided it would be a guaranteed catch, since RNG felt a bit unfair with these items. As such we have bumped the requirements a bit

eliasSFL:
> Backend is finally live!
> 
> Great initiative 👌 We've decided it would be a guaranteed catch, since RNG felt a bit unfair with these items. As such we have bumped the requirements a bit

@adamhannigan is it guaranteed catch for all baits or only for certain baits?

adamhannigan:
@eliasSFL These will only work with Red Wigglers.

I have included some code to ensure that they can only be selected with Red Wigglers

PR #3329, opened 2024-01-29
[FEAT] Remove level param from expansionRequirements()

brian-man:
# Description

The `level` parameter for the `expansionRequirements` function is redundant and the only non-test caller was just using the other parameter to calculate the value to pass.

This change moves the calculation inside the function, removes the redundant parameter, and updates the unit tests accordingly.

## Type of change


brian-man:
#3330 makes this PR obsolete

PR #3331, opened 2024-01-30
[FIX] Update BUILDINGS_DIMENSIONS for House

brian-man:
# Description

The house is rendered 4x4 but the dimensions are 4x3 which causes surprising behavior with mushrooms and item placement.

This change updates the dimensions to 4x4.  Alternatively, the art could be scaled to 4x3 if that's preferred.

## Type of change


adamhannigan:
Was completed on our list 😅 Thanks for the contribution though!

https://github.com/sunflower-land/sunflower-land/commit/ee572df79ee0124c320425cb740059e08af9567f

PR #3340, opened 2024-01-30
[FIX] Corn position in FLOWER_CROSS_BREED_*

brian-man:


elias7905:
image
Maybe you can fix the order in the codex as well? Currently Eggplant and banana are at the top of the list for some reason

brian-man:
> Maybe you can fix the order in the codex as well? Currently Eggplant and banana are at the top of the list for some reason

I think the order listed is the order you discovered, as reported in your farm data.

c-py (approved):
LGTM!

PR #3360, opened 2024-02-01
[FIX] Farmhand-equipped Chef Apron

brian-man:
# Description

The boost for Chef Apron was only being applied for the "primary bumpkin", not for farmhands.

## Type of change


adamhannigan (approved):
LGTM. Added to backend 👍

PR #3381, opened 2024-02-06
[FEAT] Remove old withdraw timestamps

brian-man:


brian-man:
Elias covered this in his PR.

PR #3389, opened 2024-02-08
[FEAT] Disable Animations for animate-float

brian-man:
# Description

Updates all current uses of `animate-float` to respect Disable Animations setting.

## Type of change


c-py (approved):
LGTM!

PR #3400, opened 2024-02-10
[FEAT] Disable Animations for chickens

brian-man:
# Description

Updates placed chickens to respect Disable Animations setting.

## Type of change

image

### DIFF
--- a/src/features/island/chickens/Chicken.tsx
+++ b/src/features/island/chickens/Chicken.tsx
@@ -131,7 +131,8 @@ interface Props {
 
 const PlaceableChicken: React.FC<Props> = ({ id }) => {
   const { scale } = useContext(ZoomContext);
-  const { gameService, shortcutItem, showTimers } = useContext(Context);
+  const { gameService, shortcutItem, showAnimations, showTimers } =
+    useContext(Context);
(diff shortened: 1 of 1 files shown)


spencerdezartsmith (changes requested):
Hey @brian-man nice one. I have tested this but have found a few issues.

The sizing of the static chicken is incorrect and also wrongly positioned as you can sede in the image here: 

Screenshot 2024-02-21 at 10 55 41 am
Screenshot 2024-02-21 at 10 56 05 am

Also I think in general this one needs more thought.

Chickens have a couple of states that are relevant even without animations: 

- hungry
- fed
- ready

We indicate these states currently using little icons above the chicken. I think these are still needed even when animations are off because otherwise it is not very clear what the chicken is doing especially if you have the timers off. 

Screenshot 2024-02-21 at 11 18 19 am

In the above image all the chickens look exactly the same if the timers are off but they all have different states. A player should be able to understand the state of the chicken just by looking at it rather then needing to put the timers on to see the state of the chicken.

spencerdezartsmith:
I am going to close this PR as it hasn't had any movement for a while.

PR #3406, opened 2024-02-12
[FIX] Dimension heights

adamhannigan:
# Description

This PR changes the height of some items which unnecessarily had extra height (bears, scarecrows, trophy cabinets).

This gives players more space for items and aesthetically helps with designing the land and home.

One caveat is that mushrooms may be partially hidden behind items - it is up to users to place & organise their items


c-py (approved):
LGTM!

PR #3440, opened 2024-02-23
[FIX] Remove Large White Background

c-py:


spencerdezartsmith (approved):
Some of our best work 🚀

PR #3447, opened 2024-02-24
[CHORE] Update Sequence to newest versions

c-py:
# Description

This PR updates sequence to use their latest SDK version which includes a new access key functionality


adamhannigan (dismissed):
LGTM

PR #3451, opened 2024-02-25
[FIX] Hoarder limit for Tulip Bulb

brian-man:
# Description

Adds front-end limit for Tulip Bulb that matches what the contract is enforcing.

## Type of change

### DIFF
--- a/src/features/game/lib/processEvent.ts
+++ b/src/features/game/lib/processEvent.ts
@@ -148,7 +148,7 @@ export const maxItems: Inventory = {
   "Dawn Breaker Ticket": new Decimal(750),
   "Crow Feather": new Decimal(750),
   "Mermaid Scale": new Decimal(1500),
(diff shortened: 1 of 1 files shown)


src/features/game/lib/processEvent.ts
```diff
@@ -148,6 +148,7 @@ export const maxItems: Inventory = {
   "Dawn Breaker Ticket": new Decimal(750),
   "Crow Feather": new Decimal(750),
   "Mermaid Scale": new Decimal(1500),
+  "Tulip Bulb": new Decimal(1750),
```
c-py, line 151:
```suggestion
  "Tulip Bulb": new Decimal(1500),
```

c-py (approved):
LGTM! Thanks @brian-man ! I've changed it to 1500. We keep a small tolerance to ensure the number doesn't overflow the on chain max.

c-py (approved):

PR #3474, opened 2024-03-03
[FEAT] Mint account NFT

adamhannigan:
# Description

Now NFTs are not needed for signup, we will require players to manually mint their own (pay for gas).

This will prevent the Farm Minter running out of gas error.

**Note** Legacy code will be removed once tested on mainnet

## How to test?

Create a new account. Go to Deposit > Mint Farm NFT > Submit Transaction ✅

### DIFF
--- a/src/features/wallet/actions/mintFarm.ts
+++ b/src/features/wallet/actions/mintFarm.ts
@@ -46,7 +46,7 @@ export async function mintNFTFarm(request: Request) {
   });
 
   if (response.status >= 400) {
-    throw new Error(ERRORS.LOGIN_SERVER_ERROR);
+    throw new Error(ERRORS.MINT_SERVER_ERROR);
   }
 
   const payload: {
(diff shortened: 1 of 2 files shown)


src/features/wallet/actions/mintFarm.ts
```diff
@@ -31,3 +34,43 @@ export async function mintFarm(request: Request) {
...
+    headers: {
+      "content-type": "application/json;charset=UTF-8",
+      "X-Transaction-ID": request.transactionId,
+      Authorization: `Bearer ${request.jwt}`,
+    },
+  });
+
+  if (response.status >= 400) {
+    throw new Error(ERRORS.LOGIN_SERVER_ERROR);
+  }
```
c-py, line 50:
```suggestion
  if (response.status >= 400) {
    throw new Error(ERRORS.MINT_SERVER_ERROR);
  }
```

The team can get discombobbed when these are wrong

c-py (approved):
LGTM!, Tested both minting flow, and also refreshing half way through. Both worked well!

c-py (approved):
LGTM!, Tested both minting flow, and also refreshing half way through. Both worked well!

I pushed a small change to the error messaging.

PR #3476, opened 2024-03-04
[FIX] Chicken chum counts

brian-man:
# Description

When using mutant chickens as chum, the fishing UI wasn't excluding placed mutant chickens from the counts and requirements.

## Type of change


adamhannigan (approved):
Good spot!

PR #3482, opened 2024-03-05
[FIX] Add Honey hoarder limit

brian-man:
# Description

The back-end is enforcing a limit of 100 for honey.

This updates the front-end to match.

Observed AS-001 on Testnet 162 when trying to Save after adding to 99.62 honey in the session.

## Type of change

### DIFF
--- a/src/features/game/lib/processEvent.ts
+++ b/src/features/game/lib/processEvent.ts
@@ -104,6 +104,7 @@ export const maxItems: Inventory = {
   Stone: new Decimal("800"),
   Wood: new Decimal("4000"),
   "Wild Mushroom": new Decimal("100"),
+  Honey: new Decimal("100"),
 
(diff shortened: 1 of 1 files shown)


src/features/game/lib/processEvent.ts
```diff
@@ -104,6 +104,7 @@ export const maxItems: Inventory = {
   Stone: new Decimal("800"),
   Wood: new Decimal("4000"),
   "Wild Mushroom": new Decimal("100"),
+  Honey: new Decimal("100"),
```
c-py, line 107:
```suggestion
  Honey: new Decimal("80"),
```

c-py (dismissed):
LGTM! @brian-man  I've updated it 80 - we keep a small tolerance in case the number overflows the backend max.

PR #3490, opened 2024-03-07
[FEAT] Island tax logic

adamhannigan:
# Description

Reduce taxes as players unlock new islands. Spring island = `2.5%` tax reduction


brian-man:
Any thoughts on resource tax reduction?

adamhannigan:
@brian-man Those Goblins are greedy!

To be honest, it is unlikely resources would get reduced. We've got some big resource trading announcements coming you will see in Beta chat - we doubt many will be trading outside after this 😄 🙏

PR #3528, opened 2024-03-16
[FEAT] Bud box

adamhannigan:
# Description

Wen bud box.

Each day of the week, a player can open the bud box if they have a certain bud.

The Bud Box gives cosmetics + items to help with farming & gameplay.

Screenshot 2024-03-16 at 2 47 16 pm


eliasSFL:
Hi @adamhannigan, if you have both buds that's eligible to open the bud box in a day, will u be able to open twice that day?

adamhannigan:
The code currently just supports one open per day (regardless the amount + types you have)

PR #3551, opened 2024-03-28
[FEAT] Real time mail

adamhannigan:
# Description

On each autosave, check for new mail. This helps with onboarding messages and updates.


JCtapuk:
@adamhannigan Why don’t you implement it via Server Send Event? In it you can get real time from any delivery to the post office =)

adamhannigan:
@JCtapuk We have considered websockets + sync architecture - however, most of our infrastructure is dependent on our current polling/REST approach.

Unless critical, we are opting to keep our architecture simple. It is a tradeoff for speed + consistency

PR #3569, opened 2024-04-03
[FIX] NPC badges clickable from anywhere

0xSacul:
# Description

For a long time now, clicking on NPC badges from anywhere would open their modal, this was known for a long time but not intended.

I've added a `checkDistanceToSprite(sprite, maxDistance)` function in BaseScene that can be called in any scene in the `on("pointerdown")` function. 

*not a bug, a feature*

## Type of change

# How Has This Been Tested?

local & yarn test
image

# Checklist:


spencerdezartsmith (changes requested):
Nice job. I think this needs a bit more work so that it is consistant. Currently, the badges for the treasure chests are never clickable, its only the chest. 

However, the badges on Stella and Haz are. We should make them all clickable or all not. I think not makes sense and just leave the actual "thing" be clickable.

For Hammerin Haz, the badge isn't clickable from afar but the chest is. Probably should be put on the chest rather than the badge.

src/features/world/scenes/BeachScene.ts
```diff
@@ -191,10 +190,8 @@ export class BeachScene extends BaseScene {
       this.add.sprite(320, 580, "locked_disc").setDepth(1000000000);
     }
 
-    const chest = this.add.sprite(320, 600, "wooden_chest");
-    chest.setInteractive({ cursor: "pointer" }).on("pointerdown", () => {
```
spencerdezartsmith, line 195: I don't think we want to remove this interaction. I can't click on the wooden chest in the beach any more.

src/features/world/scenes/RetreatScene.ts
```diff
@@ -151,21 +143,13 @@ export class RetreatScene extends BaseScene {
     });
     greedclaw.play("greedclaw_animation", true);
 
-    const bank = this.add.sprite(422, 94, "bank");
-    bank.setInteractive({ cursor: "pointer" }).on("pointerdown", () => {
```
spencerdezartsmith, line 155: We need this as well otherwise we can't open the bank.

src/features/world/scenes/RetreatScene.ts
```diff
@@ -151,21 +143,13 @@ export class RetreatScene extends BaseScene {
...
-      interactableModalManager.open("bank");
-    });
-
+    this.add.sprite(422, 94, "bank");
     this.add.sprite(532, 51, "raffle_disc").setDepth(1000000000);
-
     this.add.sprite(147, 200, "exchange_disc").setDepth(1000000000);
-
-    const wishingWell = this.add.sprite(532, 71, "wishing_well");
-    wishingWell.setInteractive({ cursor: "pointer" }).on("pointerdown", () => {
```
spencerdezartsmith, line 164: We need this also.

src/features/world/scenes/RetreatScene.ts
```diff
@@ -119,12 +119,6 @@ export class RetreatScene extends BaseScene {
     });
     bigGoblin.play("big_goblin_animation", true);
 
-    const exchange = this.add.sprite(114, 215, "exchange");
-    // On click
-    exchange.setInteractive({ cursor: "pointer" }).on("pointerdown", () => {
-      interactableModalManager.open("goblin_market");
-    });
-
```
spencerdezartsmith, line 127: We need all of this also. These interactables should just be on the building/npc so that we can interact with it.

spencerdezartsmith (changes requested):
Some of the required interactables have been removed. We will need these back. The badges are all not clickable which is fine. 

I think that Hammerin' Haz should also say "you're too far away" or whatever it is the other npc's say if you try to click on them and they're too far away. 

Also the goblin at the exchange has a box through his head. Not sure if the ordering of these sprites has somehow messed with his index.

spencerdezartsmith (commented):
Tested and working well for me. 

One thing I think would be beneficial is adding the "I am too far away" on all the non npc clickables. Just because the cursor pointer is there when you hover over it but nothing happens if you're far away so the player might think its not working due to no feedback. Examples would be wishing well, chests, books. 

What do you think? Would you be keen to do that?

0xSacul:
> Tested and working well for me.
> 
> One thing I think would be beneficial is adding the "I am too far away" on all the non npc clickables. Just because the cursor pointer is there when you hover over it but nothing happens if you're far away so the player might think its not working due to no feedback. Examples would be wishing well, chests, books.
> 
> What do you think? Would you be keen to do that?

Indeed that's a good idea! On my way to hunt every non npc clickables to add the text 👀 

*Will do that in a day or two, currently busy on other PRs*

PR #3576, opened 2024-04-04
[FIX] Potion House button disabled b/c SFL

brian-man:


brian-man:
@c-py submitted a newer fix ><

c-py:
> @c-py submitted a newer fix ><

Whoops! Thanks @brian-man !

PR #3643, opened 2024-04-30
[FEAT] VIP Access UI updates

adamhannigan:
# Description

Small, non blocking UI updates for VIP UI components.

Screenshot 2024-04-30 at 11 01 30 AM
Screenshot 2024-05-01 at 10 53 55 AM

- “VIP Access Required” label does not look clickable - Put a plus button on the end of it.
- VIP Items - Banner - Show time left before next deal - “X hours left”
- Was unsure what the “Support Gift” was with the Lifetime banner. Since we don’t mention “Support Gift” with the normal banner I would suggest we just omit it so players don’t think they get something extra. I was looking for a collectible and/or wearable after purchase
- "VIP Access" → "Season VIP Access" = a bit clearer it is not permanent VIP access

### DIFF
--- a/src/features/game/components/VipAccess.tsx
+++ b/src/features/game/components/VipAccess.tsx
@@ -39,7 +39,7 @@ export const VIPAccess: React.FC<VIPAccessProps> = ({ onUpgrade, isVIP }) => {
     <Label
       type="warning"
       icon={vipIcon}
-      secondaryIcon={lock}
+      secondaryIcon={SUNNYSIDE.ui.add_button}
       onClick={onUpgrade}
       className="whitespace-nowrap"
     >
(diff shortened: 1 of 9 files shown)


src/features/game/components/VipAccess.tsx
```diff
@@ -39,7 +39,7 @@ export const VIPAccess: React.FC<VIPAccessProps> = ({ onUpgrade, isVIP }) => {
     <Label
       type="warning"
       icon={vipIcon}
-      secondaryIcon={lock}
+      secondaryIcon={SUNNYSIDE.ui.add_button}
```
c-py, line 42: Looks great!

c-py (approved):
LGTM!

PR #3661, opened 2024-05-03
[FIX] Extra Ticket Issue

adamhannigan:
This PR fixes the issue where players who just purchased a banner, would not get the bonus tickets from deliveries and chores.

The issue was we were storing the number of tickets, at the point of time the delivery or chore was created (which in some cases was before they attained a banner).

This PR changes it to no longer store the `ticket` in the DB. Instead, these are dynamically calculated at the point of `order.delivered` or `chore.completed`

**Chore ticket update**

The chore difficulty now determines how many tickets the task will yield:

```
export const CHORE_TICKETS: Record<ChoreV2Name, number> = {
  "1": 1,
  "2": 2,
  "3": 3,
  "4": 4,
  "5": 5,
};
```

This will change some of the task's tickets that were previously hardcoded.

**Extras**

1. Fixed up flaky chore + delivery tests that were breaking due to season changeover
2. Old delivery and chore code was removed


SchopenhauerBoi:
!!

spencerdezartsmith (approved):
🚀  Tested an working well

PR #3671, opened 2024-05-07
[FIX] Lock icon on unlocked tools

brian-man:


spencerdezartsmith:
All fixed mate. Thanks for jumping on it.

PR #3683, opened 2024-05-10
[FEAT] Manor

adamhannigan:
# Description

This PR includes the Manor Building (4x4)

- Exterior Design (5x4)
- Interior Design (16x16)

Screenshot 2024-05-10 at 12 58 37 PM
Screenshot 2024-05-10 at 1 26 38 PM


eliasSFL:
Love it!

Dionis404:
Maybe add walls inside the house? To create coziness and beauty?

adamhannigan:
@Dionis404 This was my exact thought 👌 

One of the challenges we face however is that all of the collectibles previously in the Home (Spring Island), automatically get moved into the Manor.

It would mean that some walls would overlap with the items that are already placed in there.

Do you have any ideas how we could approach this?

Dionis404:
@adamhannigan  The only thing I could think of is to make a binding to the entrance, and when enlarging, new cells will remain empty.  
image
Well, or when you zoom in, just dump the walls into your inventory

c-py (approved):
LGTM!

adamhannigan:
@Dionis404 This was my initial thought as well. However, I discovered that we didn't get much edge space on each upgrade.

For example, when we go from 12x12 to 16x16 we only get an extra 2 squares on each side which doesn't give us much room to design a room.

Instead of growing from the center, we may need to make the anchor point the bottom left. This would give us more space to create some rooms.

I am interested in doing this - it will just mean a breaking UI change for players when we make the change. Some walls will overlap objects until they fix it 😅

PR #3686, opened 2024-05-11
[FIX] Delivery NPCs locked on desert

brian-man:
# Description

Delivery NPCs should be unlocked for desert (prestige 2) regardless of how many expansions.

## Type of change


c-py:
LGTM!

PR #3696, opened 2024-05-14
[FIX] Unwanted confetti()

brian-man:


c-py (approved):
Thanks for updating this @brian-man !

PR #3704, opened 2024-05-14
[FEAT] Desert Resources Art

c-py:
# Description

Makes the resources look desert-ish

look

## Type of change

# How Has This Been Tested?

localhost

# Checklist:


adamhannigan (dismissed):
Looking good mate! Didn't test all the flowers but the one I tried looked great 👌 

A few non blockers:

- The bright yellow spots on the dirt path look a bit off. I'd suggest using the darker tones
- When buying a new path, it still shows the old image. I'm not too worried about this one since it kind of makes sense. You have a normal path that turns to desert only once placed 😅 

Screenshot 2024-05-14 at 5 20 36 PM
Screenshot 2024-05-14 at 5 20 46 PM

c-py:
> Looking good mate! Didn't test all the flowers but the one I tried looked great 👌
> 
> A few non blockers:
> 
>     * The bright yellow spots on the dirt path look a bit off. I'd suggest using the darker tones
> 
>     * When buying a new path, it still shows the old image. I'm not too worried about this one since it kind of makes sense. You have a normal path that turns to desert only once placed 😅

Thanks mate, I didn't realise those were the dirt paths 😬 Ill see what I can do, those spots are not even visible under the crops

PR #3724, opened 2024-05-20
[FEAT] Apply hoot boost to greenhouse

adamhannigan:
# Description

Apply Hoot's boost to the overnight crops (>24 hours and <36 hours)


brian-man:
@adamhannigan 
Olive is 44 hours, so "the overnight crop" (singular).

PR #3725, opened 2024-05-20
[FIX] Bud boosts for greenhouse crops

brian-man:
# Description

3 Leaf Clover and Saphiro should be applied to GH crops

## Type of change

### DIFF
--- a/src/features/game/events/landExpansion/plantGreenhouse.ts
+++ b/src/features/game/events/landExpansion/plantGreenhouse.ts
@@ -62,7 +62,7 @@ export const OIL_USAGE: Record<GreenhouseSeed, number> = {
 
 export const MAX_POTS = 4;
 
-function isCrop(
+function isGreenhouseCrop(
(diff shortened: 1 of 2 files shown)


brian-man:
I think Saphiro is already being applied on the back-end but 3 Leaf Clover is not.  So back-end needs an update too.

eliasSFL:
> I think Saphiro is already being applied on the back-end but 3 Leaf Clover is not. So back-end needs an update too.

I don't think Saphiro is applied. Cuz if it is Rice and olive should have the same growth time as wheat/radish and kale respectively. But right now it's not the same

eliasSFL:
> I think Saphiro is already being applied on the back-end but 3 Leaf Clover is not. So back-end needs an update too.

@brian-man I don't think Saphiro is applied. Cuz if it is Rice and olive should have the same growth time as wheat/radish and kale respectively. But right now they are not the same

adamhannigan (approved):
Ooooh great spot! Cheers ❤️

PR #3726, opened 2024-05-20
[FIX] Apply Fruit Picker Apron to all fruit

brian-man:
# Description

Fruit Picker Apron should apply to grape.

## Type of change


adamhannigan:
@brian-man I'm not sure in this case the Apron should apply to Grapes.

The Apron only has oranges, blueberries, bananas + apples in the artwork

adamhannigan:
@eliasSFL I will further discuss in today's weekly dev stream. The Apron was never designed to impact all fruit (i.e. the code specifies individual fruits to apply the boost, not all). I think along the way it got mixed up that it will affect all fruits

PR #3882, opened 2024-06-16
[FEAT] Hide captcha during onboarding

adamhannigan:
# Description

During initial gameplay, hide the captcha. This streamlines the game


brian-man:
Only removing captcha during the beginning of the game could arguably be considered misrepresenting the game experience at exactly the point where players are evaluating the game. No?

adamhannigan:
@brian-man It is a fair point.

We often remove friction areas of the game for onboarding. There are bonus coins, time warp totems etc that appear early in the game but not later.

I think the Captcha is a solid part of the gameplay to introduce after 1 day, it should hopefully give the player a sense of security measures in the game.

PR #3894, opened 2024-06-17
[FIX] Collecting animation now respects DisableAnimations

brian-man:


brian-man:
On my Surface Laptop 5, it often becomes unresponsive for multiple seconds.  Here's a screenshot in one such case:
image

c-py (approved):
LGTM!

PR #3909, opened 2024-06-20
[FIX] Disable clicking the Maneki Neko a second time inside the house

andresgargav:
# Description

Disable clicking the Maneki Neko a second time inside the house. This happens because it doesn't recognize any Maneki Neko collectibles, leaving the array `manekiNekos` empty, which always results in a `false` in this operation:

```typescript
const hasShakenRecently = manekiNekos.some((maneki) => {
    const shakenAt = maneki.shakenAt || 0;

    return !canShake(shakenAt);
});
```

Fixes #issue

## Type of change

# How Has This Been Tested?

- Moving the maneki neko in and out of the house to check how it reacts when clicked.
- Changing the shakenAt dates to check if it still works correctly.

# Checklist:


spencerdezartsmith (approved):
Nice catch!

PR #3916, opened 2024-06-21
[FEAT] Simplify some old timestamps

brian-man:


src/lib/flags.ts
```diff
@@ -8,9 +8,7 @@ const defaultFeatureFlag = ({ inventory }: GameState) =>
 const testnetFeatureFlag = () => CONFIG.NETWORK === "amoy";
 
 const clashOfFactionsFeatureFlag = () => {
-  if (testnetFeatureFlag()) return true;
-
-  return Date.now() > new Date("2024-05-01T00:00:00Z").getTime();
+  return true;
```
spencerdezartsmith, line 11: We should actually just remove these old feature flags.

src/lib/flags.ts
```diff
@@ -75,12 +73,7 @@ const featureFlags: Record<FeatureName, FeatureFlag> = {
   },
   FACTION_HOUSE: defaultFeatureFlag,
   EASTER: (game) => {
```
spencerdezartsmith, line 75: This one makes sense to keep as easter rolls around each year 🐰

spencerdezartsmith (changes requested):
I love a good tidy up. I think we can remove the old season flag now.

src/lib/flags.ts
brian-man: I was going to do that in a separate PR.

spencerdezartsmith (approved):
LGTM

PR #3938, opened 2024-06-26
[CHORE] Update Rank Label 

c-py:
Update faction ranks UI so that it shows the label in the same format as the rest of the game

lb


spencerdezartsmith (approved):
👏🏼

PR #3959, opened 2024-07-02
[FEAT] Remove Shelly and Kraken Tentacle trading

brian-man:

PR #3967, opened 2024-07-03
[FEAT] Remove Kraken Tentacle from fishing infra and UX

brian-man:
# Description

Players are confused and perhaps frustrated by the Kraken Tentacle in the Codex, as it is no longer achievable.

This change removes Kraken Tentacle from various locations that treat it as still catchable.


brian-man:
@adamhannigan There are a couple places that hardcode "Kraken Tentacle" as a catch result.  I changed those to "Anchovy".  What's the right approach for those cases?

brian-man:
#3959 needs to merge first

adamhannigan:
@brian-man You have taken the right approach here 👍 That hard coded value was for local testing

adamhannigan (approved):

PR #3984, opened 2024-07-08
[FEAT] Noticeboards

adamhannigan:
# Description

Add the noticeboard concept. A go to place to discover information about a scene. 

Currently added to the Kingdom + Faction Houses.

Screenshot 2024-07-08 at 12 00 55 PM
Screenshot 2024-07-08 at 11 20 24 AM


spencerdezartsmith (approved):
Working well. The map is damaged a little here where it is cutting of some of the decorations. There is also a collision that is a lot bigger than the question mark. It stretches a fair way down at the bottom of the interactable question mark.

Screenshot 2024-07-08 at 12 29 46 PM

PR #4003, opened 2024-07-10
[FEAT] Faction Membership Updates

adamhannigan:
# Description

This PR includes the following functionality:

1. Leave a faction
2. Dynamic prices to join factions (5 SFL for bottom faction, 50 SFL for winning faction) - based on previous week.
3. No longer give emblems when joining a faction

It also cleans up the old pledging events.

Screenshot 2024-07-10 at 7 23 15 PM
Screenshot 2024-07-10 at 7 23 22 PM

**Leaving a faction**

So you're a traitor huh? To leave a faction, you must satisfy the following criteria:

1. No remaining emblems (you must sell them all)
2. You did not recently join (in last 24 hours).

When you leave a faction you will be punished and all marks will be lost as well as your banner.

## How to test?

1. Run against local API
2. Leave a faction (You can find the button at the bottom of the Emblem Board)
3. Join a faction

Ensure when leaving, that all marks and banners are removed.

The previous week winners, will determine the price to join a faction. You should walk around the Kingdom and notice that each recruiter is offering a different price to join - the Champions will be most expensive (50 SFL)

Note, there is a 24 hour cooldown after joining a faction before you can leave. You may need to comment this code out if you want to test multiple factions and/or update the DB pledgedAt value.


Dionis404:
I believe the punishment for betrayal should be more severe. 
For example: 
SFL fine
Loss of all faction items (Decor, Knight's Set)

brian-man:
> I believe the punishment for betrayal should be more severe. For example: SFL fine Loss of all faction items (Decor, Knight's Set)

Adam has previously expressed that he _wants_ some players to (attempt to) collect multiple faction items.  As such, loss of those items on faction switch wouldn't make much sense.

ShinKan42:
I don't think it's at all lucrative to be a traitor if
1. Ppl need to sell emblems by 200 each time until clear up first
2. Ppl lose almost everything from the previous faction

### Lack of Incentives
It's predictable that being a traitor is pure pain.
Other than providing opportunity to be a trader keeping an eye on all factions, Ppl need to commit fully to each factions, then there aren't real incentive to leave at all.
**There should be a middle ground for traders since that is the main reason Ppl would betray.**

### Strategy not reasonable enough to betray
As for the expected all-round Smith, i don't think Ppl would like to be one since you pretty need to either give up the rank bonus or buy up and sell out all those emblems.
- With emblems, you need to go through the 200 trading pain.
- Without emblems, slow mark accumulation is not at all economic or satisfying.

adamhannigan:
Thanks @Dionis404 @brian-man @ShinKan42 For the feedback 👌 

Some great comments, also related to this proposal: https://github.com/sunflower-land/sunflower-land/discussions/4010

We will go forward with the current "Leave faction" as it is useful functionality. Regarding the 'punishment' and restrictions for leaving a faction, we can update based on the outcome of the proposal 👍

PR #4029, opened 2024-07-13
[FEAT] Add Inventory Amount in Buy Screen

brian-man:
This PR allows players to see their inventory amount in the buy screen, so that users don't have to go to their basket to check their stock while buying

image


eliasSFL (approved):
LGTM! Thanks!

PR #4045, opened 2024-07-17
[FEAT] Trade island limits

adamhannigan:
# Description

In the past 2 weeks, there has been a spike of bot activity. We've discovered they are mainly using trading to extract SFL from the game. This not only affects the P2P economy, but overall $SFL as well.

We have increased the bot detection mechanisms in our API. However, we are also introducing daily trade limits to reasonable amounts so the P2P trading is not attractive to bots.

- **Basic Island** = 5 daily trades
- **Spring Island** = 10 daily trades
- **Desert Island** = 20 daily trades.

We will gather feedback on these amounts to ensure it doesn't affect real players.

Screenshot 2024-07-18 at 8 47 52 AM


brian-man:
> Desert Island = 20 daily trades.

This is way too low.  My trade activity would be impaired significantly.  Also this will have a significant impact on the weekly faction  leaderboards and cross-faction competition.

adamhannigan:
@brian-man To clarify, this is only for listings.

There are still unlimited buys. How many listings do you aim for each day?

brian-man:
> @brian-man To clarify, this is only for listings.
> 
> There are still unlimited buys. How many listings do you aim for each day?

Depends on the market.  When something is way up you want to liquidate a lot while the demand is high.  This gimps that.  Also small batches sell more quickly, so this gimps that as well since the metric is listings rather than volume.

When competing for leaderboards it often makes sense to sell X to buy Y.  This will directly impact LB because of that.

adamhannigan:
@brian-man Cheers.

We are targeting botters with API mechanisms.

However, we also don't want this to appear as a good extraction avenue for malicious actors. If we make something too easy, it becomes attractive to the botters.

The battle of bots becomes a game of cat and mouse. They love a challenge (if they see a revenue opportunity). When game's add new mechanisms, it becomes a 'puzzle' for them to solve.

PR #4059, opened 2024-07-19
[FIX] Conditional display of bonusPrizeWeek label

brian-man:


eliasSFL (approved):
LGTM! Thanks for the fix!

PR #4133, opened 2024-07-31
[FEAT] Desert rewards

adamhannigan:
# Description

Find all the missing items for a bonus reward!

Screenshot 2024-07-31 at 12 34 50 PM
Screenshot 2024-07-31 at 12 34 58 PM
Screenshot 2024-07-31 at 12 35 02 PM
Screenshot 2024-07-31 at 12 35 18 PM
Screenshot 2024-07-31 at 12 35 22 PM

## How to test?

1. Connect to local API
2. Find all missing items
3. Claim prize at Digby


spencerdezartsmith (dismissed):
Working well. There is one issue with the modal opening to reveal the prize. There is logic inside of Digby to determine which tab to show. If there are no digs remaining then it will default to the "Extras" tab. If its opening and a player has all the items then we will want to open on the "Patterns" tab so they can claim the reward.

PR #4137, opened 2024-07-31
[FIX] Increase desert tool stock

adamhannigan:
# Description

Bump desert tool stock


eliasSFL:
@adamhannigan Should we let the warehouse affect the desert tool stock as well?

adamhannigan:
@eliasSFL For simplicity we will leave that for the main tools.

This increase should solve problems with digging - I don't think people should be restricted by the desert tool stock since it is tied to the minigame. If we need to increase it, I would prefer to just increase the stock for everyone (not just for Warehouse)

brian-man:
> @adamhannigan Should we let the warehouse affect the desert tool stock as well?

@eliasSFL 
Do you mean Toolshed?

PR #4180, opened 2024-08-05
[FIX] ts-jest memory leak

c-py:
# Description

Why this started happening is unknown, however there seems to be a suggestion to set `isolatedModules` here: https://github.com/kulshekhar/ts-jest/issues/1967


0xSacul (approved):
LGTM 🚀

PR #4196, opened 2024-08-07
[FEAT] Remove cloneDeep

c-py:
# Description

Wraps every handler in an immer `produce` function. This library allows you to perform immutable updates using `mutable` syntax. e.g.

```js
hello.world = 1
``` 

instead of 

```js
{...hello, world: 1}
```

Immer is referenced in the xstate docs: https://stately.ai/docs/immer

Why:
- This removes cloneDeep from the code (which takes 6-8ms per gameState cloned).
- `useSelector` will not trigger a re-render unless the object truly reference changed. 

### Regression Test

1. Click around the game, chop a tree, harvest plants etc and ensure that the game renders correctly.

### Memoize Test

Test `useSelector` by:

1. Checkout `main`
1. Memoize The `CodexButton` and add a log to check if it renders

```diff
--- a/src/features/island/hud/components/codex/CodexButton.tsx
+++ b/src/features/island/hud/components/codex/CodexButton.tsx
@@ -17,7 +17,7 @@ const _delivery = (state: MachineState) => state.context.state.delivery;
 const _level = (state: MachineState) =>
   getBumpkinLevel(state.context.state.bumpkin?.experience ?? 0);
 
-export const CodexButton: React.FC = () => {
+export const CodexButtonFC: React.FC = () => {
   const [isOpen, setIsOpen] = useState(false);
 
   const { gameService } = useContext(Context);
@@ -25,6 +25,7 @@ export const CodexButton: React.FC = () => {
   const deliveries = useSelector(gameService, _delivery);
   const level = useSelector(gameService, _level);
 
+  console.log({ render: "CodexButton", deliveries, level });
   const hasDeliveries =
     // Show if any new orders has popped up (but not for new players)
     (hasNewOrders(deliveries) && level >= 2) ||
@@ -143,3 +144,5 @@ export const CodexButton: React.FC = () => {
     </div>
   );
 };
+
+export const CodexButton = React.memo(CodexButtonFC);
```

3. Play the game and review the logs
4. Checkout `feat/immer-3` and see if the CodexButton still re-renders on state change

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

localhost

# Checklist:

- [X] Title of the PR is relevant and is prefixed with [FEAT], [CHORE] or [FIX]
- [ ] Screenshot if it includes any UI changes
- [ ] I have read the contributing guidelines and agree to the T&Cs
- [ ] I have performed a self-review of my own code
[...]

PR #4218, opened 2024-08-11
[CHORE] Remove deep clone

c-py:
# Description

This PR is a follow on from: https://github.com/sunflower-land/sunflower-land/pull/4196

Instead of doing a single `produce` call, this new version introduces `produce` into each domain function. This is to make the change less magic, reduce the chance invisible errors in the future.

In each function the following has been changed:

```ts
const stateCopy = cloneDeep(state)
domainLogic()
return stateCopy
```

to 

```ts
return produce(state, (stateCopy) => {
    domainLogic()
    return stateCopy
})
```

The test steps are the same as in: https://github.com/sunflower-land/sunflower-land/pull/4196

## Type of change

# How Has This Been Tested?

localhost

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "gameanalytics": "^4.4.5",
     "howler": "^2.2.3",
     "i18next": "^23.6.0",
+    "immer": "^10.1.1",
     "jwt-decode": "^3.1.2",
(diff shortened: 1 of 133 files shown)


adamhannigan (dismissed):
🏃 🏃 🏃 🏃 🏃

PR #4319, opened 2024-08-30
[FIX] Prevent Wallets Showing Twice

c-py:
# Description

- Prevent EIP-6963 wallets showing twice
- Prevent Metamask showing twice in the list


eliasSFL (approved):
LGTM!

PR #4345, opened 2024-09-02
[FIX] Mobile Dapp Login

c-py:
# Description

Adds a `Web3 Wallet` Login button for injected providers that do not support EIP-6963.

## Type of change

# How Has This Been Tested?

localhost

# Checklist:


eliasSFL (approved):
LGTM!

src/features/wallet/Wallet.tsx
```diff
@@ -52,18 +52,29 @@ const WrappedWallet: React.FC<Props> = ({
...
+    const walletServiceAddress = walletService.state.context.address;
+
+    // Metamask Mobile accidentally triggers this on route changes
+    const didChange = address !== walletServiceAddress;
+    if (
+      didChange &&
+      (
+        ["initialising", "signing", "linking", "minting", "migrating"] as const
+      ).every((state) => !walletState.matches(state))
+    ) {
```
c-py, line 67: This is to stop wallet machine from resetting while we are in the middle of a service. An example of this is when a player connects their wallet in `initialising`. No need to reset the wallet machine as it was us who invoked this. We should only react to wallet changes when the machine is idle.

PR #4357, opened 2024-09-03
[FEAT] Update ranks

adamhannigan:
# Description

Update the ranks based on the latest percentile data.

I have refactored the code so the rank data is split up from the details (image and description). This makes it easier to keep in sync with the backend data.


0xSacul (approved):
LGTM 🚀

PR #4428, opened 2024-09-13
[FIX] Showing German Terms

Droidze:
This PR fixes the German language configuration to show the terms inside the game.
| Before | After |
|--------|--------|
| image | image |


eliasSFL (approved):
Oops I forgot to add that line in after generating the German terms. Thanks for helping me add!

PR #4430, opened 2024-09-16
[FEAT] Tax free

adamhannigan:
# Description

This PR supports tax free withdrawals for any SFL that was deposited or earned in marketplace - this is to avoid double taxes.

Screenshot 2024-09-16 at 10 24 11 AM

## How to test?

Update your `.env` with the updated Withdrawals contract.

1. Deposit SFL into the game
2. Go to withdraw (you should not be taxxed)
3. Go to withdraw a second time (you should be taxed on any excess)
4. 
## TODO

Deploy + destroy mainnet contraacts


spencerdezartsmith (approved):
Working well. Tested both:
- deposit -> withdraw
- withdraw -> deposit -> withdraw

src/lib/utils/tax.ts
```diff
@@ -5,32 +5,51 @@ import { GameState } from "features/game/types/game";
...
+}): number {
+  amount = amount.sub(game.bank.taxFreeSFL);
+
+  if (amount.lte(0)) return 0;
+
+  let percentage = getTaxPercentage(amount);
+
   if (game.island.type !== "basic") {
-    tax -= 25; // 2.5% reduction
+    percentage -= 2.5; // 2.5%
```
brian-man, line 45: Is this going to become float or decimal?  You want to ensure decimal, right?

adamhannigan: I'm not too worried about float versus Decimal in this case, the business logic is handled on the backend when preparing this.

This util is mainly for UI purposes where there is small chance of funky behaviour (we typically cut off to 2 decimal places in displays though)

adamhannigan:
I am closing this PR until marketplace work is resumed.

c-py (approved):
LGTM! Working well!

src/lib/utils/tax.ts
brian-man: @adamhannigan My `bank/taxfreeSFL` value bounces (increases/decreases) relative to "fixed point math" because of float issues with incrementing (via marketplace) or decreasing (via withdrawals).  So there are still (Dec 26 2026) issues at least with the back-end and perhaps with the front-end too.

PR #4435, opened 2024-09-16
[FEAT] Price history

adamhannigan:
# Description

This PR includes a basic charting to display price history of an item. I've opted not to use a charting library as they are quite bloated & difficult to get correct styles with different fonts + retro look.

Screenshot 2024-09-16 at 4 16 46 PM

## How to test?

The history updates every 10 minutes.

1. Point to API
2. Fulfill a trade
3. Chart updates


brian-man:
Will this API and UI be coming to resources (plaza trading)?

adamhannigan:
@brian-man Yes this will be available when resources are added into the marketplace as well 👍 

Not confirmed whether it will be publicly accessible in API format yet

PR #4454, opened 2024-09-23
[FIX] Fix buds showing in chest

adamhannigan:
# Description

Buds should not be showing in the chest when placed.


eliasSFL (approved):
LGTM!

PR #4463, opened 2024-09-25
[FEAT] Gems

adamhannigan:
# Description

Introduces Gems into Sunflower Land. Replaces all occurences of BBs with Gems.

Allows players to exchange BBs for Gems at a 1:100 ratio. This modal will pop up if a player ever has a Block Buck in their inventory - this could catch cases where BBs are escrowed in Auctions or other cases they could reappear.

Smaller gem amounts will help us with upcoming initiatives. We can also reduce Gems required for the following instantly:

1. Digging
2. Fishing Lures
3. Expansions

Screenshot 2024-09-25 at 3 19 30 PM
Screenshot 2024-09-25 at 3 19 33 PM

Screenshot 2024-09-26 at 2 29 59 PM

## How to test?

Update your `.env` - `VITE_BUY_GEMS_CONTRACT=0x6C9DB9fdA8b99fD5c17827ADD1926ab6dE725786`

1. Connect to local API
2. Run the game, migration modal should pop up (if you have Block Bucks).
3. Test buying Gems
4. Test using Gems on Restocks, Auctions, Farm Hands, Digs, Lures etc


Dionis404:
I hope the Gems are not going to be the NFT? 

It will be just a game currency and we can customize the purchase in Xsolla

adamhannigan:
@Dionis404 It will work the same way as Block Bucks currently work 👍

Dionis404:
> @Dionis404 It will work the same way as Block Bucks currently work 👍

It's just that right now BB is the SFT. 
And this may cause some difficulties in some countries to buy them via card or other ways not related to Web3. 
As far as I know Apple has some kind of restriction for Web3 games. It is forbidden to buy digital goods with cryptocurrency. 
Telegram introduced “Stars” because of this.

adamhannigan:
Gotcha @Dionis404 

The plan is to eventually make these off chain items. For speed of release it will use the same architecture as Block Bucks and be an SFT.

Fortunately, since it is not a tradeable cryptocurrency it does not get classified under the regulations of Apple + Xsolla.

The tricky part is that they can be used to attain NFTs in game - that is the major hurdle that platforms have

brian-man:
I know it's preliminary numbers but the methodology for calculating the percentages seems off.
```
20 vs 19
25 vs 23.9
38 vs 36.2
45 vs 43.6
50 vs 48.5
```
image

brian-man:
The post (twitter?) said:
> 1:20 Block Buck to Gem conversion rate

The PR description says:
> Allows players to exchange BBs for Gems at a 1:100 ratio.

That's a pretty significant mismatch.

Dionis404:
I think it's better to make it 1:100 to make it easier to add paid features at a lower cost in the future.

adamhannigan:
Thanks @Dionis404 @brian-man 

Yes unfortunately this PR description is outdated, it is actually 1:20.

Also interesting to point out those % values 👍 I did those calculations based on the whole numbers $1, $5 etc. You are right, when we account for the extra 0.01 it does skew those values off.

Good catch, we can update these ✅

Dionis404:
@adamhannigan  
Take a look at this idea: https://github.com/sunflower-land/sunflower-land/discussions/4467

PR #4480, opened 2024-10-02
[FEAT] Add Label in Faction Shop Item Details if player is not in a Faction

eliasSFL:
# Description

image

Fixes #issue

# Checklist:

### DIFF
--- a/src/features/world/ui/beach/treasure_shop/TreasureShopBuy.tsx
+++ b/src/features/world/ui/beach/treasure_shop/TreasureShopBuy.tsx
@@ -33,7 +33,6 @@ import { COLLECTIBLE_BUFF_LABELS } from "features/game/types/collectibleItemBuff
 import lightning from "assets/icons/lightning.png";
 import { getToolPrice } from "features/game/events/landExpansion/craftTool";
 import { Keys } from "features/game/types/game";
-import { hasFeatureAccess } from "lib/flags";
 import { isMobile } from "mobile-device-detect";
 
 interface ToolContentProps {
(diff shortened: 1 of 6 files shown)


spencerdezartsmith:
Hey @eliasSFL I think that the wording here is too strong. It almost implies that the player is doing something wrong by not being in a faction. 

Could we maybe just make this label a warning/info which says something like "Faction members only"?

eliasSFL:
> Hey @eliasSFL I think that the wording/label colour here are too strong. It almost implies that the player is doing something wrong by not being in a faction.
> 
> Could we maybe just make this label a warning/info which says something like "Faction members only"?

@spencerdezartsmith 

Something like this?
image

spencerdezartsmith:
> > Hey @eliasSFL I think that the wording/label colour here are too strong. It almost implies that the player is doing something wrong by not being in a faction.
> > Could we maybe just make this label a warning/info which says something like "Faction members only"?
> 
> @spencerdezartsmith
> 
> Something like this? image

Is the full width label something new? It feels overwhelming too me 😅

eliasSFL:
> Is the full width label something new? It feels overwhelming too me 😅
@spencerdezartsmith Hmmm I made it full width to be in line with how the buttons are full width as well. Not too sure if that's the right approach

spencerdezartsmith:
> > Is the full width label something new? It feels overwhelming too me 😅
> > @spencerdezartsmith Hmmm I made it full width to be in line with how the buttons are full width as well. Not too sure if that's the right approach
Screenshot 2024-10-02 at 1 25 34 PM

I think its better to just keep our labels as the size of the text they contain. Can this just be listed along with the exp amount? It could even be default colour and change the text to be "Faction member exclusive" with no exclamation. You can use the mark icon. 

Also as and aside... can we fix the background of the images?

eliasSFL:
> I think its better to just keep our labels as the size of the text they contain. Can this just be listed along with the exp amount? It could even be default colour and change the text to be "Faction member exclusive" with no exclamation. You can use the mark icon.
> 
> Also as an aside... can we fix the background of the images?

@spencerdezartsmith Something like this?
image

I'll figure out the issue with the background after

eliasSFL:
image

Okay figured out the background issue

spencerdezartsmith:
I think it should say `Faction member exclusive` which is more inviting and generates more excitement about joining. No `!`. The spacing between the labels can be reduced by one also.

eliasSFL:
@spencerdezartsmith 
image

spencerdezartsmith:
> @spencerdezartsmith image

The right colour might be blue which indicate "information". What do you think?

eliasSFL:
> The right colour might be blue which indicate "information". What do you think?

Hmm the not-in-correct faction label used danger as well, do you think we match it or we use different questions for it?
image

eliasSFL:
> The right colour might be blue which indicate "information". What do you think?

@spencerdezartsmith 
Hmm the not-in-correct faction label used danger as well, do you think we match it or we use different questions for it?
image

spencerdezartsmith:
> > The right colour might be blue which indicate "information". What do you think?
> 
> Hmm the not-in-correct faction label used danger as well, do you think we match it or we use different questions for it? image

Yeah but only faction members can see this right? Thats a bit different because everyone already is in a faction. 

I am just hesitant on the other one because new players may feel they are doing something wrong by being hit with an error label. Having it be an exclusive thing rather than a wrong thing is better imo.

spencerdezartsmith:
> image
> 
> Okay figured out the background issue

I think just go with this one 😄

eliasSFL:
@spencerdezartsmith Alright I've updated the colours
image

spencerdezartsmith (approved):
LGTM

PR #4502, opened 2024-10-08
[FIX] awarded

brian-man:


eliasSFL (changes requested):
Same here, don't edit en.json. only edit dictionary.json

eliasSFL (approved):

PR #4511, opened 2024-10-10
[FIX] Remove Default Caching Policy

c-py:
This PR attempts to improve plaza performance by removing the default handler in the service worker.

The hope is that requests for non service worker assets cached assets will go via the standard browser HTTP cache.


spencerdezartsmith (approved):
🙏🏼

PR #4514, opened 2024-10-10
[FIX] CropName sorted by base duration

brian-man:
Moves `Soybean` to the proper location in the `CropName` order (matching the `CROPS` order).


c-py (approved):
LGTM!

PR #4517, opened 2024-10-10
[CHORE] Add Docs Link to Composters Modal

Droidze:
image

Select composters, then guide tab and click the Docs label in the top left.


c-py (changes requested):
@Droidze We are moving towards having all information in game, because there have been instances in the past where the docs end up having information different to the in-game mechanic. 

Is there information missing in the in-game guide? Do you think it's possible to update the information directly in the in-game guide instead of linking to the external documentation page?

Droidze:
@c-py Thanks for your answer mate, you're right, I'll close the two PRs related to adding Docs links. Also good idea, might think later about how can add more info to guide tabs inside the game instead of Docs link. 🌻

PR #4520, opened 2024-10-11
[FIX] Crafting Box

c-py:
# Description

This building can be used to craft items from recipes.

Adds two new events:

`crafting.started`
`crafting.collected`

**Not implemented: Recipes and Recipe Discovery**

1. Build Crafting Box at Blacksmith
2. Change readyAt to be 1
3. Test a success scenario: 
  a. Craft an Dirt Path (One stone in position 1)
  b. Collect the path.
4. Test a failure scenario:
  a. Put one Wood in position 1
  b. See the fail message

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas


src/features/game/types/game.ts
```diff
@@ -1363,6 +1363,13 @@ export interface GameState {
   experiments: ExperimentName[];
   henHouse: AnimalBuilding;
   barn: AnimalBuilding;
+
+  craftingBox: {
+    status: "pending" | "idle" | "crafting";
```
adamhannigan, line 1368: Could we derive these from other fields (i.e. item and startedAt)?

adamhannigan (approved):
LGTM!

Felt smooth - I liked the pause when using a fake recipe as well. Just a few small UI suggestions:

1. The Crafting Box placement feels a bit off. Could we center the base of the building?
2. The Red Cross when a fail happened I missed the first few times. Consider changing the label to red or some other feedback 

LFG!

c-py:
@adamhannigan Will take the feedback into consideration when I work on the recipes.

PR #4535, opened 2024-10-14
Chest Reward verification improvement - Second Edition

brian-man:
The first edition attempted to use the new player experience, but there seem to be bugs in that implementation.  Additionally, for seasoned players the experience should be improved to remove the extra clicks entirely.  More specifically, the "click again after discovering a reward is present on the plot" (for crops only) and "click on the woohoo UI" (for both crops and trees).


adamhannigan (approved):
Nice work! 👏

PR #4540, opened 2024-10-15
[FIX] Pull Changes from origin before pushing

eliasSFL:
# Description

Currently the translate and metadata workflows will fail if a commit happens while the workflows are in progress. 
I've added a check where it will fetch and pull latest changes in main branch before committing the json files in

### Fixes #issue
Bot commit fails if a new commit is pushed while the workflow is running, regardless of whether the commit creates conflicts or not

See Commits to see the tests I did
- I changed the file that will trigger the workflow and committed it
- while the workflow is running, I pushed another commit
- check to see if workflow fails or not

# Checklist:


c-py (approved):
LGTM!

It looks like the add-and-commit action has an optional `pull` setting. Using this setting may reduce the number of steps in the github action:

https://github.com/EndBug/add-and-commit/issues/498

PR #4557, opened 2024-10-17
[FIX] Radish and Banana descriptions at Market

brian-man:


HinataShoyo5106:
bananas and orange had the exact same phrase when selling them for coins. i suggest they change it for like ''With potassium in the banana, your energy is always at hand!'' or something like that.
also radish and corn are repeated too.
give radish something like: "Plant, harvest, and reap the crunchy rewards of radishes—straight from the heart of Sunflowerland’s bountiful fields!"

PR #4568, opened 2024-10-20
[FIX] Fix Factions' Join Cost

Droidze:
# Description
### fix the bug that causes factions' join cost changes each time.

# Before

https://github.com/user-attachments/assets/c9aae86f-22e4-4530-8873-eed5026ef2a1

---
# After

https://github.com/user-attachments/assets/3a5c74c0-1bd6-4f8e-a942-a157b56c5afe

Fixes #issue

try to join different factions at kingdom

### DIFF
--- a/src/features/world/ui/factions/JoinFaction.tsx
+++ b/src/features/world/ui/factions/JoinFaction.tsx
@@ -37,6 +37,7 @@ interface Props {
 const _joinedFaction = (state: MachineState) => state.context.state.faction;
 const _username = (state: MachineState) => state.context.state.username;
 const _farmId = (state: MachineState) => state.context.farmId;
+const sflCostReverse = SFL_COST.reverse();
 
 export const JoinFaction: React.FC<Props> = ({ faction, onClose }) => {
   const { gameService } = useContext(Context);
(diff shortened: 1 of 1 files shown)


src/features/world/ui/factions/JoinFaction.tsx
```diff
@@ -37,6 +37,7 @@ interface Props {
 const _joinedFaction = (state: MachineState) => state.context.state.faction;
 const _username = (state: MachineState) => state.context.state.username;
 const _farmId = (state: MachineState) => state.context.farmId;
+const sflCostReverse = SFL_COST.reverse();
```
c-py, line 40:
```suggestion
```

src/features/world/ui/factions/JoinFaction.tsx
```diff
@@ -84,7 +85,7 @@ export const JoinFaction: React.FC<Props> = ({ faction, onClose }) => {
       });
 
       const position = totals.indexOf(faction);
-      const fee = SFL_COST.reverse()[position] ?? 10;
+      const fee = sflCostReverse[position] ?? 10;
```
c-py, line 88: Great spotting on the mutating array!

Would it be possible to perform a shallow clone before the `reverse`, so that the `SFL_COST` array does not get mutated?

```suggestion
      const fee = [...SFL_COST].reverse()[position] ?? 10;
```

c-py (changes requested):
I just dropped one piece of feedback. Nice spotting @Droidze !

PR #4573, opened 2024-10-21
[FIX] Clear Crafting Details on Gem Boost

c-py:
# Description

Small UI bug on smoothie shack after instant craft. In progress section would have out of date data in machine.

Before:

After:


spencerdezartsmith (approved):
LGTM

PR #4578, opened 2024-10-22
[FEAT] Restrict Scene Transition in MMO Based on Required Level

Droidze:
# Description

add restrictions for changing the scenes and entering different areas that need a specific bumpkin lvl. Also, show the label based on the requirement for that area.

image image


spencerdezartsmith:
Thanks for this. Are you able to pull the Button out of the inner div? Typically our buttons should be siblings to the modal content so they have no padding from the panel applied to them.

Droidze:
> Thanks for this. Are you able to pull the Button out of the inner div? Typically our buttons should be siblings to the modal content so they have no padding from the panel applied to them.

@spencerdezartsmith My pleasure 🙏

Just a question, are you referring to the "Back" button mate? just want to make sure.

Droidze:
@spencerdezartsmith pulled the `Back button` out of the Div and also added `IslandNotFound` if the players typed the wrong URL. If anything needs to be changed please tell me.

Droidze:
to keep consistency between `IslandNotFound` and Forbidden, Modal added.

image

PR #4630, opened 2024-10-29
[Feat] Halloween Event Island

telk2:
# Description

This PR adds the Halloween Island for the event. 
The island already can be accessed from the world map, and all new text is already translated in all supported languages for the game.


adamhannigan (dismissed):
LGTM 👌 

Small suggestion, put the "?" disc icon on top of the noticeboard. (Similar to our other noticeboards). It wasn't clear that I could click on it

PR #4637, opened 2024-10-30
[FEAT] Update interiors

adamhannigan:
# Description

1. Update interior images for each size of barn and hen house
2. Move Feeder into center
3. Move upgrade to top left

**Testing**

You may need to bust the cache to get the latest images.

Screenshot 2024-10-30 at 3 56 11 PM


c-py (approved):
LGTM! ✅

PR #4647, opened 2024-10-31
[FIX] Halloween Map collisions

telk2:
# Description
Fixing collisions on the halloween map


c-py (approved):
Tested new collisions @telk2 !

Working well!

PR #4665, opened 2024-11-01
[FEAT] Update crop types to clearly indicate plot-specific crops

brian-man:
There are multiple instances of `isCrop()` that have inconsistent implementations.  The root cause appears to be that "plot crops" were never formally renamed from "crops" when greenhouse crops were introduced.

Additionally the predicate for one implementation for `isCrop()` was inaccurate.

This change includes the following:
- rename `CROPS` to `PLOT_CROPS`
- rename `CropName` to `PlotCropName`
- rename `CROP_LIFECYCLE` to `PLOT_CROP_LIFECYCLE`
- rename `CROP_SEEDS` to `PLOT_CROP_SEEDS`
- rename `CropSeedName` to `PlotCropSeedName`
- rename `Crop` to `PlotCrop`
- rename `isCrop` to `isPlotCrop` in harvest.ts
- fix predicate for `isCrop` in getBudYieldBoosts.ts

### DIFF
--- a/src/features/builder/components/ResourcePlacer.tsx
+++ b/src/features/builder/components/ResourcePlacer.tsx
@@ -20,7 +20,7 @@ import { Layout } from "../lib/layouts";
 import { Boulder } from "features/island/boulder/Boulder";
 import { Beehive } from "features/game/expansion/components/resources/beehive/Beehive";
 import { SUNNYSIDE } from "assets/sunnyside";
-import { CROP_LIFECYCLE } from "features/island/plots/lib/plant";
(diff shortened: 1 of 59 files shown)


eliasSFL:
We rather have our constants to be the same between FE and BE, besides it doesn't affect in game functionality

brian-man:
> We rather have our constants to be the same between FE and BE, besides it doesn't affect in game functionality

It matters b/c the use of the term "Crop" is inconsistent and players can't even read the front end in a clear way.  Especially since "isCrop" has multiple definitions that don't match.  It's chaos and bad for the game and players.

@adamhannigan

PR #4669, opened 2024-11-02
[CHORE] Disable Animal Teaser(Coming Soon Boat)

Droidze:
# Description

The season already started on November 1st, so the date on the message is not valid now. Also, we can merge it on November 4th along with releasing the Animal Update to the public.
image


eliasSFL (approved):
LGTM! @Droidze 

FYI @LittleEins

PR #4670, opened 2024-11-02
[FEAT] Hide Date Label For Seasonal Artifacts When Season Ends

Droidze:
# Description

To prevent potential confusion in the future, the label indicating the date when seasonal artifacts can be obtained through gameplay will be hidden once the corresponding season for the artifact ends.

| Before | After |
|--------|--------|
| image | image | 

- click the NPC called Jafar and check the seasonal artifacts there.


eliasSFL (approved):
LGTM @Droidze! I've updated your code with a simpler logic

PR #4700, opened 2024-11-04
[CHORE] Remove Craft Decorations and Wardrobe Modal

Droidze:
# Description

This PR removes the wardrobe modal and craft decorations codes as a result of the crafting system's public release and also prevents showing the following message that is outdated.


c-py (approved):
Excellent! Thank you @Droidze !

PR #4712, opened 2024-11-05
[FEAT] Remove New label from Beach on World Map

brian-man:


c-py (approved):
The _sands_ of time shift. What's new becomes old, what's old becomes ancient, and what's ancient becomes forgotten. 

Uh, looks good to me!

PR #4719, opened 2024-11-06
[FIX] Wrong XP showing up on love animal

c-py:
# Description

`chicken.item` was being updated before the xp was being rendered. This update uses react state to keep track of the item just used.

Pet Animal

# Checklist:


spencerdezartsmith (approved):
LGTM

PR #4722, opened 2024-11-06
[FIX] Composter text for Sprout Mix with Knowledge Crab placed

brian-man:


eliasSFL (commented):
@brian_man Why use the long descriptions for the labels?

brian-man:
> @brian_man Why use the long descriptions for the labels?

Because there isn't something like `boostedDescriptions` in the short form?

eliasSFL:
> > @brian_man Why use the long descriptions for the labels?
> 
> Because there isn't something like `boostedDescriptions` in the short form?

@brian-man 
Not sure if you know what you're changing but this is the label that you're trying to change. If anything you should be updating `BUMPKIN_ITEM_BUFFS_LABELS`
image

PR #4729, opened 2024-11-07
[FEAT] Standardize on hasVipAccess()

brian-man:

PR #4755, opened 2024-11-11
[FIX] Animal food discount

brian-man:
# Description

Animal food discount for Bull Run season was not being applied for Lifetime Farmer Banner or for in-inventory Bull Run Banner.

As per discussion that led to #4729 the proper check is `hasVipAccess()`.


spencerdezartsmith (approved):
LGTM

PR #4756, opened 2024-11-11
[FEAT] Modernize Mutant Animal Modal

Droidze:
# Description
the following changes have been made in this PR:
- Label to differentiate seasonal and permanent mutant animals
- Label to show mutant animals boost
- Update style

| Before | After |
|--------|--------|
| image | image |  
---
image
image
image

### DIFF
--- a/src/features/farming/animals/components/MutantAnimalModal.tsx
+++ b/src/features/farming/animals/components/MutantAnimalModal.tsx
@@ -9,7 +9,7 @@ import { Button } from "components/ui/Button";
 import { useAppTranslation } from "lib/i18n/useAppTranslations";
 import { Label } from "components/ui/Label";
 import chest from "assets/icons/chest.png";
-import lightning from "assets/icons/lightning.png";
+import Crown from "assets/icons/vip.webp";
 import { ITEM_DETAILS } from "features/game/types/images";
 import { SUNNYSIDE } from "assets/sunnyside";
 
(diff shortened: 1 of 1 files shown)


src/features/farming/animals/components/MutantAnimalModal.tsx
```diff
@@ -85,24 +21,52 @@ interface Props {
 
 export const MutantAnimalModal = ({ mutant, show, onContinue }: Props) => {
   const { t } = useAppTranslation();
+  const boost = COLLECTIBLE_BUFF_LABELS[mutant];
+  const isPermanentMutant = (name: string): boolean => {
+    const permanentMutants = ["Speed Chicken", "Rich Chicken", "Fat Chicken"];
+    return permanentMutants.includes(name);
+  };
```
eliasSFL, line 28: Speed, Rich and Fat chicken are no longer able to be dropped from chickens anymore, so they aren't permanent anymore

src/features/farming/animals/components/MutantAnimalModal.tsx
```diff
@@ -85,24 +21,52 @@ interface Props {
...
+          </Label>
+          {isPermanentMutant(mutant) ? (
+            <Label className="mr-0.5" type="vibrant" secondaryIcon={Crown}>
+              {t("permanent")}
+            </Label>
+          ) : (
+            <Label type="info" secondaryIcon={SUNNYSIDE.icons.stopwatch}>
+              {t("seasonal")}
+            </Label>
+          )}
```
eliasSFL, line 44: These labels are a little bit redundant since only seasonal ones can drop now

eliasSFL (changes requested):
Left a few comments, the 3 chickens that previous dropped as mutants are no longer dropping anymore. only the seasonal ones drop

src/features/farming/animals/components/MutantAnimalModal.tsx
spencerdezartsmith: I think you can leave the seasonal one because it reinforces that all mutants are now seasonal but yeah the permanent one can be removed.

Droidze: Thank you for the reviews 🙏. I was working on making it clear to players that mutant animals only drop during the season they're introduced.
image


But if the season's name is long, it doesn't look good, and  `Seasonal` text shows better in the modal. with respect, I will change it to just show `Seasonal` text.
image

spencerdezartsmith (approved):
LGTM

PR #4760, opened 2024-11-12
[FEAT] Add 'different' to the Stella tier unlocking text

brian-man:
At least a dozen times in discord people have asked if the items need to be different to trigger the unlocking progression.

Might as well make the UX explicit.


adamhannigan (approved):
Agreed ✅ 👍

PR #4771, opened 2024-11-14
[FEAT] Marketplace Hoarding

adamhannigan:
# Description

To avoid hoarding issues, we will perform any trade (over 250 SFL) through `Marketplace.sol`

If the items are not yet on chain, they will be minted and transferred during the transaction.

# How to test?

Update your `.env` to the latest version of the contract

`VITE_MARKETPLACE_CONTRACT=0x6B4c0eE875eD7F7597687a8EF35b56243FA63a47`

You will want to test the following scenarios.

1. Listing/Offering an off chain item (that is already synced on the blockchain) for over 250 SFL.
2. Listing/Offering an off chain item (**not on the Blockchain**) for over 250 SFL.

In the second example, you should see the item is minted in the transaction and then transferred.

Screenshot 2024-11-14 at 3 03 55 PM


spencerdezartsmith (approved):
LGTM! Tested and working well.

PR #4773, opened 2024-11-14
[FEAT] Feeder icon

adamhannigan:
# Description

For consistency with other clickable elements, add an icon.

Screenshot 2024-11-14 at 4 04 50 PM


eliasSFL (approved):
LGTM!

spencerdezartsmith:
To me that bag looks way too big for the disc 😅 The kernel blend might be better or even a barley, corn and wheat icon 🤔

PR #4789, opened 2024-11-20
[FEAT] Marketplace testing

adamhannigan:
# Description

This PR enables marketplace **instant trades** for testing.

The UI matches the API and the following are restricted:

- On chain trades **(mainnet only)**
- Viewing non resources in the old P2P trade views
- Enabling marketplace for Beta Pass


eliasSFL (approved):
LGTM!

PR #4790, opened 2024-11-20
[FEAT] Modernize Claim Reward Modal

Droidze:
- Show consumables xp based on farm buffs instead of their base xp
- Add buff label for wearables
- Only show description of collectibles and wearables if they do not have any boost
- Update style

| Before | After |
|--------|--------|
| image | image |


eliasSFL (approved):
LGTM @droidze! I've changed the code to use the card variant for some touch

Droidze:
> I've changed the code to use the card variant for some touch
@eliasSFL  ok, just seems that the Card variant and the default one(SUNNYSIDE.ui.primaryButton) are the same.

eliasSFL:
> > I've changed the code to use the card variant for some touch
> 
> @eliasSFL  ok, just seems that the Card variant and the default one(SUNNYSIDE.ui.primaryButton) are the same. 

It's ever so slightly different, around the border area

PR #4841, opened 2024-12-01
[FEAT] SC improvements

adamhannigan:
# Description

- Improvements to Game.sol security
- Marketplace V1 on mainnet

I've included Fees in the mainnet, since it was needed again (even though not used in app). It feels the most central place to sore this contract.


c-py (approved):
LGTM!

PR #4843, opened 2024-12-02
[FIX] Collectible Search

adamhannigan:
# Description

Treats resource and collectibles as different filters.

## How to test?

1. Connect to local API
2. Go to Resources > Only resources show
3. Go to Collectibles > Only collectibles show
4. Search for something > Both collectibles & Resources show


c-py (approved):
LGTM!

PR #4847, opened 2024-12-02
[FEAT] Do not reload on search

c-py:
# Description

- Turns off reloading when searching
- Fixes a race condition in what was rendered when switching categories: now we only render the last emitted promise on filters change

Try to search and make sure the page doesn't reload

# Checklist:


adamhannigan (approved):
Nice solution 👌

PR #4879, opened 2024-12-05
[CHORE] Bump React Router Version

c-py:
# Description

Clicking the same navigation route seemed to cause re-render. Bump the react router version to prevent it from having errors.


adamhannigan (dismissed):
LGTM ✅ I tested all of the main flows.

- Login
- Farming
- MMO

PR #4900, opened 2024-12-08
[FIX] Wearables Image Loading

Droidze:
# Description
This PR should address the issue that prevents background-less images for wearables from loading in the friendship info panel.


eliasSFL (approved):
LGTM!

PR #4921, opened 2024-12-10
[FIX] Toast Panel Z Index

Droidze:
# Description

This PR fixes the long-standing issue where mobile players do not see the items or XP they gain when performing actions such as buying tools, feeding the bumpkin, completing deliveries, etc. Additionally, it ensures that players can see when items are removed from their inventory.

Fixes #issue


eliasSFL (approved):
LGTM! Thanks for the contribution!

PR #4949, opened 2024-12-14
[FIX] Remove Close Button Panel for Poppy Intro

Droidze:
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/src/features/world/ui/flowerShop/FlowerBounties.tsx
+++ b/src/features/world/ui/flowerShop/FlowerBounties.tsx
@@ -4,6 +4,8 @@ import classNames from "classnames";
 import { Button } from "components/ui/Button";
 import { Label } from "components/ui/Label";
 import { ButtonPanel } from "components/ui/Panel";
+import { CloseButtonPanel } from "features/game/components/CloseablePanel";
+import { SpeakingModal } from "features/game/components/SpeakingModal";
 import {
   generateBountyCoins,
   generateBountyTicket,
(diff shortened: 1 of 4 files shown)


eliasSFL (changes requested):
Instead of putting the speaking modal in NPCModals you could add the close button panel into with the deals component of the flower bounties

src/features/world/ui/flowerShop/FlowerBounties.tsx
```diff
@@ -55,26 +41,6 @@ export const FlowerBounties: React.FC<Props> = ({ readonly }) => {
...
-          acknowledgeIntro();
-          setShowIntro(false);
-        }}
-      />
-    );
-  }
-
   if (deal) {
     return (
       <Deal
```
eliasSFL, line 46: So close button panel here

Droidze:
> So close button panel here, and also at the return one below

@eliasSFL At the first thought about doing the easiest way, but also considered that you added `FlowerBounties` to Codex and could make issues there, so checked that 3 weeks ago you pulled out the `CloseButtonPanel` from `FlowerBounties` and added it to `NPCModals` for Poppy , probably it was not suitable to keep `CloseButtonPanel` inside. 

Briefly that's what happens when adding `CloseButtonPanel` inside `FlowerBounties`, shows it with two panels in Codex, which shouldn't be due to keeping consistency with other sections in this tab:

image

eliasSFL (approved):
I have refactored the flower bounties modal so that we can use one for poppy and one for the codex, and it will also solve the double modal issue

PR #4958, opened 2024-12-16
[FIX] Prevent Preloader from Looping

c-py:
# Description

Prevents the preloader from getting stuck in infinite loop when `sceneId` is not set.


adamhannigan (approved):
Nice catch 👏

PR #4965, opened 2024-12-16
[FIX] Show in Progress Food Details Instead of Selected Food For Boosting with Gems

Droidze:
# Description

This PR fixes the bug in the following modal that causes seeing some details(name, image and xp) related to the food different from the one is currently being cooked, it happens when players click "View recipes", select another food, close the cooking building modal and click the building again before the food is ready.

image


eliasSFL:
@Droidze I might put your PR on hold, as I may remove this modal in my gem boost feedback PR #4970

PR #4967, opened 2024-12-16
[FIX] Marketplace sort

adamhannigan:
# Description

Wearables were showing at the bottom of the marketplace, instead of ordered by price with the collectibles.

This PR copies the sort logic from the backend.

Screenshot 2024-12-17 at 9 52 20 AM


c-py (approved):
LGTM ✅ Working against testnet

PR #5007, opened 2024-12-28
[FEAT] Add setting to disable QuickSelect

brian-man:
# Description

The QuickSelect UI for the Greenhouse has always bothered me.  I know what I want to plant and it's a source of mistimed planting of something I don't want planted.

There was an effort to try this out for normal crop plots and it was very problematic.  It interrupted the flow and introduced the misplanting problem (the wrong crop, and sometimes when you don't even want to plant a crop b/c you're just trying to harvest).

This change introduces a setting to prevent this "helper" UI from doing harm to those of us that don't want or need the feature.

### DIFF
--- a/src/features/farming/hud/lib/quickSelect.ts
+++ b/src/features/farming/hud/lib/quickSelect.ts
@@ -1,0 +1,14 @@
+/**
+ * Cache show timers setting in local storage so we can remember next time we open the HUD.
+ */
+const LOCAL_STORAGE_KEY = "settings.enableQuickSelect";
+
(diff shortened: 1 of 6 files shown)


eliasSFL:
I'll merge this together with my Quick Select for Fruits PR

brian-man:
> I'll merge this together with my Quick Select for Fruits PR once you're ready, and will update the same logic to the fruit patches too

I'd prefer to merge this standalone as it doesn't require or depend on fruits, and shouldn't be "married to it" for revert purposes, etc.

PR #5008, opened 2024-12-29
[FIX] Bud Box types should be evenly distributed

brian-man:
# Description

The yearly transition is disjoint.  Some bud types have an advantage over others.

This change updates and simplifies the algorithm to be fair.

Updated the UI to show all 10 bud types and the corresponding dates of the bud box
|Before|After|
|---|---|
|image|image|


eliasSFL:
LGTM! I'll sync this with BE and approve after new years

I also updated the modal to show all bud types and the dates
image

c-py (approved):
Nice addition! @brian-man

PR #5014, opened 2025-01-02
[FEAT] Cool Plaza Dogs

c-py:
# Description

Adds two dogs into the plaza. Why not

dog

Chase the dogs around the plaza.

# Checklist:


adamhannigan (approved):
Nice work 👏 Just a few small suggestions:

1. Can we add a shadow under the dog (similar to the Bumpkin shadow). They look like they are floating 👻 
2. My first thought was I needed to chase them and click on them - since they are moving I think they will draw the most attention in the plaza. Consider adding a modal when you click on them. It could also be nice if when you clicked on them they ran away 😅

PR #5037, opened 2025-01-10
[FEAT] Daily Reset For Season Events

spencerdezartsmith:
# Description

This introduces a `DailyEvent` component which will fire a no op event on UTC midnight change over to trigger a seasonal event. The game machine will then handle moving the player through an acknowledgement step so they are informed on what has occurred.

There can be more added to this component like a countdown, more "new day" communication. This is not included in this PR.

Fixes #issue

- Inside of `DailyReset` uncomment the test version and comment out the prod version. This will trigger the reset every minute. Once this is done then go the BE and uncomment the tornado for today.
- Come back to the FE and on the next reset you should automatically be notified of the tornado. 
- Acknowledge
- Refresh, tornado should still be in affect but you will not be notified again

# Checklist:


c-py (approved):
LGTM ✅ !

PR #5042, opened 2025-01-12
[CHORE] Improve Codex Performance

eliasSFL:
# Description

This PR converts all the components in the Codex into stateless components, and calls for gameService and gameState from Codex

Fixes #issue
Codex taking very long to load after clicking on the Codex Button

Run FE with API 

Compare the speed of Codex opening between live game and this PR.

From preliminary testing, it reduced the load times by at least 40%

# Checklist:

### DIFF
--- a/src/features/island/delivery/components/Orders.tsx
+++ b/src/features/island/delivery/components/Orders.tsx
@@ -1,6 +1,5 @@
-import { useActor, useSelector } from "@xstate/react";
 import { SUNNYSIDE } from "assets/sunnyside";
-import React, { useContext, useEffect, useState } from "react";
+import React, { useEffect, useState } from "react";
 import classNames from "classnames";
 import Decimal from "decimal.js-light";
 
@@ -11,7 +10,6 @@ import lock from "assets/icons/lock.png";
 import lightning from "assets/icons/lightning.png";
 
 import { DynamicNFT } from "features/bumpkins/components/DynamicNFT";
-import { Context } from "features/game/GameProvider";
 import {
   QuestNPCName,
   TICKET_REWARDS,
@@ -39,7 +37,6 @@ import {
 import { RequirementLabel } from "components/ui/RequirementsLabel";
 import { Button } from "components/ui/Button";
 import { ButtonPanel, InnerPanel, OuterPanel } from "components/ui/Panel";
-import { MachineState } from "features/game/lib/gameMachine";
 import { getSeasonalTicket } from "features/game/types/seasons";
 import { secondsTillReset } from "features/helios/components/hayseedHank/HayseedHankV2";
 import { Revealing } from "features/game/components/Revealing";
@@ -56,6 +53,7 @@ import { isMobile } from "mobile-device-detect";
 import { getImageUrl } from "lib/utils/getImageURLS";
 import { ITEM_IDS } from "features/game/types/bumpkin";
 import { isCollectible } from "features/game/events/landExpansion/garbageSold";
+import { MachineInterpreter } from "features/game/lib/gameMachine";
 
 // Bumpkins
 export const BEACH_BUMPKINS: NPCName[] = [
@@ -81,12 +79,9 @@ interface Props {
   selectedId?: string;
   onSelect: (id?: string) => void;
   onClose: () => void;
+  state: GameState;
+  gameService: MachineInterpreter;
 }
-
-const _delivery = (state: MachineState) => state.context.state.delivery;
-const _inventory = (state: MachineState) => state.context.state.inventory;
-const _sfl = (state: MachineState) => state.context.state.balance;
-const _coins = (state: MachineState) => state.context.state.coins;
 
 export function hasOrderRequirements({
   order,
@@ -141,8 +136,6 @@ export const OrderCard: React.FC<OrderCardProps> = ({
   onClick,
   state,
 }) => {
-  const { coins, balance: sfl, inventory } = state;
-
   const tickets = generateDeliveryTickets({ game: state, npc: order.from });
 
   return (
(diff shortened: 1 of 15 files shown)


src/features/island/hud/components/codex/Codex.tsx
```diff
@@ -52,6 +51,17 @@ export const Codex: React.FC<Props> = ({ show, onHide }) => {
...
+  const {
+    username,
+    bertObsession,
+    npcs,
+    bounties,
+    delivery,
+    choreBoard,
+    kingdomChores,
+    faction,
+  } = state;
```
spencerdezartsmith, line 63: This would be better if you used `useSelector` here for each of these pieces of state.

spencerdezartsmith (commented):
Nice job here. I do have some questions:

- Do we need to pass the service down as a prop? I imagine that if we are only pulling the service out in the child components in order to send events then there would be no real difference here. The context value is the same reference so its relatively stable eg. Not causing rerenders. 

- Rather than using an actor, I think you could create useSelectors for the context.state values you need and then pass them down. The useActor will cause a re render when anything changes in the machine.

src/features/island/delivery/components/Orders.tsx
```diff
@@ -323,25 +315,16 @@ export const DeliveryOrders: React.FC<Props> = ({
...
   onSelect,
   onClose,
+  state,
 }) => {
   const { gameService } = useContext(Context);
+  const { delivery, balance: sfl, coins, npcs, bumpkin } = state;
 
   const navigate = useNavigate();
 
-  const delivery = useSelector(gameService, _delivery);
```
spencerdezartsmith, line 331: This is actually a better pattern. If there is nowhere in the component that needs the entire state passed then this pattern would be preferable. You can continue to remove the useActor and just add a useSelector for npcs and bumpkin. 

Our state object is massive so selectively taking what you need makes sense.

src/features/island/hud/components/codex/Codex.tsx
```diff
@@ -37,20 +36,39 @@ import { hasFeatureAccess } from "lib/flags";
...
+const _username = (state: MachineState) => state.context.state.username;
+const _bertObsession = (state: MachineState) =>
+  state.context.state.bertObsession;
+const _npcs = (state: MachineState) => state.context.state.npcs;
+const _bounties = (state: MachineState) => state.context.state.bounties;
+const _delivery = (state: MachineState) => state.context.state.delivery;
+const _choreBoard = (state: MachineState) => state.context.state.choreBoard;
+const _kingdomChores = (state: MachineState) =>
+  state.context.state.kingdomChores;
+const _faction = (state: MachineState) => state.context.state.faction;
```
spencerdezartsmith, line 57: If you're getting the `_state` above then there is no real benefit in the additional selectors here. You can just grab what you need from `state`. 

I guess the rule is:

If somewhere in the component you need to pass the whole `state` then just create a selector for the `state`. 

If you don't require it then be selective about exactly the values you need.

It also highlights the issue of required props for components or arguments for functions. We should always ask "do we need to pass the whole state?". A lot of the time we just use one value once inside the component/function. Could we just pass that one value instead. 

This comment is not to say you need to change that because it occurs everywhere but its good to think about it in the future.

src/features/island/hud/components/codex/Codex.tsx
```diff
@@ -37,20 +36,39 @@ import { hasFeatureAccess } from "lib/flags";
...
 export const Codex: React.FC<Props> = ({ show, onHide }) => {
   const { t } = useAppTranslation();
   const { gameService } = useContext(Context);
-  const [
-    {
-      context: { state, farmId },
-    },
-  ] = useActor(gameService);
+  const farmId = useSelector(gameService, _farmId);
+  const state = useSelector(gameService, _state);
```
spencerdezartsmith, line 63: Same here as comment above

src/features/island/hud/components/codex/pages/ChoreBoard.tsx
```diff
@@ -236,7 +225,7 @@ export const ChoreBoard: React.FC = () => {
               >
                 {t("chores.complete")}
                 <div className="flex absolute right-0 -top-5">
-                  <ChoreRewardLabel chore={previewChore} />
+                  <ChoreRewardLabel chore={previewChore} state={state} />
```
spencerdezartsmith, line 228: Like here is a perfect example... does a label component really need the entire state 🤔

spencerdezartsmith (approved):
I have left some further comments here. I guess your pattern is mostly just to pass the state down as a prop to the children. I have left some comments around this pattern and my thoughts regarding the passing around of the whole state. 

I am going to approve because this is a small improvement so I don't want to block it but just added some food for thought. If you want to continue to iterate you can and I will review again or if you want to merge and just take the ideas on board thats ok too.

src/features/island/hud/components/codex/pages/ChoreBoard.tsx
eliasSFL: For this case it has a function that requires the whole state inside it

src/features/island/delivery/components/Orders.tsx
eliasSFL: There are places that still require the entire state, so we'll still require passing the state down as a prop

eliasSFL:
@spencerdezartsmith I went through all the items in the codex pages and made some changes to components that doesn't necessarily need the whole state, but it seems like most of the components still require the whole GameState due to functions in the components needing them.

spencerdezartsmith:
> @spencerdezartsmith I went through all the items in the codex pages and made some changes to components that doesn't necessarily need the whole state, but it seems like most of the components still require the whole GameState due to functions in the components needing them.

Yeah I knew that would be the case. Thanks for updating. Again, it was more food for thought for future building. Happy for this to go in.

spencerdezartsmith (approved):

PR #5057, opened 2025-01-16
[FIX] Memoize Collectible Buff Labels

c-py:
# Description

Memoizes collectible buff labels.

On Collection.tsx time how long to filter items before and after change

```ts
  const startTime = performance.now();
  const items =
    data?.items.filter((item) => {
      const display = getTradeableDisplay({
        type: item.collection,
        id: item.id,
        state,
      });

      if (filters.includes("utility") && display.buffs.length === 0) {
        return false;
      }

      if (filters.includes("cosmetic") && display.buffs.length > 0) {
        return false;
      }

      return display.name.toLowerCase().includes(search?.toLowerCase() ?? "");
    }) ?? [];
  const endTime = performance.now();
  console.log(`Filter took ${endTime - startTime}ms`);
  ```

# Checklist:


spencerdezartsmith (approved):
LGTM!

_Before_
Screenshot 2025-01-16 at 9 44 37 PM

_After_
Screenshot 2025-01-16 at 9 46 08 PM

PR #5072, opened 2025-01-18
[FEAT] Skill points disambiguation

brian-man:
The `getAvailableBumpkinSkillPoints()` function has two implementations with different behavior.

This change renames one of them to resolve the ambiguity when reading the code.


eliasSFL (approved):
LGTM! 

We'll probably remove all the old skill point code after skill revamp but good change to keep the code clean

eliasSFL:
I'll merge this after I copy this to the BE

PR #5073, opened 2025-01-19
[FIX] Withdraw SFL integer-only

brian-man:
# Description

Players keep running into the `STF` error when trying to withdraw.

Since the contract currently requires integers, the front-end might as well match.  For now at least.


eliasSFL (approved):
LGTM! Makes sense!

PR #5092, opened 2025-01-23
[FIX] Emblems tradeables

adamhannigan:
# Description

The UI was not disabling the buttons to list and ful-fill offers for factions you are not in.

I have also updated the description of the Emblems - this is a quick way to give more context to readers about why the item is not tradeable.


brian-man:
What's the plan for those of us with (from long ago) "wrong faction" emblems?  Any reason they can't be unlocked for sale?

adamhannigan:
@brian-man We will need to manually sort those out. Opening up for trading could lead to some security issues with the emblems. I'd suggest reaching out on Beta chat to swap these

PR #5144, opened 2025-01-31
[FIX] Node Amounts

adamhannigan:
# Description

Accidentally changing a constant, that was used in multiple calculations 😭


spencerdezartsmith (approved):
LGTM

PR #5169, opened 2025-02-02
[FEAT] Remove HALLOWEEN_2024 feature flag

brian-man:


brian-man:
I think the test failures are unrelated to my changes:
- deliver › gives 100% more seasonal ticket on completed deliveries if double delivery is active
- deliver › can deliver items from the wardrobe

eliasSFL:
I think we have to check with @telk2 and @adamhannigan if they wants to release this portal to the public or not. If they want to release it, I might release it with #5199. For now I'll close this PR since it also has conflicts

PR #5228, opened 2025-02-08
fix: typos in documentation files

maximevtush:


src/lib/blockchain/abis/Withdrawals.ts
```diff
@@ -232,7 +232,7 @@ export default [
   },
   {
     inputs: [{ internalType: "uint256", name: "_minimum", type: "uint256" }],
-    name: "setTeamWithrawalMinimum",
+    name: "setTeamWithdrawalMinimum",
```
eliasSFL, line 235: Thanks for noting this typo out! Unfortunately this is also the same term that is used in our smart contracts. changing it will break the withdrawal contract

PR #5233, opened 2025-02-10
[FIX] Preload React.lazy routes

c-py:
# Description

Prevents errors loading `World` between release versions.

`World` is chunked as a dynamic import via `React.lazy`.  Seems like the default module preloader in vite ignores this chunk. This PR adds a plugin to preload these.

1. `yarn build && yarn vite preview --port 3000`
2. Check `World` is loaded in the developer console

# Checklist:


spencerdezartsmith (approved):
Working well. Confirming that World loaded.

PR #5244, opened 2025-02-11
[FEAT] Increase cooking oil capacity for queueing

brian-man:
The current capacity of "2 days" coupled with the current implementation of cooking oil based on base duration, the queueing system is difficult to use correctly with cooking oil.

This change increase the reservoir for cooking buildings to 5 days to support the following:
1. fill oil to max for building
2. queue 4 of any recipe you want
without having to repeat step 1 in the middle of step 2.


eliasSFL (approved):
LGTM!

PR #5276, opened 2025-02-18
[FIX] Remove Richie Introduction

c-py:
# Description

Richie has the same code as pumpkin Pete during introduction. And the condition to display is the same.

This PR deletes Richie code to prevent two Panels showing in game introduction.

Before:

richie

After:

pete


spencerdezartsmith (approved):
Richie never had a chance 😢

PR #5277, opened 2025-02-18
[FIX] Hide Fisherman Events for Unqualified Farms

Droidze:
# Description

There were some reports that players on Basic Island saw "fish frenzy" signs (like lightning icon, fish frenzy intro, etc.) on the fisherman NPC and related modal but didn't get the event boost after trying to catch the fish. This PR will update the code so that fishing events are not seen there if the farm is not qualified yet.

Fixes #issue

- On the basic island, when fishing events like Fish Frenzy happen, try to click the fisherman and catch a fish.

# Checklist:


eliasSFL (approved):
LGTM!

PR #5279, opened 2025-02-19
[FEAT] Ronin Login

c-py:
# Description

Features:

- Removes `switchWallet` from sign in page to support multiple chains.
- Introduces`NON_POLYGON_ACTIONS` into the wallet machine which are actions that are supported on all `EVM` chains.
- Adds Ronin Browser Wallet, Ronin Mobile Wallet and Ronin Waypoint to the Login Screen.
- Reintroduces an extra state in the walletMachine for browser wallets that require a `click` event in order to open a new window (Sequence and Waypoint).

- Test logging in with Ronin. (Unfortunately the wallet wall cannot add amoy to ronin wallets, so that fails, however ronin browser wallet does support mainnet).

- Use https://sunflower-land.com/ronin to test the PWA Login

# Checklist:


src/features/auth/components/SignIn.tsx
```diff
@@ -252,15 +345,51 @@ const PWAWallets: React.FC<Props> = ({ onConnect }) => {
...
   const { connectors } = useConnect();
 
+  const walletConnectConnector = connectors.filter(
+    ({ id }) => id === "walletConnect",
+  )[0];
+
+  const [provider, setProvider] = useState<any>(null);
+
+  useEffect(() => {
+    (async () => {
```
spencerdezartsmith, line 368: Never thought to use an IIFE in this case

spencerdezartsmith (commented):
Browser + Extension ✅ 
Browser + Waypoint 👎🏼 - I can just continue to select this and fail repeat....
Screenshot 2025-02-20 at 1 31 56 PM
Browser + Mobile 👎🏼 `Proposed chains are not supported` - This is expected I think?

**PWA** - Couldn't install. Will chat directly with you about this.

spencerdezartsmith (approved):
Retested and working well!

PR #5294, opened 2025-02-21
[FIX] Force all contracts to use polygon

c-py:
# Description

Forces all contracts to use polygon as a defensive mechanism, to prevent calls going to other chains where contracts are not deployed.

# Checklist:


spencerdezartsmith (approved):
LGTM

PR #5300, opened 2025-02-22
[FIX] Recalculation of Queue Recipes

spencerdezartsmith:
# Description

Players reported that they were getting some inconsistent cooking times after cancelling, insta cooking or using instant gratification skill. 

I have fixed this bug in all of these cases and created a single function to recalculate the queue after these actions. I have also fixed the queue so the order is correct.

Fixes #issue

- Tear this feature apart, cook, cancel, cook, insta cook etc. 

# Checklist:


eliasSFL (approved):
LGTM!

PR #5316, opened 2025-02-25
[FIX] Expires At Time In VIP 

spencerdezartsmith:
# Description

The expires at time inside of the VIP screen was not considering the expire at time for Ronin NFT's. I have updated this.

- Shows the correct expiry time. 
- Shows the longest expiry time if a player has `vip` and also has an NFT or if a player has two passes. It will show the expiry of Platinum over Gold.
- Blocks the purchase of `vip` if the player still more than 1 day available on their Ronin VIP. This is because the purchased `vip` doesn't stack onto the Ronin VIP. We can't stack because the VIP is transferrable.
- Allows purchase of `vip` as normal for everyone else.

_Ronin VIP active_
Screenshot 2025-02-25 at 1 23 08 PM

Fixes #issue

All the above statements hold true.
You will need to test this on a Ronin account with an NFT in your wallet.

# Checklist:


eliasSFL (approved):
LGTM!

PR #5320, opened 2025-02-26
[FIX] Crafting box initial time

brian-man:


eliasSFL (approved):
LGTM! I've updated the props to pass in state instead of gameService

PR #5328, opened 2025-02-26
[FIX] Broken Offers

spencerdezartsmith:
# Description

This PR fixes an issue where `Make an Offer` was breaking for wearables. The reason being is we were not considering the collection type when checking if something was a resource. So `Banana Onesie` for example had the same id as `Soybean` so `isResource` was evaluating to true. This broke the game.

Fixes #issue

- Try to make an offer on a Banana Onesie

# Checklist:


c-py (approved):
LGTM!

PR #5346, opened 2025-03-02
[FEAT] VIP Ronin Cooldown

adamhannigan:
# Description

This PR adds a 24 hour cooldown before Ronin VIP Access is activated (after claiming the NFT).

This prevents multi-accounters sharing VIP across their farms.


spencerdezartsmith (dismissed):
LGTM

PR #5370, opened 2025-03-07
[FIX] Switch to polygon on collecting daily reward

c-py:
# Description

Puts an extra step to switch to polygon as a last resort when collecting daily chest

# Checklist:


spencerdezartsmith (approved):
LGTM 👑

PR #5413, opened 2025-03-18
[FEAT] Flower gems

adamhannigan:
# Description

This PR adds the ability to buy gems with FLOWER 💰 

**MATIC**

Please note there will only be 2 options - once the FLOWER option is live you can no longer purchase with MATIC.

## How to test?

1. Run your local API
2. Navigate to the Gem purchases, select your bundle and pay with FLOWER.

Screenshot 2025-03-18 at 7 57 35 PM
Screenshot 2025-03-18 at 7 57 49 PM
Screenshot 2025-03-18 at 7 57 54 PM
Screenshot 2025-03-18 at 7 58 21 PM


src/features/game/components/modal/components/BuyGems.tsx
```diff
@@ -120,6 +127,88 @@ export const BuyGems: React.FC<Props> = ({
     );
   }
 
+  if (!!price && showFlowerConfirm) {
+    const flowerPrice =
+      gameService.getSnapshot().context.prices.sfl?.usd ?? 0.0;
```
spencerdezartsmith, line 132: Is there a reason to mix these accessors :
`gameService.getSnapshot()`
`gameService.state.context.state.balance`

Neither of these subscribe to changes. The `useSelector` pattern which is used above is a better one to follow. Unless there is a reason for using this "point in time" value. 

Mixing the ways we access data *can* lead to unexpected behaviour. I guess so long as you use these with intention its fine.

spencerdezartsmith (dismissed):
LGTM. Just one non blocking question and one "wrapping" issue on mobile.

Screenshot 2025-03-19 at 9 30 40 AM

src/features/game/components/modal/components/BuyGems.tsx
adamhannigan: Thanks mate. I will patch those up.

I often use the gameService when re-renders are not necessary - I will make them consistent

PR #5414, opened 2025-03-18
[FIX] Daily Chest Level Requirement Bug

Droidze:
# Description

- Fix the issue where the daily chest is displayed to players without checking the minimum required level. If they haven't reached level 6, clicking on it results in a blank modal.
- Add a condition to show the dedicated tab for `DailyRewardContent` in `Rewards.tsx`, only if players have reached the required level.
- Fix the issue in `RewardsButton.tsx` that causes `bumpkinLevel` to return experience instead of level.

Fixes #issue

- Try to check the visibility of the Daily Chest with and without the required level.
- On the testnet, check the Rewards Button in the top left to verify the visibility of Daily Reward both with and without the required level.

# Checklist:


spencerdezartsmith (approved):
LGTM

PR #5433, opened 2025-03-23
[FIX] Fruitful Bounty Skill

Droidze:
# Description

Fixes an issue where Fruitful Bounty does not work as intended (i.e., applying the Fruitful Blend after planting the fruit patch plants does not grant the skill boost on the first harvest but instead applies it to subsequent harvests).

Fixes #issue

Apply Fruitful Blend to fruit patches both before and after planting a fruit, with and without the Fruitful Bounty skill.

# Checklist:


eliasSFL (commented):
Thanks for looking into this for me. Now I see where the code went wrong for this skill. I'll create a BE PR to implement your changes

PR #5439, opened 2025-03-25
[FEAT] Remove VIP Offer

adamhannigan:
# Description

Removes the middle tier VIP offer (20% off). This was only meant to last the first 2 weeks, but we extended it due to the large intake of Ronin players.


eliasSFL (changes requested):
in VIPItems.tsx
You may want to remove the following:

line 286
```tsx
                {name === "3_MONTHS" && (
                  <>
                    <Label type="info" className="absolute -top-5 -right-2">
                      {t("vip.3Months.discount")}
                    </Label>
                    <p className="text-xs absolute bottom-5 line-through right-0">
                      {`1500`}
                    </p>
                  </>
                )}
```

eliasSFL (approved):
LGTM!

PR #5443, opened 2025-03-26
[FIX] Add to queue button not disabled when queue is full

dwi:
# Description

`isQueueFull` was not considering the current selected recipe being cooked so when queue was full the condition was always `3 >= 4` and never disabled the button.

Before:
image

After:
image

Fill the VIP queue and see the "Add to queue" button being disabled.

# Checklist:


c-py (approved):
Thank you! @dwi 🚀

PR #5444, opened 2025-03-26
[FIX] EXPANSION_ORIGINS for tiles 26+

brian-man:
Game crashes after expanding tile 26 on Volcano island.

The issue is that `ExpandIcon` is passed `EXPANSION_ORIGINS[26]` (missing) for `position` which fails here:
```
      <MapPlacement
        x={position.x - LAND_SIZE / 2}
```


adamhannigan (approved):
Thanks for catching this 🙏

PR #5491, opened 2025-04-06
[FIX] Coin Balance Overlap with New Love Charm Icon Button

Droidze:
# Description

- This PR fixes an issue where the coin balance could be overlapped by the newly added Love Charm icon button on smaller screens. The coin balance has been repositioned to ensure proper visibility across all screen sizes.

| Before | After |
|--------|--------|
| image | image | 

- Minor UI FIxes

Fixes #issue

Check the top-right corner of your game's screen to see how the balances for gems, flowers, and coins are displayed across different screen sizes. Click on those that have click events.

# Checklist:

### DIFF
--- a/src/components/Balances.tsx
+++ b/src/components/Balances.tsx
@@ -26,26 +26,15 @@ export const Balances: React.FC<Props> = ({ sfl, coins, gems, onClick }) => {
           className="flex cursor-pointer items-center space-x-3 relative"
           onClick={onClick}
         >
-          <div className="h-9 w-full bg-black opacity-30 absolute coins-bb-hud-backdrop" />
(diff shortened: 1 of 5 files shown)


spencerdezartsmith:
Hey @Droidze I like this but can we limit it to only stack them like this when the screen is very small? I think the current design is better in general but agree that it should be flexible so that it doesn't get blocked.

Droidze:
@spencerdezartsmith I’ve updated the code so that it now only stacks them on mobile screens, while on larger screens it remains the same as it currently appears on the mainnet.
| Mobile Phones | Larger Screens |
|--------|--------|
| image | image |

PR #5495, opened 2025-04-07
[FEAT] Facial recognition for unverified farms

adamhannigan:
# Description

Enabled facial recognition on farms that haven't yet been verified (including old farms that were not verified)


eliasSFL (approved):
LGTM!

PR #5531, opened 2025-04-19
[FEAT] Easter Event

telk2:
# Description

This PR adds the Easter Island for the easter event. In this island players can find a themed portal where they earn event tokens, a shop to spend those tokens on special items and also a donation NPC with the names of all contributors that helped to create the event.

### (Some stuff that I couldn't do because of time constraints and/or tried but could not make it work)
If possible add limits to the items on the shop (for the limited items section) and lower their price by 50%.

Also if possible, make the carrot pitchfork, the bunny mask and the white tunnel bunny items that players can buy spending 15 FLOWER. 

### TO-DO
The metadata generation and dictionary entries have also to be done. I tried generating the metadata  and it did not worked for me.


public/world/event_island_assets/pablo.png
adamhannigan, line 1: Pablo returns! 😍

adamhannigan (approved):
Great work! 👏 

I will work on the backend events + tweaks as requested. For now I will merge this in so we don't get blocked 🚀 

A few suggestions & feedback:

- Can we put a label above the portal? It wasn't clear I needed to click on the rabbit to start the portal
- The Portal entry is missing a panel (it has a transparent background)
- Is it possible to give players more than 1 free attempt? The first attempt is usually just understanding controls. I think 2-3 would get more people engaged and lead to more people paying to play more 👍

PR #5544, opened 2025-04-22
[FIX] Leaderboard Dates

telk2:
# Description

Changed leaderboard dates to better reflect the scores without exploited runs. The exloit was already fixed.


adamhannigan (approved):
LGTM

PR #5598, opened 2025-05-02
[FEAT] Save improvements

adamhannigan:
# Description

Reduce the amounts of time saved is called.

- Increase interval 30s > 60s
- No save when opening map
- No session reload when going from farm > world (moved GameProvider to Navigation)

## How to test?

Run connected to dev API.

Travel between your island + plaza (should be instant) 🏃


spencerdezartsmith (approved):
Nice job 👏🏼 

Working well

PR #5633, opened 2025-05-09
[FIX] Prevent leaving server on scene change

c-py:
# Description

Only leaves the server if the mmo is unmounted.


adamhannigan (approved):
LGTM

PR #5637, opened 2025-05-11
[FIX] Update Love Charm price for Super Totem

brian-man:


eliasSFL (approved):
LGTM! I double checked with the BE and the cost for the super totem matches what you have here

PR #5641, opened 2025-05-12
[FEAT] Increase Level Cap to 150

brian-man:
- Added new levels for lvl 126-150
- Cleaned up the old pickSkill event from the former skills


igs-raiderler:
Interesting solution to this problem! I've been working on similar challenges recently.

c-py (approved):
LGTM!

PR #5667, opened 2025-05-18
[FIX] Quarry removable criteria

brian-man:


eliasSFL (approved):
LGTM! Just checked the BE, seems like it's only the FE that has this discrepancy

PR #5695, opened 2025-05-26
[FEAT] Add New Deposit UI

spencerdezartsmith:
# Description

Split deposit UI to allow for:

- Easy deposit from Linked Wallet
- Ability to add from another wallet (exchange etc)

Screenshot 2025-05-26 at 2 39 13 PM

**NOTE: The FlowerOFT contract is not verified on testnet so you will get a random ? in metamask when the transaction pop up opens. This will not be the case on mainnet**

Fixes #issue

- Deposit flower using the new UI

# Checklist:

- [x] Screenshot if it includes any UI changesles


eliasSFL (approved):
LGTM! The balance now updates after the deposit is handled

PR #5702, opened 2025-05-28
[FIX] Prevent Wallet Connections Fighting

c-py:
# Description

This PR defers loading of wallet connections until the wallet is selected on the wallet wall. 

This helps to prevent two wallets from being active at the same time and conflicting. This is most common with WalletConnect + Metamask, Coinbase + Metamask and Ronin waypoint + Metamask.

Do an on chain action to ensure the wallet wall still works. For example claim the daily chest.

# Checklist:


spencerdezartsmith (approved):
LGTM

PR #5782, opened 2025-06-24
[FIX] Event Portal linking to wrong URL

telk2:
# Description

This will fix the portal minigame for the next event and enable beta testers to test it before the event starts.


adamhannigan (approved):
LGTM

PR #5792, opened 2025-06-25
[FIX] Add abort controller to autosaving request

c-py:
# Description

Adds an abort controller to the autosaving request, in case fetch never resolves (possibly in bad network conditions). This throws an unhanded exception as we don't know if the server received the request or not - and if the autosave failed or not. This should only happen in bad network conditions.

Manual inspection

# Checklist:


spencerdezartsmith (approved):
LGTM!

PR #5816, opened 2025-06-30
[FIX] Replace wallet drop down portal with popper

c-py:
# Description

Updates the wallet drop down to use popper instead of a portal.

These drop downs were not working due to being portaled outside of the modal, triggering the modal onHide instead of changing the chain.

1. Checkout `main`
2. Verify you cannot change chain on the daily chest
3. Checkout this branch
4. Verify you can change chain

# Checklist:


eliasSFL (approved):
LGTM!

PR #5819, opened 2025-06-30
[FIX] Event Shop Wings Price showing incorrect value

telk2:
This PR makes it so the value for the slime wings on the event shop shows their right price.  It does not increase it, since it already is charging 150 FLOWER for it(for VIPs which have 50% discount).


eliasSFL (approved):
I have refactored your festivalOfColours code to use the consistent data from FESTIVAL_OF_COLORS_STORE

PR #5838, opened 2025-07-07
Making the Festival of Colors event last for one more day

telk2:
This changes the feature flag for the festival of colors event for it to last one more day since there were some problems where players could not buy the attempts at the portal on the first hours of the event


c-py (approved):
LGTM!

PR #5845, opened 2025-07-09
[FEAT] Wire Up Chat

spencerdezartsmith:
# Description

WIP: Chat interactivity is wired up. Does not currently work on mobile as I am doing a refactor on the next ticket.

# Checklist:

### DIFF
--- a/src/features/social/PlayerModal.tsx
+++ b/src/features/social/PlayerModal.tsx
@@ -22,50 +22,16 @@ import { SUNNYSIDE } from "assets/sunnyside";
 import { FollowerFeed } from "./components/FollowerFeed";
 import { IslandType } from "features/game/types/game";
 import { useTranslation } from "react-i18next";
-import useSWR from "swr";
+import useSWR, { KeyedMutator } from "swr";
 import { getPlayer } from "./actions/getPlayer";
 import { Context } from "features/game/GameProvider";
 import { useSelector } from "@xstate/react";
 import { MachineState } from "features/game/lib/gameMachine";
 import { postEffect } from "features/game/actions/effect";
 import { randomID } from "lib/utils/random";
-
-export type FarmInteraction = {
-  id: string;
-  sender?: string;
-  timestamp: number;
-  text: string;
-  type: "comment" | "action" | "milestone" | "announcement";
-};
-
-export const dummyInteractions: FarmInteraction[] = [
-  {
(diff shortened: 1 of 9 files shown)


src/features/social/components/ChatInput.tsx
```diff
@@ -75,18 +60,7 @@ export const ChatInput: React.FC<Props> = ({ disabled, onEnter }) => {
...
-            id: `${Date.now()}`,
-            text,
-            sender: username,
-            type: "comment",
-            timestamp: Date.now(),
-          })
-        }
-        disabled={text.length === 0}
-      >
+      <Button onClick={() => onEnter(text)} disabled={text.length === 0}>
```
c-py:
```suggestion
      <Button onClick={() => onEnter(text)} disabled={        disabled={text.length === 0 || !isValidText()}
}>
```

This button might need to be disabled if is not validText

Image

src/features/social/PlayerModal.tsx
```diff
@@ -190,18 +172,46 @@ export const PlayerDetails: React.FC<Props> = ({ player }) => {
...
+    };
+
+    // Optimistically update the messages
+    mutate(
+      (current) =>
+        mergePlayerData(current!, {
+          messages: [...(current?.data?.messages ?? []), newMessage],
+        }),
+      { revalidate: false },
+    );
```
c-py:
```suggestion
mutate(
      async (current) => {
        const { data: response } = await postEffect({
          effect: {
            type: "message.sent",
            recipientId: player.farmId,
            message,
          },
          transactionId: randomID(),
          token: token as string,
          farmId: farmId,
        });

        return mergePlayerData(current!, response);
      },
      {
        optimisticData: (_, current) =>
          mergePlayerData(current!, {
            messages: [...(current?.data?.messages ?? []), newMessage],
          }),
      },
    );
```

I sent multiple messages, and found that sometimes data endpoint being queried while messages were still in flight. The state gets a bit confused about what the `current` value is.

Looks like mutate can take an async function, and it will block the refresh while the mutation is in progress.

c-py (dismissed):
LGTM! Just a couple of comments below!

src/features/social/components/ChatInput.tsx
spencerdezartsmith: Looks good 😅

PR #5857, opened 2025-07-15
[FIX] Zucchini spelling

brian-man:
image
The title spelling is correct.  The boost is incorrect.


jcads (dismissed):

eliasSFL (changes requested):
Please remove the changes in all the dictionary files as the translate script wont work if the terms are the same. Just edit dictionary.json is enough

brian-man:
> Please remove the changes in all the dictionary files as the translate script wont work if the terms are the same. Just edit dictionary.json is enough

Why does the translate script need to be involved in this case at all?

adamhannigan:
@brian-man Apologies for the delay - it looks like this got caught up in the Chapter launch work.

We only need to update the `dictionary.json` file when adding, removing or changing text. Our CI/CD pipeline produces all of the en.json, fr.json, es.json etc automatically.

We manually change the `fr.json`, `es.json` files directly when we manually want to override one of the automatic translations.

What you have done won't cause any issues. It is redundant work that we call out in PRs only to standardise the process = when a language file is changed we know it is a manual translation override and can focus on that.

PR #5877, opened 2025-07-24
[FEAT] Remove crop locking AOE restriction

c-py:
# Description

- Persists the AOE used timestamp in `gameState.aoe`. This is used to ensure the AOE bonus can only be used a fair amount of time, when moving crops, stones, iron and gold around.

# How to Test

**Test 1**
1. Run FE & BE
2. plant something within range of "Basic Scarecrow"
3. Move the crop out and move another plot in
4. Ensure the AOE does not get applied on the second crop

**Test 2**

1. Plant two crops
2. Move crop into Sir Goldensnout
3. Harvest
4. Move second crop into same Sir Goldensnout
5. Harvest

# Checklist:


eliasSFL (approved):
LGTM!

PR #5934, opened 2025-07-31
[FEAT] Allow Follow From Visit

spencerdezartsmith:
# Description

This PR allows a player to follow another player while visiting them. If a logged in player visits another player and clicks on their bumpkin, they are now able to click the "follow" button and follow them. 

The "follow" effect does not go through the game machine so it doesn't interfere with visitor context changes. 

I have also changed some naming from `farmId`
-> `loggedInFarmId` just so it is clear what we're referencing as the value could either be `farmId` or `visitorId` depending on whether they're visiting or not.

I have also changed the world button in the visitor hud to a back button so its clear that that button takes you back.

Screenshot 2025-07-31 at 10 50 29 PM

- Visit a player either directly from the url or via a bumpkin in the mmo.
- Click on their little bumpkin by the house
- Click follow
- Click unfollow
- Click around back and forth, following and unfollowing players while visiting just to ensure everything flows as its supposed to

# Checklist:

### DIFF
--- a/src/features/game/GameProvider.tsx
+++ b/src/features/game/GameProvider.tsx
@@ -10,8 +10,7 @@ import {
   cacheShortcuts,
   getShortcuts,
 } from "features/farming/hud/lib/shortcuts";
-
-import { startGame, MachineInterpreter, MachineState } from "./lib/gameMachine";
+import { startGame, MachineInterpreter } from "./lib/gameMachine";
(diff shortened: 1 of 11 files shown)


c-py (approved):
LGTM! Worked well!

There is a small bug when you send the player visiting a message.

It temporarily pops up with the visited farm NPC before updating to visited.

PR #5988, opened 2025-08-04
[FIX] Duplicating following data on infinite load fetch in feed

spencerdezartsmith:
# Description

Every time the feed fetched more data it was duplicating the following list.

This PR fixes the issue. We now just take the following data from the first fetch. If a follow happens in between a mutate is called which refreshes the list which updates the first fetch correctly.

I have also made it that the feed closes if you click outside of it.

Fixes #issue

- Scroll to the bottom of the feed to trigger a refetch. 
- Confirm the following list doesn't duplicate

- Open the feed, click outside to close it.

# Checklist:


c-py (approved):
LGTM!

PR #6043, opened 2025-08-06
[FIX] Digby Dialogue Condition

Droidze:
# Description

Fixes the issue where the following dialogue appears when all 3 artefacts of the day are found and the reward has already been claimed. With this update, the dialogue no longer appears once the reward is claimed.

image


eliasSFL (approved):
LGTM!

PR #6045, opened 2025-08-06
[FEA] Carry-over unused bin space

jcads:
# Description

Carry over unused bin storage to the next day

- Pickup clutter until bin is full
- Increase bin limit
- Use only a portion of the increased limit
- The next day, limit should be `TRASH_BIN_LIMIT` + unused limit from before


eliasSFL (approved):
LGTM!

eliasSFL (approved):

PR #6062, opened 2025-08-08
[FEAT] Differentiate Follow Requests by List Type

spencerdezartsmith:
# Description

Add `networkType` to the followers requests so we can return only followers or followedby. This is required for the cursor pagination to work correctly.

Fixes #issue

- Check followers and following lists work correctly and are ordered by social point.

# Checklist:


c-py (approved):
LGTM! ✅

PR #6110, opened 2025-08-13
[FEAT] Hash response

adamhannigan:
# Description

This PR includes an object hashing approach to minimise API responses. Instead of returning the entire Gamestate on response, the API will only return fields that have changed.

Before sending a request to `autosave`, we calculate a hash of each field in the game state and send this up to the API.

An example of this looks like:

```
{
    inventory: "sha256:934779132aa43dd98052bcbc37e457912d5014ead70dae11039a5537864d9cd6"
    airdrops: "sha256:4f53cda18c2baa0c0354bb5f9a3ecbe5ed12ab4d8e11ba873c2f11161202b945"
    // ... every other field on the game state
}
```

When the API responds, it will only return fields where the hash does not match (i.e. something has changed)

Screenshot 2025-08-13 at 11 26 17 AM

We then merge the payloads together

**Excludes**

- Excludes more optimal hashing algorithms. 
- Excludes deep field checking and key sorting
- Excludes other endpoints


c-py (approved):
LGTM!

PR #6122, opened 2025-08-14
[CHORE] Check Progress Tests

spencerdezartsmith:
# Description

I have removed the "resources" type from `checkProgress` and updated it to use `getCollectionName`. I have also added tests for `checkProgress`.

Fixes #issue

- Test the game works
- Test it against a listing without a `collection`

# Checklist:

### DIFF
--- a/src/features/marketplace/lib/getCollectionName.ts
+++ b/src/features/marketplace/lib/getCollectionName.ts
@@ -1,6 +1,5 @@
 import { KNOWN_IDS } from "features/game/types";
-import { BumpkinItem, ITEM_IDS } from "features/game/types/bumpkin";
-import { InventoryItemName } from "features/game/types/game";
+import { ITEM_IDS } from "features/game/types/bumpkin";
 import {
   MarketplaceTradeableName,
   CollectionName,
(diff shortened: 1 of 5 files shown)


src/features/marketplace/lib/getCollectionName.ts
```diff
@@ -0,0 +1,25 @@
...
+export function getCollectionName(
+  itemName: MarketplaceTradeableName,
+): Exclude<CollectionName, "resources"> {
+  if ((itemName as InventoryItemName) in KNOWN_IDS) {
+    return "collectibles";
+  }
+
+  if ((itemName as BumpkinItem) in ITEM_IDS) {
+    return "wearables";
+  }
```
eliasSFL: I think we can do away with type casting here as it's not really causing any errors with Typescript.

```suggestion
  if (itemName in KNOWN_IDS) {
    return "collectibles";
  }

  if (itemName in ITEM_IDS) {
    return "wearables";
  }
```

eliasSFL (dismissed):
LGTM! 
Just one non-blocking comment regarding type casting

PR #6136, opened 2025-08-15
[FEAT] Flower InstaGrow Confirmation Modal

Droidze:
# Description
Shows a Confirmation modal for Flower InstaGrow

image


eliasSFL (changes requested):
https://github.com/user-attachments/assets/38a455a7-a43e-4246-8441-425f7f5a835f

I think instead of a confirmation modal, we can show additional buttons instead of a separate component/modal

PR #6140, opened 2025-08-15
[FIX] Avoid duplicate display of Monuments in Chest

Droidze:
# Description
Prevent boost-type Monuments from appearing in both categories in Chest:
| Before | After |
|--------|--------|
| image | image |


eliasSFL (approved):
LGTM!

PR #6141, opened 2025-08-15
[FIX] Daily reward message for yearly streaks

Droidze:
# Description

Previously, the daily reward message for a 1-year streak was only shown when players reached their first 365-day streak. This update modifies the condition to display the message for every yearly streak milestone (e.g., 730 days, 1095 days, etc.).


eliasSFL (approved):
LGTM!

PR #6146, opened 2025-08-17
[FEAT] Shrine Improvements

Droidze:
# Description
**This PR makes the following updates to Shrines:**
- Exclude Shrines from “Remove/Remove All" actions (aligned with other expiring boosts)
---
- Show a confirmation modal when crafting Shrines to inform players of boost conditions:
Confirm_Craft

---
- Add a time-remaining label to the Obsidian Shrine (for other Shrines, it appears in the popover):
image

---
- Improve boost/expiry label descriptions and styles for consistency and accuracy
image


eliasSFL (changes requested):
I think that confirmation modal is a bit outdated. Besides the user still has a chance to cancel when they are forced into landscaping mode.

Image

 I think instead we can have a warning label here that tells the user that the boost is activated upon purchase

PR #6147, opened 2025-08-17
[FIX] Boosted Monuments Popover Condition

Droidze:
# Description
Fixes the issue shown in the image below, where the popover containing help progress appears when clicking a completed monument. After the fix, only the modal with the boost description will be shown.

image


eliasSFL (approved):
LGTM!

PR #6164, opened 2025-08-19
[FEAT] Display Withdraw Cooldown for Restricted Boost Items

Droidze:
- Display the withdraw cooldown for restricted items in a popover when clicked.
- Show a lock icon on restricted boosted items.
- Sort wearables and collectibles by longest cooldown first, followed by boosted items without a withdrawal cooldown, and then cosmetic/decoration items. For Buds, display the item with the longest cooldown first.

https://github.com/user-attachments/assets/a2f8bda4-3c5c-441a-8d74-456ae0fcf6fd


eliasSFL (changes requested):
Image

Do you think you can add a similar message for items that needs to be stored on chain before withdrawals?

PR #6166, opened 2025-08-20
[FEAT] Stone Forging

jcads:
# Description 
This PR only implements stone upgrades, we will iterate based on feedback. We will add more resources as we go

## Upgrade (merge) Stone in the Workbench

Requirements:
L2 - Fused Stone Rock

    4x Stone Rock
    1x Obsidian
    Requires 3x Pickaxe to mine

L3 - Reinforced Stone Rock

    4x Fused Stone Rock
    5x Obsidian
    Requires 13x Pickaxe to mine

**Shows the delta of the required tools in the tooltip _(Tier 3 stone requires 13x Pickaxe_)**

image

- Go to Workbench
- Upgrade Stone Rock to Fused Stone Rock
- Mine the upgraded stone, it should require less tools
- Verify boosts work, Place a Tin Turtle within AOE
- Verify Chores are credited properly
   - In the backend edit bert's chore to "Mine Stones 50 times"
    - **Tier 2 Stone = 4 stone mines**, **Tier 3 Stone = 16 stone mines**
- Verify landscaping with the upgraded stones still work


jcads:
**Forging and Mining upgraded stones**

https://github.com/user-attachments/assets/45151976-b0dc-4d2c-98f9-6e487a36abda

Xeko-92:
In your video, the stone node to the top left of the tin turtle wasn't impacted by the tin turtle.

image

Also, visual "error", the upgraded stone nodes while being mined display the original stone node.

image


I also wonder how a skill like Money tree with a % chance to get 200 coins when chopping wood would work. Would the % chance be increased or would the amount of coins be increased?

The Fused Stone Requires 3 pickaxes for 4 stone. The Refined Stone requires 13 pickaxes for 16 stone. Shouldn't it be 12 or less? The Refined one is actually a nerf from the Fused one.

I also think that the cost to upgrade these might be too cheap. It saves resources since you'll need fewer pickaxes, it saves space, and it makes the AoE boosts better since they can affect more nodes.

HansWonnie:
The fusion is quite interesting , but with the current state , the highest number of fusion for t2 would be around 5-6 for an high level player and with a t3 it would just be around 1, not really sure if a +1 pickaxe can compensate the space reduction, maybe reduce the cost to 12 or less as Xeko suggest since were paying more resources ( 5 obsidian )

Another test I would suggest would be to check how the native +1 (%) works , like if we consider mining 20 normal stone , the output would be around 20 - 25 stones, can you test if a L2 stone gives like 4.4 , 5.4 and 8.4 if the native skill actually works and also do the test to the L3 stone

brian-man:
This is pretty exciting.  This would definitely add some new strategy around Sunstone=>Rock spending as well.

markymartinhmiranda:
oooh love the cost change, this is pretty exciting! Now just wondering how this will work with instant recover stat like TTA and Sharkie Pickaxe. Great update team!

raulmonoanivia:
Those are still free to mine with Quarry?

PR #6178, opened 2025-08-22
[FIX] Team work monument

adamhannigan:
# Description

Fixes the issue where Team Work Monument was not increasing it's helped amount.

**How to tests?**

1. Place a teamwork monument
2. Get someone else to cheers it

Also worth testing digging up other monuments, placing them again and helping.

### DIFF
--- a/src/features/game/events/landExpansion/placeCollectible.test.ts
+++ b/src/features/game/events/landExpansion/placeCollectible.test.ts
@@ -135,6 +135,35 @@ describe("Place Collectible", () => {
     });
   });
 
+  it("adds monument to village projects", () => {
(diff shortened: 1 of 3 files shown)


spencerdezartsmith (changes requested):
I have tested this both placing a teamwork monument and also digging up and replacing a monument and I can't interact with either once placed. I am getting this error: 
Screenshot 2025-08-25 at 10 49 41 AM

adamhannigan:
@eliasSFL Looks like it is a cross-over from the helped today changes. Since these branches didn't have them yet and now they do. I've tested after rebase and appears to be working

PR #6179, opened 2025-08-22
[FIX] Visitor Guide Pending Projects

Droidze:
# Description

Replaces WORKBENCH_MONUMENTS with villageProjects to include Teamwork Monument as a pending project when conditions are met.


spencerdezartsmith (approved):
LGTM

PR #6182, opened 2025-08-23
[FIX] Allow pet shrines to be burned with correct condition

Droidze:
# Description

Burning pet shrines throws "Cannot burn..." error. this PR will correct the condition to prevent this error.


eliasSFL:
This requires BE as well. Will get to it on monday

eliasSFL (approved):
LGTM!

PR #6193, opened 2025-08-25
[FEAT] Increase Trade Limits for Resources

vitt0:
# Description

I increased the max trade limits of some resources following the increase in the Factions donations/Lava Pits requirements. I also increase the min listings of some resources that were too low to avoid having that many listings spamming the market.

# Checklist:


eliasSFL:
I'll need to make a corresponding PR on the BE as well. Once that's approved I'll approve this as well

PR #6205, opened 2025-08-27
[FIX] Missing Inventory after Selling Placed Treasure

c-py:
# Description

When selling treasure, that is placed, such as the Giant Apple, the inventory amount was incorrectly being calculated. The game removes more Giant Apples than the players had.

1. Checkout `main`
2. Place Giant Apple
3. Sell a different Giant Apple
4. See two apples disappear
5. Checkout the fix
6. Retest and assert only one apple is sold

# Checklist:


eliasSFL (approved):
LGTM!

PR #6230, opened 2025-09-03
[FEAT] Show Inventory Amount for Crafting Box Recipes

Droidze:
# Description

Displays the inventory amount for known patterns in the Recipes tab and for the selected recipe in the Craft tab. This helps players who want to complete tasks requiring certain amounts of crafting items or for other use cases.

image
image


eliasSFL:
image

Do you think the number here would confuse players that they might be getting that amount of dolls on craft? I feel like it would be better if there wasn't any number there

eliasSFL (commented):
LGTM!

eliasSFL (approved):
LGTM!

PR #6239, opened 2025-09-04
[FEAT] Default minigame leaderboard to past 7 days

adamhannigan:
# Description

Leaderboards can only be processed for the last 30 days. This PR removes the old from dates, defaulting the leaderboards to the past 7 days.


c-py (approved):
LGTM!

PR #6250, opened 2025-09-05
Fix/z index 2

spencerdezartsmith:
# Description

The z index of things was messed up during a fix for Tiles z index. Fixes tile and rug ordering.

This PR fixes that issue.

Its also fixes player search and social leaderboard opening in the mmo.

| Before | After |
|--------|--------|
|Screenshot 2025-09-05 at 3 34 00 PM|Screenshot 2025-09-05 at 3 28 33 PM|

Fixes #issue

- Just confirm that everything has rendered correctly. Popovers should be on top and also when harvesting, harvest amounts should not be rendering under neighbouring plots etc. Tiles should render under rugs and everything else.  
- Basically all should look and work right.
- Search already tested

# Checklist:


eliasSFL (approved):
LGTM!

PR #6283, opened 2025-09-12
[FIX] Weather Shop Buy Button Condition

Droidze:
# Description

Disables the buy button if the player does not have enough ingredients for the selected item.


eliasSFL (approved):
LGTM!

PR #6290, opened 2025-09-14
[FIX] Hoarder limits for cookables

brian-man:


eliasSFL:
Curious what's the context for this? Is it cuz the BE has hoard limits but FE doesnt prompt it?

brian-man:
> Curious what's the context for this? Is it cuz the BE has hoard limits but FE doesnt prompt it?

When collecting 1001st Mashed Potato, the FE doesn't not enforce the limit so you get BE error when trying to save.  It's not obvious to most players what's going on; game just feels broken.

spencerdezartsmith (approved):
LGTM

PR #6291, opened 2025-09-14
[FEAT] Display upcoming Animal Mutant

Droidze:
# Description
Displays the details for the mutant animal that will be produced on the next harvest. Helps players who might sell the animal before harvesting and lose the mutant.

https://github.com/user-attachments/assets/c682898d-da13-4fd0-8825-ec1a968dd4bc


eliasSFL:
@Droidze I think with mutants, we want the element of surprise at the point of harvest for those who don't tinker with the API as much, showing it before harvest would ruin the surprise element that we want to have here.

~~What's the bug that you're trying to address here? Players can't sell before harvest anyways so the player would have claimed the animal before selling it~~ Actually, I see what you mean now but I think there's a different solution that we can go by here instead of showing beforehand

PR #6295, opened 2025-09-15
[FEAT] No Marketplace Escrow

adamhannigan:
# Description

This PR introduces a new system for the marketplace where items are no longer escrowed. This will support:

1. Creating FLOWER offers without it being subtracted
2. Listing collectibles that are in-use on your land
3. Listing wearables that are equipped
4. Listing buds on your land

The goal is to make the marketplace simpler to use and increase liquidity of low-supply items. A few outcomes:

- Much less clicks. No longer need to dig up /dequip before trading
- Players can still benefit from buffs of listed items
- We can increase NFT liquidity (Lots of Pet NFTs available on sale on launch - instead of just using/hoarding)
- Players can put in multiple offers on different Bud/Pet NFTs at the same time to try get one
- Available trades increases across all items = lowers risk of marketplace funnelling

**Fulfilment**

1. If you have a collectible placed that is bought, it will automatically be removed from your land
2. If you have an equipped wearable that is bought, it will be removed from your Bumpkin
3. If you have a bud placed that is bought, it will automatically be removed from your land

### Cleared items

If you no longer have an item you listed or the FLOWER for an offer, the trade will be removed from the marketplace. Players will be notified with a popup.

Screenshot 2025-09-15 at 5 37 25 PM
Screenshot 2025-09-15 at 5 43 38 PM

## How to test?

Connect to the API and try the following flows:

1. You can list items that are placed on your land
2. Listed items disappear from your land once purchased
3. You can list equipped wearables
4. Listed wearables disappear from your bumpkin once purchased
5. You can list buds on your land
6. Listed buds disappear once purchased.
7. You can make an offer - FLOWER does not disappear.
8. You can accept offers with collectibles placed on your land
9. You can accept offers with equipped wearables
10. You can accept offers with placed buds

A few edge cases to check:

1. You cannot list more multiple of the same item (if you don't have enough)
2. If you sell a required equipped item, it will default to a basic item (i.e. Farm background)


src/features/bumpkins/components/BumpkinEquip.tsx
```diff
@@ -136,41 +136,7 @@ export const BumpkinEquip: React.FC<Props> = ({ equipment, onEquip }) => {
...
 
-  const warn =
-    isMissingHair ||
-    isMissingBody ||
-    isMissingShoes ||
-    isMissingShirt ||
-    isMissingBackground ||
-    isMissingPants;
-
   const { t } = useAppTranslation();
```
eliasSFL, line 139: We should continue to support the existing rules for bumpkin equip. Otherwise it would result in this nsfw issue:

image

It should still ensure that shirt & pants or dress is equipped before saving. same for the required wearables, they should be equipped before saving


Instead if a player sells something that is one of the required bumpkin items. it will fallback to the default that we set
```suggestion
  const warn =
    isMissingHair ||
    isMissingBody ||
    isMissingShoes ||
    isMissingShirt ||
    isMissingBackground ||
    isMissingPants;

  const { t } = useAppTranslation();
  const warning = () => {
    if (isMissingHair) {
      return t("equip.missingHair");
    }

    if (isMissingBody) {
      return t("equip.missingBody");
    }

    if (isMissingShoes) {
      return t("equip.missingShoes");
    }

    if (isMissingShirt) {
      return t("equip.missingShirt");
    }

    if (isMissingPants) {
      return t("equip.missingPants");
    }

    if (isMissingBackground) {
      return t("equip.missingBackground");
    }
    return "";
  };
```

src/features/game/events/landExpansion/equip.ts
```diff
@@ -53,22 +52,6 @@ export function assertEquipment({
     throw new Error("Cannot equip pants while wearing dress");
   }
```
eliasSFL, line 54:
```suggestion
  }

  if (!equipment.body) {
    throw new Error("Body is required");
  }

  if (!equipment.shoes) {
    throw new Error("Shoes are required");
  }

  if (!equipment.hair) {
    throw new Error("Hair is required");
  }

  if (!equipment.dress && !(equipment.shirt && equipment.pants)) {
    throw new Error("Bumpkin is naked!");
  }
```

adamhannigan:
@eliasSFL Thanks for the feedback 👌 

That NSFW should be solved by using the default items. Your `VITE_ANIMATION_URL` may be pointed at dev instead of local API which would show the old result (nude bumpkin)

Now that these items can be optional, we need to allow players to save if one of those items are missing.

It would be annoying for a player if they couldn't equip a new Hat, because they sold their only background. They would need to go and buy another background, just to place the hat.

I don't see a benefit anymore having 'required' items. We can make the experience a lot simpler by having them optional 👌

adamhannigan:
@eliasSFL I've added some specific 'underwear' fallback (instead of default red shirt/black pants) - so players know when they don't have something on.

Screenshot 2025-09-18 at 3 42 22 PM

adamhannigan:
Screenshot 2025-09-19 at 2 46 23 PM

Some updated underwear 👀

PR #6296, opened 2025-09-16
[FEAT] Withdraw Ronin

c-py:
# Description

Adds the ability to withdraw to Ronin.  This feature is flagged for testnet only.

roninwithdraw

1. Complete a withdraw on ronin
2. Check the flag enforces base only when `FLOWER_RONIN` feature flag is toggled (
Note: Test 2 will break on testnet  as`src/features/game/components/bank/components/Withdraw.tsx` enforces base chain. This can be updated to enforce `baseSepolia` for the test

```jsx
        <GameWallet
          action="withdrawFlower"
          enforceChainId={enforceBaseWithdrawal ? baseSepolia.id : undefined}
        >
```

# Checklist:

### DIFF
--- a/src/features/game/components/bank/components/WithdrawFlower.tsx
+++ b/src/features/game/components/bank/components/WithdrawFlower.tsx
@@ -86,7 +86,7 @@ export const WithdrawFlower: React.FC<Props> = ({ onWithdraw }) => {
         onBackdropClick={() => setShowConfirmation(false)}
       >
         <InnerPanel>
-          <div className="p-1">
+          <div className="p-1 mb-1">
             <div className="flex flex-col gap-1 mb-3">
               <Label type="default" icon={withdrawIcon}>
                 {t("withdraw.flower.confirm")}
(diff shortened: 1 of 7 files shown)


src/features/game/components/bank/components/WithdrawFlower.tsx
```diff
@@ -69,10 +76,58 @@ export const WithdrawFlower: React.FC<Props> = ({ onWithdraw }) => {
...
+    amount.greaterThan(balance) || amount.lessThanOrEqualTo(0);
 
   return (
     <>
+      <ModalOverlay
+        show={showConfirmation}
+        onBackdropClick={() => setShowConfirmation(false)}
+      >
+        <InnerPanel>
+          <div className="p-1">
```
spencerdezartsmith: This just gives a little space to break up the small text.

```suggestion
          <div className="p-1 mb-1">
```

spencerdezartsmith (changes requested):
I have tested the withdraw and it was successful. 

In regards to the flag, when I turn the flag to return false and head back to withdraw it still has `Saigon Testnet` selected and has no options available for me to change to. I guess if there is an enforced chain than it should force the change?

Screenshot 2025-09-16 at 2 00 23 PM

PR #6304, opened 2025-09-16
[FEAT] Remove old fields

adamhannigan:
# Description

Removes old fields from the Game state.


brian-man:
I sent a bunch of old stuff to Craig back when.  Like Kraken.  Lots to trim.

src/features/game/events/landExpansion/utils.ts
```diff
@@ -27,7 +27,8 @@ export const getFruitHarvests = (
 export const getSupportedChickens = (state: Readonly<GameState>) => {
```
eliasSFL: This function can be deleted since it's for the old chickens

eliasSFL (changes requested):
Some comments in the BE that should also be addressed in the FE

brian-man:
Especially for blacklisted farms, it would be good to do a pass on the backend to trim the old data from the farms as well.  This should shrink the daily snapshot size somewhat.

PR #6315, opened 2025-09-18
[FIX] Craft Tool Popover Issue

Droidze:
# Description

Fixes the following issue in the image, where the craft tool popover continues to show after moving the mouse away from certain nodes:
image

### DIFF
--- a/src/features/game/lib/landDataStatic.ts
+++ b/src/features/game/lib/landDataStatic.ts
@@ -49,7 +49,6 @@ export const STATIC_OFFLINE_FARM: GameState = {
     "Better Together Banner": new Decimal(1),
     Geniseed: new Decimal(400),
     Wheat: new Decimal(400),
-    Pickaxe: new Decimal(1),
     "Blue Tile": new Decimal(1000),
     "Beta Pass": new Decimal(1),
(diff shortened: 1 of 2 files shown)


c-py:
@Droidze do you have insights on how to replicate this issue on `main` by any chance? So that I can test the fix works correctly before and after?

Droidze:
> @Droidze do you have insights on how to replicate this issue on `main` by any chance? So that I can test the fix works correctly before and after?

@c-py It fixes the issue for tree, stone, iron, and gold nodes that were not upgraded. Additionally, the fix includes the upgraded versions of the mentioned nodes, except for the upgraded gold nodes, which already have this fix.

- ### To replicate it on desktop, the mouse should be moved quickly over the mentioned nodes when the required tools are not available:

https://github.com/user-attachments/assets/dfb5ee61-a705-4169-ba34-675aea0a9beb

https://github.com/user-attachments/assets/fbda4f13-2912-47bb-afb0-f9c5a21c3381

---
- ### For touch devices like mobile phones should click the node:

https://github.com/user-attachments/assets/9f92547c-72c4-487f-9dc1-d9e88f6eebb5

PR #6329, opened 2025-09-22
[FEAT] Less marketplace API calls

adamhannigan:
# Description

Every time the details for an item is opened, the API is hit 3 times with the same call.

The parent component was already calling the API so we don't need these child components calling them as well.

## How to test?

Run against dev API and visit an item. Everything should load correctly.


adamhannigan:
@c-py Good catch 👌 

Ive added the state machine listeners for this. When testing locally it does take a few seconds for them to load + disappear once bulk removed

PR #6335, opened 2025-09-24
[FEAT] Community API Key

adamhannigan:
# Description

A unique API key for players to hit our APIs.

We had suspicious usage - the purpose of this is to identify which tools/services are calling the API.


spencerdezartsmith (approved):
Working well.

PR #6337, opened 2025-09-24
[FIX] Hide Pest Net on restock screens

Droidze:
# Description

Pest Net is currently disabled, but it still appears on the restock screens (image below). This PR updates the code so that tools are only shown there if they are not disabled.

image


eliasSFL (commented):
Will need BE for this so I'll get back to you when the PR is up

PR #6341, opened 2025-09-25
[FEAT] Remove heartbeat from useSocial

c-py:
# Description

This PR removes the heartbeat from the `useSocial` hook. This is a performance optimisation, as the data in the heartbeat is already available directly from the API.

This hook is now only used for subscribing to real time events.

1. Message between two clients and ensure that messages show up in real time

# Checklist:


adamhannigan (approved):
Tested and messages were populating ✅

PR #6343, opened 2025-09-25
[FIX] Marketplace Search Performance

c-py:
# Description

Marketplace Search Performance is very laggy as it is calling `translate` too many times when fetching buff labels.

This PR narrows `COLLECTIBLE_BUFF_LABELS` functions, so that we only call `translate` on the items we are interested in, rather than all items.

Fixes #issue

1. Test marketplace search on `main` and then on this PR.

# Checklist:


spencerdezartsmith (dismissed):
Nice one. Just the one TSC error on the missed call.

PR #6344, opened 2025-09-25
[FIX] Place craftables

adamhannigan:
# Description


eliasSFL (approved):
I have updated the PR with the right fix.
Will approve and merge this in

PR #6357, opened 2025-10-07
[Chore] Zh_CN - Gam3 and Ronin Translation

Kanzuki-L:
Description
Chinese Localization updates

Fixes #issue
What needs to be tested by the reviewer?
code assistances if action fails


Kanzuki-L:
ya, there's a new update, already fix with that, thanks Elias.

PR #6387, opened 2025-10-16
[FIX] Remove stock from collectible.crafted

c-py:
# Description

`stock` existed in `collectible.crafted` for the `Immortal Pear` and other non-tradable SFTs. `Immortal Pear` is no longer in `stock` and instead we rely on the inventory limit to determine if the pear is purchasable or not.

This PR removes `stock` from `collectible.crafted` and enforces inventory limits instead.

1. Start Frontend and Backend
2. Buy a non-tradable SFT
3. Ensure you cannot buy another non-tradable SFT

# Checklist:


eliasSFL (approved):
LGTM!

PR #6405, opened 2025-10-21
[FIX] Pet Neglect doesn't trigger if pet has never been fed before

eliasSFL:
# Description

fedAt was not compulsory, hence in isPetNeglected, if it's not assigned, it defaulted to current time. This results in the pet being able to earn experience from petting and never have to be worried on whether a pet is neglected.

Fixes #issue

No manual testing could be done but just understand the code for isPetNeglected and how it worked before and after.

# Checklist:

### DIFF
--- a/src/features/game/events/claimPurchase.test.ts
+++ b/src/features/game/events/claimPurchase.test.ts
@@ -474,6 +474,7 @@ describe("purchase.claimed", () => {
               id: 1,
               requests: {
                 food: [],
+                fedAt: 0,
               },
               energy: 0,
               experience: 0,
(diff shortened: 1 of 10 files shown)


eliasSFL:
> **Findings**
> 
> * src/features/game/types/pets.ts:673 (also 711) – Now that `lastFedAt` comes directly from `pet.requests.fedAt`, any pet whose state arrives without that field (common for pets placed before this patch and never fed) will hit `new Date(undefined)`, which throws when `.toISOString()` is called. That breaks `isPetNeglected`/`isPetOfTypeFed`, so actions like fetch/feed will start failing for those players. Please keep a safe fallback or backfill `fedAt` when loading legacy pets before calling these helpers.

We will run a migration to assign fedAt

spencerdezartsmith (approved):
LGTM

PR #6455, opened 2025-10-31
[FEAT] Paw prints map

adamhannigan:
# Description

Includes the Chapter based map updates:

1. Pet themed Decorations around the Plaza
2. Auction Items

**Dynamic System**

Updated the code so it automatically rolls over + switches layers + switches banners

## How to test?

Update the `const chapter = ` lines in the PlazaScene to override which chapter we are in.

You should see new banners & decorations around the plaza.

The Auction items are permanent - we don't need to wait for Chapter change over. It will build good hype.


src/features/world/scenes/PlazaScene.ts
```diff
@@ -201,17 +234,17 @@ export class PlazaScene extends BaseScene {
...
 
     // Auction Items
-    this.load.image("groovy_gramophone", "world/groovy_gramophone.webp");
-    this.load.image("oil_gallon_npc", "world/oil_gallon_npc.webp");
-    this.load.image("giant_onion", "world/giant_onion.webp");
-    this.load.image("giant_turnip", "world/giant_turnip.webp");
+    this.load.image("pet_nft_egg", "world/pet_nft_egg.png");
+    this.load.image("pet_bed", "world/pet_bed.webp");
+    this.load.image("paw_prints_rug", "world/paw_prints_rug.webp");
+    this.load.image("giant_acorn", "world/giant_acorn.webp");
```
eliasSFL: Giant acorn is now a megastore item, might wanna switch it to something else instead

PR #6463, opened 2025-11-03
[FIX] Reverse the mutant IDs

adamhannigan:
# Description

These IDs were not matching the backend and causing the wrong images + marketplace trades to activate.


eliasSFL (approved):
LGTM!

PR #6483, opened 2025-11-04
[FIX] Resource calculation when expanding land

jcads:
# Description

Fix land reveal to count upgraded resource nodes using their base-equivalent multipliers, preventing expansions from “refunding” base Trees after upgrades

Upgrade Tree, Stone, Iron, and Gold nodes and then expand your land. It should not "refund" any resources after upgrades


eliasSFL:
Closing this cuz there's a newer version of this PR #6494

PR #6504, opened 2025-11-08
[FIX] Temporary Collectibles being deduced from inventory on upgrade

eliasSFL:
# Description

This PR Fixes a bug where if you have a tempoary collectible (hourglass,totem/shrine) it would deduct one from your inventory 

Video of bug:

https://github.com/user-attachments/assets/622a672d-db11-409e-90c9-7c06062cd71d

The fix is just to remove that function that was doing that previously.

## Some context as to why this bug is happening
The function that removes expireItems was there because previously there was a bug that returned an expired temporary collectible back into a player's inventory when a player upgraded their farm, since it would previously remove everything from their farm before placing new items from the new upgrade.

During the Landscaping update, there was a change with the upgradeFarm where we started using other functions in the upgradeFarm, one of them is the removeAll function. the removeAll function would remove all collectibles, buildings and buds, however it does not remove temporary collectibles anymore. Hence since temporary collectibles are not removed, there's no need to deduct from inventory anymore

Fixes #issue

# Checklist:

### DIFF
--- a/src/features/game/events/landExpansion/upgradeFarm.test.ts
+++ b/src/features/game/events/landExpansion/upgradeFarm.test.ts
@@ -929,47 +929,6 @@ describe("upgradeFarm", () => {
     expect(state.island.previousExpansions).toEqual(16);
   });
 
-  it("removes all temporary collectibles", () => {
-    const now = Date.now();
-
-    const state = upgrade({
-      action: {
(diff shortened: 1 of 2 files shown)


adamhannigan (approved):
LGTM!

PR #6554, opened 2025-11-17
[FEAT] React Compiler

spencerdezartsmith:
# Description

This PR installs [React Compiler](https://react.dev/learn/react-compiler/introduction) into the codebase and upgrades eslint to be able to pick up issues related to this install. The purpose is to help with our performance issues as it handles memoization at build time. 

There are four big issues that have been identified during this process based on the [Rules of React](https://react.dev/reference/rules) which need to be addressed as they break the compilers ability to apply memoization. I have addressed 1 and 4 in this PR. 

1. Our use of [refs](https://react.dev/reference/eslint-plugin-react-hooks/lints/refs) during render. I have addressed these issues in this PR.  
2. React components must be [pure](https://react.dev/reference/eslint-plugin-react-hooks/lints/purity) functions. I will address in a following PR. Our main issue here is using `Date.now` during the render process.
3. [set-state-in-effect](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect). We are big offenders of this one. I will also address as a seperate PR.
4. [static-components](https://react.dev/reference/eslint-plugin-react-hooks/lints/static-components). We have a lot of places where we do this. Basically defining a new component inside of another component.

This update highlighted a lot of area's where we break the Rules of React. Its a great opportunity to relearn these patterns and should set us up for a much healthier code base moving forward. 

I did see performance improvements running react-scan before and after. Before we had many hard dips in our fps when performing tasks like planting/harvesting, chopping trees etc. At times down to below 60fps. After this change is consistently was saying close to 120fps. I also am no longer seeing flashing rocks, trees 🤞🏼 

Fixes #issue

- Smoke testing of everything. Try to perform all actions, including opening treasure chests, digging, plant/harvest, chop, mine, cook etc.

# Checklist:


c-py (approved):
LGTM ✅ 

Game Events Tested: 
Start Day, Pet Reveal, Place Pet, Pick Mushroom, Plant Flower, Mine Stone, Chop Tree, Upgrade Well, Speed Up Well, Mix Food, Feed Chicken

Other Parts Tested:
Navigate to Plaza, Toggle Audio Menu

PR #6571, opened 2025-11-22
[CHORE] Hornet Mask description update

Xeko-92:
Change Hornet Mask description. This comes after seeing yet another question about how the Hornet Mask works. The description is misleading when you look at other items such as scarecrow. Scarecrow is +20% crop yield. This is multiplied by green amulet which is chance of 10x crops. Whereas with bee swarms, the Bee Collective skill which is +20% Bee Swarm chance isn't multiplied by Hornet Mask which is 2x Bee Swarm Chance.

image

image

image

image

image

# Checklist:

### DIFF
--- a/src/lib/i18n/dictionaries/dictionary.json
+++ b/src/lib/i18n/dictionaries/dictionary.json
@@ -499,7 +499,7 @@
   "bumpkinItemBuff.crimstone.hammer": "+2 Crimstones on 5th mine",
   "bumpkinItemBuff.crimstone.amulet": "-20% Crimstone Cooldown Time",
   "bumpkinItemBuff.crimstone.armor": "+0.1 Crimstones",
-  "bumpkinItemBuff.hornet.mask": "2x base Bee Swarm chance (+10% chance)",
+  "bumpkinItemBuff.hornet.mask": "2x base Bee Swarm chance",
   "bumpkinItemBuff.honeycomb.shield": "+1 Honey per Full Beehive",
   "bumpkinItemBuff.flower.crown": "-50% Flower Growth Time",
   "bumpkinItemBuff.goblin.armor": "+20% Marks",


brian-man:
So this basically bakes in officially that the boost was sold with deceptive labelling?

Xeko-92:
> So this basically bakes in officially that the boost was sold with deceptive labelling?

Well, technically it is 2x I believe. The problem is that 
A) The formuals aren't public, so you'd have to know the base chance is 10% to know what 2x actually means
B) All other boosts are applied afterwards, so 2x and +10% effectively mean the same thing in this case. It's much more clear if it just says +10%.

brian-man:
> > So this basically bakes in officially that the boost was sold with deceptive labelling?
> 
> Well, technically it is 2x I believe. The problem is that A) The formuals aren't public, so you'd have to know the base chance is 10% to know what 2x actually means B) All other boosts are applied afterwards, so 2x and +10% effectively mean the same thing in this case. It's much more clear if it just says +10%.

That's one perspective.

Another is that the boost should be honored as stated at time of sale.

For example, if the base ever changes to 15% then the +10% is less than the "double" the item was sold with.

Xeko-92:
> That's one perspective.
> 
> Another is that the boost should be honored as stated at time of sale.
> 
> For example, if the base ever changes to 15% then the +10% is less than the "double" the item was sold with.

My PR is to just change the description. If the base ever changes, the description of this item could change alongside that.


The problem with the current description is, let's take some other items for example such as scarecrow. Scarecrow is +20% crop yield. This is multiplied by green amulet which is chance of 10x crops.

Whereas with bee swams, the Bee Collective skill which is +20% Bee Swarm chance isn't multiplied by Hornet Mask which is 2x Bee Swarm Chance.

Another alternative would be to change the description to be "2x **base** Bee Swarm chance"

brian-man:
> My PR is to just change the description. If the base ever changes, the description of this item could change alongside that.

We're more likely to spontaneously get a 2nd sun in our solar system.

Again, it's never okay to change a boost after its sold IMO.  You're promoting and, in this case, advancing bad behavior.

Xeko-92:
> It's not really +10% because it multiplies from base chance before doing the additive ones. if we ever add new multiplicative chances, it would multiply from base and hornet mask before going to the others

But saying 2x is also inconsistent like I said with things like the Green amulet and items like scarecrow or Easter bunny. So either the hornet mask description needs to change (2x base Bee Swarm chance (+10% chance)) or all of the other bee swarm chances need to change.

brian-man:
> > It's not really +10% because it multiplies from base chance before doing the additive ones. if we ever add new multiplicative chances, it would multiply from base and hornet mask before going to the others
> 
> But saying 2x is also inconsistent like I said with things like the Green amulet and items like scarecrow or Easter bunny. So either the hornet mask description needs to change (2x base Bee Swarm chance (+10% chance)) or all of the other bee swarm chances need to change.

Then the proper adjustment would be "2x base chance" so when base chance changes the item doesn't get nerfed relative to what was sold.

Xeko-92:
> > > It's not really +10% because it multiplies from base chance before doing the additive ones. if we ever add new multiplicative chances, it would multiply from base and hornet mask before going to the others
> > 
> > 
> > But saying 2x is also inconsistent like I said with things like the Green amulet and items like scarecrow or Easter bunny. So either the hornet mask description needs to change (2x base Bee Swarm chance (+10% chance)) or all of the other bee swarm chances need to change.
> 
> Then the proper adjustment would be "2x base chance" so when base chance changes the item doesn't get nerfed relative to what was sold.

That's what I have currently. I added the +10% in brackets so that people know what the base chance is because afaik, there isn't anything in the game that indicates what it is.

PR #6575, opened 2025-11-24
[FEAT] React Compiler

spencerdezartsmith:
# React Compiler Integration & React Rules Compliance

This PR integrates the React Compiler into our codebase and refactors existing code to follow React’s core rules more strictly. The compiler assumes components are pure and follow the rules of hooks; when they do, React can safely optimize and we avoid a class of subtle bugs.

## What Changed

We addressed four main categories of React rule violations:

1. Impure Components — calling time- or randomness-based functions during render
2. Reading Mutable Values in Render — especially ref.current
3. Unnecessary Derived State from Props — mirroring props or external values into state
4. Components Defined Inside Components — nested component definitions

These changes are mostly mechanical but important. The goal of this PR is to set clear patterns for future code.

At a high level:

- React Compiler is wired in via `babel-plugin-react-compiler` in `vite.config.ts` (no new commands are required — it runs as part of the normal dev/prod build).
- Time and randomness access have been pushed into hooks or into domain / event functions instead of being called in React render.
- Several domain APIs now accept an optional `createdAt` / `now` / `cookStartAt` argument (defaulting to `Date.now()`), so tests and game logic can stay explicit about time while UI components remain pure.
- A number of effects/components were simplified to avoid mirroring props into state or defining components inside components.

---

### The Problem

React components must be pure functions — meaning:

- Same inputs → same outputs
- No side effects during render
- No unstable values like Date.now() or Math.random() in render

Violating purity prevents React Compiler optimizations, causes unnecessary re-renders, and can produce visual flicker.

### The Solution

We created hooks to provide stable or controlled values:

```ts
// Stable random number (chosen once per mount)
const random = useMathRandom();

// Stable random item (index chosen once per mount)
const item = useRandomItem(["option1", "option2", "option3"]);

// Time values
const now = useNow(); // Snapshot at mount
const now = useNow({ live: true }); // Live updating
const now = useNow({ live: true, autoEndAt }); // Live until a target time
```

Key properties of these hooks:

- `useMathRandom()` and `useRandomItem()` only call `Math.random()` inside a lazy `useState` initializer, so the value is picked once when the component mounts and then stays stable across re-renders.
- `useNow()` hides `Date.now()` and interval logic inside the hook. Components just receive a number (ms since epoch) and remain pure.
- `useCountdown(targetDate)` hides both `Date.now()` and the `setInterval` loop; components receive a stable object describing the countdown.

### Examples

Before:

```ts
// ❌ Impure — Date.now() in render
export const GoldEgg = ({ readyAt }) => {
  const isReady = readyAt <= Date.now();
};
```

After:

```ts
// ✅ Pure — time from hook
export const GoldEgg = ({ readyAt }) => {
  const now = useNow({ live: true, autoEndAt: readyAt });
  const isReady = readyAt <= now;
};
```

---

Before:

```ts
// ❌ Impure — computing seconds in render using Date.now()
export const WakesIn = ({ awakeAt }) => {
  const secondsLeft = (awakeAt - Date.now()) / 1000;
};
```

After:

```ts
// ✅ Pure — countdown handled by hook
export const WakesIn = ({ awakeAt }) => {
  const { totalSeconds: secondsLeft } = useCountdown(awakeAt);
};
```

---

Before:

```ts
// ❌ Ref storing a random index read during render
export const useRandomItem = (array) => {
  const randomIndex = useRef(Math.floor(Math.random() * array.length));
  return array[randomIndex.current];
};
```

After:

```ts
// ✅ Use stable state initializer
export const useRandomItem = (array) => {
  const [index] = useState(() => Math.floor(Math.random() * array.length));
  return array[index];
};
```

---

### The Problem

Reading ref.current in render violates purity:

- Refs are mutable
- They can be changed outside React’s render cycle
- Reading them makes components impure
- React Compiler cannot optimize components that depend on mutable reads

### Example

Before:

```ts
// ❌ Using ref for random direction and reading in render
const { current: direction } = useRef(randomInt(0, 2) === 0 ? "left" : "right");
```

After:

```ts
// ✅ State initializer runs once
const [direction] = useState(() => (randomInt(0, 2) === 0 ? "left" : "right"));
```

Rule: If a value affects what you render, it must live in state, not a ref.

---

### The Problem

A common anti-pattern looks like:

```ts
useEffect(() => {
  setValue(props.value);
}, [props.value]);
```

This causes:

- Redundant state
- Unnecessary re-renders
- Possible render loops
- Hard-to-follow data flow
- Worse React Compiler outcomes

### Solutions

Use props directly unless you truly need a one-off snapshot via a lazy `useState` initializer.

### Examples

Before:

```ts
// ❌ Mirroring localStorage into state with an effect
export const PlazaSettings = () => {
  const [mutedPlayers, setMutedPlayers] = useState([]);

  useEffect(() => {
    setMutedPlayers(JSON.parse(localStorage.getItem("plaza.muted") || "[]"));
  }, []);
};
```

After:

```ts
// ✅ Lazy initializer — no effect needed
export const PlazaSettings = () => {
  const [mutedPlayers] = useState(() =>
    JSON.parse(localStorage.getItem("plaza.muted") || "[]"),
  );
};
```

---

Before:

```ts
// ❌ Syncing message into state
export const TypingMessage = ({ message, forceShowFullMessage }) => {
  const [displayedMessage, setDisplayedMessage] = useState("");

  useEffect(() => {
    if (forceShowFullMessage) {
      setDisplayedMessage(message);
    }
  }, [message, forceShowFullMessage]);

  return displayedMessage;
};
```

After:

```ts
// ✅ Use prop directly when appropriate
export const TypingMessage = ({ message, forceShowFullMessage }) => {
  const [displayedMessage] = useState("");
  return forceShowFullMessage ? message : displayedMessage;
};
```

---

### The Problem

Defining components inside another component:

- Recreates the component function on every render
- Causes re-mounting
- Breaks identity
- Prevents React Compiler optimizations

### Example

Before:

```tsx
export const Message = () => {
  const Content = () => {
    // ...
  };
  return <Content />;
};
```

After:

```tsx
export const Message = () => {
  return (
    // ...inline JSX...
  );
};
```

---

## When We Intentionally Override Rules

Some cases require `eslint-disable` because the side effect must happen in an effect.

### 1. DOM Measurement

```ts
useLayoutEffect(() => {
  // eslint-disable-next-line react-hooks/set-state-in-effect
  setImageWidth(widthFromDOM);
}, []);
```

### 2. Portal Setup

```ts
useLayoutEffect(() => {
  // eslint-disable-next-line react-hooks/set-state-in-effect
  setWrapperElement(el);
}, [wrapperId]);
```

### 3. Animation Triggers

```ts
useEffect(() => {
  if (open) {
    // eslint-disable-next-line react-hooks/set-state-in-effect
    shake();
  }
}, [open]);
```

### 4. Controlled Fade-Out / Transition Timing

```ts
useEffect(() => {
  if (showPopover) {
    // eslint-disable-next-line react-hooks/exhaustive-deps
    setContent(children);
  }
}, [showPopover, children]);
```

---

## Updated useCountdown API

Before (in render):

```ts
const seconds = (readyAt - Date.now()) / 1000;
```

After:

```ts
const { totalSeconds } = useCountdown(readyAt);
```

`useCountdown(targetDate)` now:

- Internally computes time remaining using `Date.now()` and an interval.
- Returns an object `{ days, hours, minutes, seconds, totalSeconds }`.
- Stops its internal timer automatically once it reaches zero, so callers do not need to manage cleanup.

Call sites typically use the `totalSeconds` field for gating UI, e.g.:

- `const { totalSeconds: secondsLeft } = useCountdown(awakeAt);`
- `const { totalSeconds: secondsTillReady, ...ready } = useCountdown(readyAt ?? 0);`

All time-based UI logic now goes through hooks like `useNow` and `useCountdown` rather than computing with `Date.now()` directly in render.

---

## Notable Domain & API Changes

These changes are mostly signature tweaks to keep React components pure while preserving domain semantics:

- Reward / airdrop flows:
  - `ClaimReward` now accepts `reward: Omit<IAirdrop, "createdAt"> & { createdAt?: number }`.
  - Many callers no longer synthesize `createdAt: Date.now()` in the UI; when needed, creation time is determined deeper in the game logic or provided by the backend.
- Time-aware game logic functions:
  - `getPetEnergy`, `feedPet`, `getCookingTime`, and `getFoodExpBoost` now take an optional time argument (e.g. `createdAt` / `cookStartAt`) that defaults to `Date.now()`.
  - This keeps these functions testable and explicit about time while allowing React components to call them without computing `Date.now()` in render.
- Minigame helpers:
  - `isMinigameComplete` now works directly with the `minigames` subtree (`GameState["minigames"]`) instead of the full `GameState`, which makes it easier to reason about and test in isolation.
- Types:
  - A `FarmHands` type alias was extracted and is now used by `GameState["farmHands"]` to clarify intent without changing runtime behavior.

If you are calling these functions from new code, prefer passing an explicit `createdAt` / `now` when you are in game logic or tests, and omit it when you are in React UI code.

---

## Moving Forward (Rules of Thumb)

1. Never call Date.now() or Math.random() in render.
2. Never read ref.current in render.
3. Avoid copying props into state unless you need a one-off snapshot.
4. Don’t define components inside components.
5. When overriding lint rules, add clear justification.

### Hook Reference

- useNow()
- useMathRandom()
- useRandomItem()
- useCountdown()

---

## Tooling & Infrastructure Notes

- React Compiler is enabled via `babel-plugin-react-compiler` in `vite.config.ts` through the React plugin; no new scripts or commands are required.
- Several Github workflows (`pwa.yml`, `testnet.yml`) and PWA-related code paths were touched to keep builds and deployment aligned with the new setup.
- The `ServiceWorker` component in `App.tsx` no longer calls `useServiceWorkerUpdate`; service worker registration remains, but the automatic update prompt hook is currently disabled and can be revisited separately if needed.

---

## Summary

This PR:

- Integrates the React Compiler
- Eliminates impure render logic
- Replaces time/random reads with hooks
- Removes redundant state
- Removes nested component definitions

All functionality remains the same, but components now follow the strict purity model required by React Compiler.

## Testing

Its important to test on the test url [here](https://sunflower-land.com/pwa/#/) as we need to confirm action after compiled.

- All game events: planting, chopping, mining
- Weather events starting/ending
- Turn of the day happening correctly
- An auction timer working
- Deliveries, bounties and marketplace

### DIFF
--- a/src/features/game/expansion/components/potions/MixingPotion.tsx
+++ b/src/features/game/expansion/components/potions/MixingPotion.tsx
@@ -9,10 +9,11 @@ import potionMasterBombSheet from "assets/npcs/potion_master_sheet_bomb.png";
 import potionMasterSuccessSheet from "assets/npcs/potion_master_sheet_success.png";
 import { SpringValue } from "react-spring";
 import { PotionHouseMachineInterpreter } from "./lib/potionHouseMachine";
-import { useActor } from "@xstate/react";
+import { useActor, useSelector } from "@xstate/react";
 import { Context } from "features/game/GameProvider";
 import { calculateScore } from "features/game/events/landExpansion/mixPotion";
 import { SpeechBubble } from "./SpeechBubble";
+import { MachineState } from "features/game/lib/gameMachine";
 
 export type DesiredAnimation =
   | "idle"
@@ -77,6 +78,12 @@ const getFPS = (animation: DesiredAnimation, frameNumber: number): number => {
   }
 };
 
+const _potionHouseLastAttempt = (state: MachineState) => {
+  const potionHouse = state.context.state.potionHouse;
+  const previousAttempts = potionHouse?.game.attempts ?? [];
+  return previousAttempts[previousAttempts.length - 1] ?? [];
+};
+
 export const MixingPotion: React.FC<{
   potionHouseService: PotionHouseMachineInterpreter;
   feedbackText: string | null;
@@ -100,10 +107,7 @@ export const MixingPotion: React.FC<{
   const currentAnimation = getCurrentAnimation();
   const settings = SETTINGS[currentAnimation];
 
-  const potionHouse = gameService.getSnapshot().context.state.potionHouse;
-  const previousAttempts = potionHouse?.game.attempts ?? [];
-  const lastAttempt = previousAttempts[previousAttempts.length - 1] ?? [];
-
+  const lastAttempt = useSelector(gameService, _potionHouseLastAttempt);
   const isGuessing = lastAttempt.some((potion) => potion.status === "pending");
 
   useEffect(() => {
@@ -116,13 +120,9 @@ export const MixingPotion: React.FC<{
     } else {
       potionHouseService.send("BOMB");
     }
-  }, [isGuessing, potionHouseService]);
+  }, [isGuessing, lastAttempt, potionHouseService]);
 
   const handleNextAnimation = () => {
-    const potionHouse = gameService.getSnapshot().context.state.potionHouse;
-    const previousAttempts = potionHouse?.game.attempts ?? [];
-    const lastAttempt = previousAttempts[previousAttempts.length - 1] ?? [];
-
     const isGuessing = lastAttempt.some(
       (potion) => potion.status === "pending",
     );
@@ -130,10 +130,6 @@ export const MixingPotion: React.FC<{
 
     potionHouseService.send("NEXT_ANIMATION", { score });
   };
-
-  // useEffect(() => {
-  //   setLoaded(true);
-  // }, []);
 
   return (
     <div className="flex flex-col items-center justify-evenly relative w-full h-full">


spencerdezartsmith:
> Fanstastic mate 👏  it seems to be achieving the goal. A number of components are suddenly not appearing in the debugger when planting. It's looking very promising.
> 
> **Before**:
> Before
> 
> **After**:
> After
> 
> A few notes:
> 
> - We pushed an update where water well was throwing an exception on load, in the situation a water well was max level, 4.
> - Another bug yet to be patched: The inventory select buttons on the HUD aren't functioning. 
> - @eliasSFL also found some timer related issues 

Excellent to hear! I will get a list of bugs tomorrow and finish this beast off <3

PR #6591, opened 2025-11-27
[FIX] Fruit Tree Timers

spencerdezartsmith:
# Description

Fruit tree timers were not updating. The React Compiler treated `getFruitTreeStatus(plantedFruit)` as a pure, memoizable computation (because its only argument plantedFruit doesn’t change while a tree is replenishing), even though it was secretly reading the clock via getTimeLeft → Date.now(). So:
useUiRefresher did cause FruitTree to re‑render every second.

But the compiler could safely reuse the previous result of
getFruitTreeStatus(plantedFruit) since the argument hadn’t changed.

Fixes #issue

- `yarn build && yarn preview`
-  harvest a fruit tree
-  confirm the timer goes down

# Checklist:

### DIFF
--- a/src/features/island/fruit/FruitTree.tsx
+++ b/src/features/island/fruit/FruitTree.tsx
@@ -1,8 +1,7 @@
 import React, { useContext } from "react";
 
-import { getTimeLeft } from "lib/utils/time";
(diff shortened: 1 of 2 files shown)


c-py (approved):
LGTM ✅

PR #6595, opened 2025-11-27
[FEAT] Allow translation of nodes and resources

AzaroQL:
# Description

Nodes and resources will use translated name if provided. That will enable better game localization.

image

image

This is only first PR that includes nodes and resources. If approved Ill add translation term for buildings, crops, animal produce and etc, excluding collectibles and wearables.

English version remains unaffected, localized versions will have new translations.(After relevant json file is updated.)


eliasSFL (approved):
Resolved conflicts for you

PR #6600, opened 2025-11-28
[FIX] Spawning mushrooms

spencerdezartsmith:
# Description

This PR fixes a problem where mushrooms would either not show up at all, or show up very small, after we switched to the React Compiler.

The issue was inside our `SpriteAnimator` (`Spritesheet`) component. It assumed that layout and zoom values were ready as soon as the component mounted. With the React Compiler, these values can briefly be `0` or not measured yet, which caused bad size calculations.

## What was going wrong

- On the first render:
  - `this.spriteEl.getBoundingClientRect().width` often returned `0`.
  - `zoomScale.get()` sometimes returned `0`.
- We used those values to calculate `spriteScale`, which decides how big the sprite should be drawn.
- This caused:
  - `spriteScale` to sometimes become `0`, so mushrooms were effectively invisible until you resized the window.
  - When we avoided the invisible case, mushrooms could still be too small because we weren’t using their intended styled width.

### Safer `resize` in `SpriteAnimator`

- If `spriteEl` or `spriteElContainer` is not ready yet, `resize` now returns early instead of trying to measure them.
- If `zoomScale.get()` is `0` or invalid, we now fall back to `zoom = 1` so we never divide by zero.
- If the DOM width for the sprite is `0`, we now:
  - Look at `this.props.style.width` (for example, the mushroom uses `PIXEL_SCALE * 10`).
  - Parse that value and use it as the effective width.
- If we still can’t get a good width, we fall back to a neutral `spriteScale = 1` so the sprite is at least visible.
- We then:
  - Apply the scale to the container’s `transform`.
  - Compute the container height and set the sprite’s height based on that and the zoom.

### Safer `getInfo("width")` and `getInfo("height")`

- If the container does not exist, we now return `0` instead of throwing an error.
- Both `width` and `height`:
  - Use `zoom = 1` when `zoomScale.get()` is `0` or invalid.
  - Divide the container’s measured width/height by this safe zoom value.

## Why this fixes mushrooms

- On initial load:
  - If the DOM reports a width of `0`, the component now falls back to the explicit `style.width` passed in, which matches the size we expect for mushrooms.
  - This value is used to compute `spriteScale`, so mushrooms render at the correct size right away.
- We no longer produce a `spriteScale` of `0`, so mushrooms do not disappear until you resize the window.
- Overall, mushrooms now:
  - Show up on first load.
  - Have the correct size that matches their styled width and the current zoom.

Fixes #issue

- You can auto give yourself mushrooms inside `populateMushrooms` in the api and then autosave and confirm the mushrooms spawn.

# Checklist:

### DIFF
--- a/src/components/animation/SpriteAnimator.tsx
+++ b/src/components/animation/SpriteAnimator.tsx
@@ -251,13 +251,44 @@ class Spritesheet extends React.Component<Props> {
 
   resize = (callback = true) => {
     const { widthFrame, onResize } = this.props;
+    if (!this.spriteEl || !this.spriteElContainer) return;
 
-    this.spriteScale =
-      this.spriteEl.getBoundingClientRect().width /
-      widthFrame /
-      this.zoomScale.get();
+    const elementWidth = this.spriteEl.getBoundingClientRect().width;
+    // Fallback to an explicit style width if the element has not yet been laid out
+    let effectiveWidth = elementWidth;
+    if ((!effectiveWidth || Number.isNaN(effectiveWidth)) && this.props.style) {
+      const styleWidth = this.props.style.width as number | string | undefined;
+
+      if (typeof styleWidth === "number") {
+        effectiveWidth = styleWidth;
+      } else if (typeof styleWidth === "string") {
+        const parsed = parseFloat(styleWidth);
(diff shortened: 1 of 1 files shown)


eliasSFL (approved):
LGTM!

PR #6613, opened 2025-12-02
[FIX] Existing node search when placing nodes

jcads:
# Description

Fixes the issue where a Sacred Tree disappeared from the chest after being dug from the farm. 

Also simplified the existing node search logic and `getChestItems` - it now compares by name and added a handler for new accounts if resources don't have a `name` yet

- Place and unplace node resources, making sure it still shows up in the chest after digging a node resource

### DIFF
--- a/src/features/game/lib/resourceNodes.ts
+++ b/src/features/game/lib/resourceNodes.ts
@@ -17,6 +17,8 @@ import {
   RESOURCES,
   UpgradeableResource,
   TreeName,
+  BASIC_RESOURCES,
+  BasicResourceName,
 } from "../types/resources";
 
 export const canGatherResource = (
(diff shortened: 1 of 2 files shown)


jcads:
@c-py 
> Bug 1:
After forging a Sacred Tree on infernos island, it's not showing up in my inventory to be placed. getChestItems may need to be double checked for brand new nodes.

Forge nodes are automatically placed on your island, which is why it’s not appearing in the inventory

> I also noticed I have 9 normal trees in my inventory, yet the forge screen is only displaying the count of trees I have placed (1). I need to place the trees in order to use them in the forge.

This has always been the intended behavior for node forging - trees need to be placed before they can be upgraded. That said, I agree the UX could be improved to make this clearer

PR #6620, opened 2025-12-03
[FEAT] Daily boxes

adamhannigan:
# Description

This PR revamps the daily reward system.

1. No longer needs an on-chain TX. On chain activity is no longer critical (DappRadar closure + less industry reliance on these metrics)
2. Add more enticing rewards + 7 day big reward to encourage new players to continue playing.

Screenshot 2025-12-03 at 1 22 55 pm

**Flagged**

This feature is flagged for beta testing. When the flag is on, you can no longer see the old Daily Reward system.

## How to test?

Claim your daily reward each day. Go into the DB and update your `gameState.dailyReward.chest.collectedAt` to be the previous day timestamp so you can claim again.

Test these areas:
1. Modal should auto pop-up when available
2. Go into the top left HUD rewards section

### DIFF
--- a/src/features/game/types/dailyRewards.ts
+++ b/src/features/game/types/dailyRewards.ts
@@ -30,7 +30,7 @@ const WEEKLY_REWARDS: DailyRewardDefinition[] = [
     },
   },
   {
-    id: "weekly-tiki-totem",
+    id: "weekly-time-warp-totem",
     label: "Time Warp Totem",
     items: {
       "Time Warp Totem": 1,
--- a/src/features/game/events/landExpansion/claimDailyReward.test.ts
+++ b/src/features/game/events/landExpansion/claimDailyReward.test.ts
@@ -215,6 +215,7 @@ describe("claimDailyReward", () => {
     const state = claimDailyReward({
       state: {
         ...INITIAL_FARM,
+        inventory: {},
         bumpkin: {
           ...TEST_BUMPKIN,
           experience: 10000,


src/features/game/types/dailyRewards.ts
```diff
@@ -0,0 +1,143 @@
...
+    label: "Basic Food Box",
+    items: {
+      "Basic Food Box": 1,
+    },
+  },
+  {
+    id: "weekly-tiki-totem",
+    label: "Tiki Totem",
+    items: {
+      "Tiki Totem": 1,
```
vitt0: Shouldn't this be Time Warp Totem?

adamhannigan: Oops. Good catch

spencerdezartsmith (dismissed):
This is a great update. 

I have pushed a couple of changes to remove the `Date.now()` from the static state that was being set. I have actually updated that to a memoised value instead. I narrowed the arguments required by the functions `getDailyRewardStreak` and `isDailyRewardReady`. That meant I could use very clear dependencies in the `useMemo` hook. 

```ts
  const bumpkinExperience = useSelector(gameService, _bumpkinExperience);
  const dailyRewards = useSelector(gameService, _dailyRewards);

  const now = useNow({ live: true });
  const currentDate = new Date(now).toISOString().substring(0, 10);

  const [showClaim, setShowClaim] = useState(false);

  const hasClaimed = !isDailyRewardReady({
    dailyRewards,
    bumpkinExperience,
    now,
  });

  const rewards = useMemo(() => {
    let streak = getDailyRewardStreak({ dailyRewards, currentDate });

    if (hasClaimed) {
      streak -= 1;
    }

    return new Array(7).fill(null).map((_, index) => {
      return {
        day: streak + index + 1,
        reward: getRewardsForStreak({
          streak: streak + index,
          currentDate,
        }),
      };
    });
  }, [dailyRewards, hasClaimed, currentDate]);
```

I have also used the `useNow` hook with a live value because the timer was not updating in real time. Also, now pass the `now` into `secondsTillReset` just so we are explicit about wanting to rerender the time on the tick. 

There is one issue and that is when you load the game and its a you have a new daily reward AND leagues try to update you get the reward pop up first and click `claim` but then you immediately get the league modal and have to update leagues and THEN you get your reward box. 

Non blocking for now though. Nice improvement!

Xeko-92:
Idk if this matters, but it still says "weekly tiki totem"
image

spencerdezartsmith (approved):

adamhannigan:
Thanks @spencerdezartsmith Those changes make sense.

I was attempting to extract it into the setState, but didn't use the correct hooks. Cheers

PR #6518, opened 2025-11-11
[FIX] Node Forging showing wrong requirements

eliasSFL:
# Description

This PR Fixes an issue where the balance checks was incorrectly displayed for T3 nodes, it uses T2 nodes but it was displaying the balance of the T1 instead of the T2

Fixes #issue

# What needs to be tested by the reviewer?

- Go to Forge in Infernos
- Ensure that the balance for the T3 is using T2 nodes instead of T1

# Checklist:

- [ ] Title of the PR is relevant and is prefixed with [FEAT], [CHORE] or [FIX]
- [ ] Screenshot if it includes any UI changes
- [ ] I have read the contributing guidelines and agree to the T&Cs
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #6635, opened 2025-12-09
[FIX] use tier-based filtering for node forging requirements (crafting)

malkovitc:
## Summary
- Fixes #6518 - Node Forging showing wrong requirements

The previous logic in `CraftingRequirements.tsx` incorrectly counted base tier nodes when checking requirements for upgraded resources. For example, when checking requirements for "Reinforced Stone Rock" (tier 3) which needs 4 "Fused Stone Rock" (tier 2) nodes, the code was counting all stone nodes regardless of tier.

### Changes
- Determines the expected tier from `ADVANCED_RESOURCES` for upgraded resources (tier 2 or 3)
- Base resources default to tier 1
- Filters nodes by their actual `tier` property (defaults to 1 if not present)
- Ensures only placed nodes are counted (x/y coordinates defined)

This approach is consistent with `getUpgradeableNodes()` in `resourceNodes.ts`.

## Test plan
- [ ] Check that forging "Fused Stone Rock" correctly shows count of tier 1 Stone Rock nodes
- [ ] Check that forging "Reinforced Stone Rock" correctly shows count of tier 2 Fused Stone Rock nodes
- [ ] Verify the same behavior for Trees, Iron, and Gold upgrade paths

### DIFF
--- a/src/components/ui/layouts/CraftingRequirements.tsx
+++ b/src/components/ui/layouts/CraftingRequirements.tsx
@@ -28,7 +28,6 @@ import {
   TemporaryCollectibleName,
 } from "features/game/lib/collectibleBuilt";
 import {
-  RESOURCES_UPGRADES_TO,
   ADVANCED_RESOURCES,
   RESOURCES,
   UpgradedResourceName,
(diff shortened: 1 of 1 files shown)


malkovitc:
Got it, cleaned up! Now it's only the tier-based filtering change in CraftingRequirements.tsx.

Split the other stuff into separate PRs:
- #6662 - 730 day streak label fix
- #6663 - marketplace price history fix

PR #6646, opened 2025-12-10
[FIX] Throttle

adamhannigan:
# Description

For players who face throttle requests, auto-retry so they can keep playing


c-py (approved):
LGTM!

PR #6658, opened 2025-12-14
[CHORE] Optimize Chest component filtering performance

malkovitc:
## Summary
- Wrap collectible categorization in `useMemo` in Chest.tsx to prevent recalculation on every render
- Use `Set` for O(1) lookups instead of O(n) `array.includes()` checks in Chest.tsx
- Wrap item categorization in `useMemo` in Basket.tsx (25+ filter/sort operations)

## Problem
**Chest.tsx:** 11 sequential `.filter()` operations on every render. The `decorations` filter checked `.includes()` against 10 arrays, creating O(n²) complexity.

**Basket.tsx:** 25+ `getItems()` filter calls plus sorting (foods by cooking time, coupons/fish alphabetically) on every render.

## Solution
- Memoize all filtering logic with `useMemo` keyed on relevant dependencies
- Convert arrays to Sets before checking membership (O(1) vs O(n)) in Chest
- Move helper functions inside useMemo for proper scoping

## Performance Impact
- **Chest:** 11 filter + multiple O(n) includes → cached, only recalculates when collectibles change
- **Basket:** 25+ filter/sort operations → cached, only recalculates when basketMap changes

## Test plan
- [x] All 2439 tests passing
- [x] TypeScript compilation successful
- [x] ESLint checks passing
- [ ] Manual testing: open inventory, verify all categories display correctly

### DIFF
--- a/src/features/island/hud/components/inventory/Chest.tsx
+++ b/src/features/island/hud/components/inventory/Chest.tsx
@@ -369,23 +369,24 @@ export const Chest: React.FC<Props> = ({
     );
   }
 
+  const collectibleNames = getKeys(collectibles);
+
   // Sort collectibles by type
-  const resources = getKeys(collectibles).filter((name) => name in RESOURCES);
-  const buildings = getKeys(collectibles).filter((name) => name in BUILDINGS);
-  const monuments = getKeys(collectibles).filter((name) => name in MONUMENTS);
-  const villageProjects = getKeys(collectibles).filter(
+  const resources = collectibleNames.filter((name) => name in RESOURCES);
+  const buildings = collectibleNames.filter((name) => name in BUILDINGS);
+  const monuments = collectibleNames.filter((name) => name in MONUMENTS);
+  const villageProjects = collectibleNames.filter(
     (name) => name in REWARD_ITEMS,
   );
 
-  const banners = getKeys(collectibles).filter((name) => name in BANNERS);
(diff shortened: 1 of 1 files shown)


spencerdezartsmith (changes requested):
Thanks for the PR! Appreciate you thinking about render performance 👍

Just as a bit of context: we’ve recently removed a lot of manual memoisation across the codebase as part of adopting the React Compiler. Our current direction is to rely on the compiler for render optimisations rather than adding useMemo / useCallback, as those have historically added complexity and sometimes obscured the underlying logic.

Even with 1000+ items the classification work is small compared to the rest of the UI, so the extra useMemo doesn’t really provide a meaningful real‑world benefit. I’d prefer to rely on React/React Compiler for memoization unless we see concrete performance issues in this screen, in which case we can revisit with measurements.

The `Set` update does have some merit although it does add some more code complexity. If you want to refactor to just add the `Set` update you can.

malkovitc:
Done! Removed useMemo, kept only the Set optimization for O(1) lookups. Also reverted Basket.tsx changes - that one didn't need any optimization anyway.

Makes sense about React Compiler handling memoization - good to know for future PRs 👍

.claude/settings.local.json
```diff
@@ -0,0 +1,5 @@
+{
```
spencerdezartsmith: Also we don't want to commit this you should should set up a gitignore on your side.

PR #6668, opened 2025-12-17
[FIX] Remove PartyModal

c-py:
# Description

The Party Modal was created for the Gam3s event where players would get rewards. The event is now over, however the Plaza is still showing confetti when the plaza initially loads. Even though the Party modal is not open.

This PR deletes the party modal, as it now unused since the end of the Gam3s event.

1. Open Plaza and ensure no random confetti shows

# Checklist:


eliasSFL (dismissed):
LGTM!

PR #6696, opened 2025-12-22
Remove flower teaser

brian-man:


adamhannigan:
Thanks @brian-man We picked up a few more unused cases that we've included in this PR as well: https://github.com/sunflower-land/sunflower-land/pull/6713

PR #6704, opened 2025-12-26
[Feat] duplicate marketplace listing

Josephgetzen:
This PR introduces a "Copy" button to the My Listings tab in the Marketplace profile. This feature allows players to quickly duplicate an active listing without having to manually re-enter the item quantity and price.

Motivation:
For players using the marketplace frequently this should provide a moderate quality of life improvement.

Changes:
Modified MyListings.tsx:
Implemented handleDuplicate which immediately dispatches the marketplace.listed event using the details of the selected listing.
Added logic to calculate canDuplicate. The button is disabled if:
The item listed isn't a wearable or collectible.
The player does not have enough inventory remaining to create another listing.
The player has reached their trade limit or lacks the required reputation.
The listing is an on-chain listing (signature based).
Modified MyTableRow.tsx:
Added the "Copy" button UI.
Button styling to use variant="secondary" to match the adjacent "Cancel" button.
Added dynamic width classes to ensure the buttons fit correctly whether there is one button (Claim/Cancel) or two (Copy + Cancel).

Testing:
Please be advised this should be fully tested in a staging environment. I have only tested the functionality using mock data locally.

Styling: Verified the button looks consistent with the "Cancel" button next to it.

Functionality: Verified the "Cancel" and "Claim" buttons still function as intended.

Checklist:

I have run yarn tsc to check for type errors.

I have run yarn test to ensure no regressions.

I have manually tested the feature in my local environment with mock data.

Once/If this is merged I can add the same functionality to MyOffers.tsx


Xeko-92:
Shouldn't we just have bulk listings like we have bulk purchases?

Josephgetzen:
> Shouldn't we just have bulk listings like we have bulk purchases?

The two features aren't mutually exclusive and I like this for the simple UX/UI

Xeko-92:
> > Shouldn't we just have bulk listings like we have bulk purchases?
> 
> The two features aren't mutually exclusive and I like this for the simple UX/UI

Sure, they aren't mutually exclusive, but we also don't want a scenario where things are getting too convoluted. Since we have bulk buying, bulk listings makes more sense to have, and I think it's mostly just a better version of duplicating listings.

spencerdezartsmith:
Hey @Josephgetzen thank you so much for your PR! I really appreciate the initiative. 

However, we will leave this one for now as the intention is to build bulk listings after this next chapters work is complete. Rather than setting up this interim flow and having players learn that only to remove it soon after,  we will just wait for the feature to be done 🤗

PR #6717, opened 2025-12-31
[FIX] Gem Starter Bundle

adamhannigan:
# Description

Do not allow players to buy the starter bundle with Gems (via FLOWER) again if they already have.

**How to test?**

1. Create a new account
2. Buy the bundle with FLOWER
3. It will disappear


c-py (approved):
✅ It disappeared

PR #6740, opened 2026-01-06
[FIX] Enhance feed interactions pagination and caching behavior

spencerdezartsmith:
# Description

The `useFeedInteractions` was fetching the first page multiple times. I have addressed this in this PR.

- Swapped the feed interactions SWR key to a tuple and tightened revalidation so page 0 stops refetching; parsing now uses structured params instead of brittle string splits.

- Kept pagination intact while dialing back focus/reconnect/stale revalidation and bumping the dedupe window to cut redundant calls.

Real-time/social-room pushes still keep the feed fresh—explicit mutate can be used if we ever need an immediate refresh.

Fixes #issue

- Check that feed loads correctly on farm A
- Check that infinite scroll works on farm A
- Confirm there is just one network request for each
- Open and close, travel to the plaza the feed and confirm there is no other network request
- Open up another farm B and do an action on farm A (clean farm A from farm B). Confirm that the interaction comes through on farm A.

# Checklist:

### DIFF
--- a/src/features/social/hooks/useFeedInteractions.ts
+++ b/src/features/social/hooks/useFeedInteractions.ts
@@ -14,28 +14,56 @@ export function useFeedInteractions(
   isGlobal: boolean,
 ) {
   const getKey = (
-    _: number,
+    pageIndex: number,
     previousPageData: { feed: Interaction[]; following: number[] },
   ) => {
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):
Nice work!

I tested moving around the game, into the plaza, opening the feed etc. and it is called much less. 🚀

PR #6761, opened 2026-01-08
[FEAT] Show chums by trap type

jcads:
# Description

Show chums by trap type for a guaranteed catch

https://github.com/user-attachments/assets/a50b9a0d-13c7-4908-a3d0-6d13ab52d175

- Open the Crustacean modal, it should only show the valid chum for the selected water trap


Xeko-92:
I like that you made all the pet items 3, but moonfur should be 1. Moonfur is only obtainable by NFT pets and costs 5x as much energy.

Xeko-92:
Crabs sticks should be removed from possible chum here. We shouldn't be using crab sticks to catch crustaceans to make more crab sticks.

PR #6764, opened 2026-01-08
[FIX] Exclude disabled tools in restock modal

jcads:
# Description

The Pest Net was still rendered in the restock modal even when disabled


eliasSFL (dismissed):
LGTM!

PR #6778, opened 2026-01-12
[FEAT] Implement withdraw cooldown

adamhannigan:
# Description

We've had a spike in multi-accounters and similar IP addresses withdrawing. To reduce these cases, we are implementing IP cooldowns on the withdraw endpoint.

This change adds more friction to malicious actors.

A normal player can withdraw as many times as they wish. Shared households will need to wait the cooldown between withdrawals, use a hotspot or alternate network.

## How to test?

1. Try withdraw as many times as you want from one account
2. Login to another account and try to withdraw, you will be hit with the cooldown
3. Hotspot and withdraw from a different IP Address.

Screenshot 2026-01-12 at 2 28 09 pm


c-py (approved):
LGTM!

PR #6812, opened 2026-01-20
[FIX] Update trade types of chapter items

c-py:
# Description

These items come from Auctions.
They should be onchain trades  - so the supply does not change during trading.


spencerdezartsmith (approved):
LGTM

PR #6819, opened 2026-01-20
[FEAT] Crustacean guide and codex

jcads:
# Description

image

image

### DIFF
--- a/src/lib/i18n/dictionaries/dictionary.json
+++ b/src/lib/i18n/dictionaries/dictionary.json
@@ -7889,7 +7889,7 @@
   "petGuide.social.description": "Boost your pets by interacting with friends.",
   "petGuide.social.description2": "Friends can pet your companions for +5 XP each, up to 50 XP per pet per day.",
   "petGuide.social.description3": "NFT pets can walk with you in multiplayer areas. Enable this in the pet management screen. Only one pet can walk at a time.",
-  "petGuide.social.description4": "Visit friends' farms to pet their pets and build community!",
+  "petGuide.social.description4": "Visit your friends farms to pet their pets and build community!",
   "guide.crustacean.placeTrap": "Place a water trap in the water to catch marine creatures",
   "guide.crustacean.waitTime": "Wait for the trap to catch crustaceans over time",
   "guide.crustacean.useChum": "Use chum to attract specific types of crustaceans",


src/lib/i18n/dictionaries/dictionary.json
```diff
@@ -7889,5 +7889,10 @@
   "petGuide.social.description": "Boost your pets by interacting with friends.",
   "petGuide.social.description2": "Friends can pet your companions for +5 XP each, up to 50 XP per pet per day.",
   "petGuide.social.description3": "NFT pets can walk with you in multiplayer areas. Enable this in the pet management screen. Only one pet can walk at a time.",
-  "petGuide.social.description4": "Visit friends' farms to pet their pets and build community!"
+  "petGuide.social.description4": "Visit friends' farms to pet their pets and build community!",
```
spencerdezartsmith:
```suggestion
  "petGuide.social.description4": "Visit your friends farms to pet their pets and build community!",
```

src/features/island/hud/components/codex/pages/Crustaceans.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const [selectedCrustacean, setSelectedCrustacean] = useState<
+    CrustaceanName | undefined
+  >();
+  const [chumMapping, setChumMapping] = useState<CrustaceanChumMapping | null>(
+    null,
+  );
+
+  const { farmActivity } = state;
+
+  const allCrustaceans = getKeys(CRUSTACEANS);
```
spencerdezartsmith: The bots feedback is correct here. This would be a new array on every render and then it would recalculate the `useMemo` below.

I think you can implement this without using `useMemo`. Since the `allCrustaceans` is constant, you can initialise it outside of the component. Then you can just do the filtering inside of the `useEffect`.

```ts
const ALL_CRUSTACEANS = getKeys(CRUSTACEANS);

export const Crustaceans: React.FC<Props> = ({ state }) => {
...

useEffect(() => {
  const caught = ALL_CRUSTACEANS.filter(
    (name) => (farmActivity[`${name} Caught`] ?? 0) > 0
  );

  loadCrustaceanChums(caught).then(setChumMapping);
}, [farmActivity]);
```

spencerdezartsmith (changes requested):

PR #6842, opened 2026-01-26
[CHORE] Improve Skill Reset Remove Crops Warning

vitt0:
# Description

Multiple times I saw players asking about what this warning means, so I changed the description of skillReset.removeCrops to be more clear about what the players should do to be able to reset the skills.

# Checklist:


eliasSFL:
I think I will be removing these crop machine restrictions for skill reset. instead I need to ensure the crop machine still works even with excess packs / oil

PR #6856, opened 2026-01-28
[FEAT] Wharf Expansion

jcads:
# Description

Wharf overlay for each island type

### Basic
image

### Spring
image

### Desert
image

### Volcano

image

# Testing
- Verify that the wharf is correctly extended and that crab trap counts are accurate across island types
- Test the hitboxes for Fisherman and crab traps do not overlap


jcads:
> Fix the Fisherman NPC hitbox sizing (most likely to cause regressions).

Current code is already full-size so it should be fine

spencerdezartsmith (approved):
Working well!

PR #6858, opened 2026-01-29
[FIX] Tutorial Change Plot 5 Potatoes to Rhubarb

Jamboymagic:
Second land expansion (Plot 5) changed to contain Rhubarb rather than Potato to keep Seasonal planting consistency (Potatoes cant be planted on Spring Island, inconsistent to teach player to grow them as first Food lesson). 
Logic added to Fire Pit to default selection to a cookable item to prevent defaulting to uncookable mashed potato on first open and generically select lowest cookable item when opening the FirePit. 

New tutorial flow is completeable (can be tested in <2 mins from new account) 
Tutorial prompts are still showing 
FirePit now defaults to lowest cookable item when opened without a prior selection & there are valid items

### DIFF
--- a/src/features/island/buildings/components/building/firePit/FirePit.tsx
+++ b/src/features/island/buildings/components/building/firePit/FirePit.tsx
@@ -93,7 +93,8 @@ export const FirePit: React.FC<Props> = ({ buildingId, isBuilt, island }) => {
     }
   };
 
-  const showHelper = rhubarbCount.gte(3) && experience === 0 && !rhubarbTartCooked && !cooking;
+  const showHelper =
+    rhubarbCount.gte(3) && experience === 0 && !rhubarbTartCooked && !cooking;
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


src/features/island/buildings/components/building/firePit/FirePit.tsx
```diff
@@ -93,7 +93,7 @@ export const FirePit: React.FC<Props> = ({ buildingId, isBuilt, island }) => {
     }
   };
 
-  const showHelper = potatoCount.gte(8) && experience === 0 && !cooking;
+  const showHelper = rhubarbCount.gte(3) && experience === 0 && !cooking;
```
c-py:
```suggestion
  const showHelper = rhubarbCount.gte(3) && experience === 0 && !rhubarbTartCooked && !cooking;
```

Small suggestion, non blocking. After I cooked my first rhubarb tart the arrow showed to cook another, as I still had 5 in my inventory. This suggestion is to hide the arrow if the player has cooked a rhubarb tart already.

c-py (dismissed):
LGTM! Worked well, I've left one non-blocking comment in line around the arrow firepit helper.

c-py (approved):
LGTM!

AhmedMuhammedMuhammed4 (commented):
هلا

PR #6882, opened 2026-02-02
[CHORE] Chapter Tracks data telemetry added

Jamboymagic:
# Description

Added GameAnalytics Design events around Chapter Tracks, checking: 
* Track Opening
* Milestone Completion 
* Milestone Claiming
* Chore Completion 
* VIP Upsell Clickthrough 

Design events are firing correctly 
Claiming / track interaction is not broken as a result of these changes 
Events are not firing too often or creating performance overhead

### DIFF
--- a/src/features/game/events/landExpansion/completeNPCChore.ts
+++ b/src/features/game/events/landExpansion/completeNPCChore.ts
@@ -187,13 +187,10 @@ export function completeNPCChore({
         task: "chore",
         points: amount,
       });
-      const previousPoints =
-        draft.farmActivity[`${chapter} Points Earned`] ?? 0;
       handleChapterAnalytics({
-        chapter,
         task: "chore",
-        points: pointsAwarded,
-        previousPoints,
+        points: amount,
+        farmActivity: draft.farmActivity,
         createdAt,
       });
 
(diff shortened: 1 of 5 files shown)


eliasSFL:
Just some lint issues. You can run `yarn format` in your terminal to fix them, then push them up

src/features/game/events/landExpansion/completeNPCChore.ts
```diff
@@ -181,21 +187,70 @@ export function completeNPCChore({
     const amount = items[ticket] ?? 0;
 
     if (amount > 0) {
+      const chapter = getCurrentChapter(createdAt);
```
spencerdezartsmith, line 185: This logic appears to be repeated with just a couple different parameters. Could we turn this into a re-useable function?

```ts
function handleChapterAnalytics({ task: "chore" | "delivery" | "bounty", points: number }) {...}
```

spencerdezartsmith (approved):
Ahhhh I am sorry I didn't see that you were using the `chapter` and `pointsAwarded` further down. You could have left it as you had it. 

You can either leave it or change it back. Either way it non blocking.

PR #6887, opened 2026-02-02
[FIX] Megabox Change

Jamboymagic:
Megabox now always gives Gems to VIP players; non-verified players can only get the smaller rewards.


spencerdezartsmith (approved):
LGTM

PR #6902, opened 2026-02-03
[FEAT] Remove async crustacean pickup

jcads:
# Description

- Water trap flow: place and collect only - removed the pickup step for newly discovered crustaceans. Clicking a ready crab trap runs the collect action immediately
- Crustacean logic moved to frontend
     - Crustacean/chum mapping and catch calculation
     - Removed the loadCrustaceanChums action 
- Crustacean codex will now only show the Chum used for the caught crustacean if it can be caught with multiple chums

### UI updates

8 Crustaceans

• Crustaceans

- Ensure placing traps and collecting crustaceans still works as expected

### DIFF
--- a/src/features/island/fisherman/WaterTrapModal.tsx
+++ b/src/features/island/fisherman/WaterTrapModal.tsx
@@ -222,7 +222,7 @@ export const WaterTrapModal: React.FC<Props> = ({
                 </div>
               </InnerPanel>
             )}
-            {selectedChum && (
+            {selectedChum ? (
               <InnerPanel className="p-2">
                 <div className="flex justify-between">
                   {hasEnoughChum ? (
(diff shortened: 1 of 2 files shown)


src/features/game/events/landExpansion/placeWaterTrap.ts
```diff
@@ -20,12 +21,14 @@ type Options = {
...
   action: PlaceWaterTrapAction;
   createdAt?: number;
+  farmId?: number;
 };
 
 export function placeWaterTrap({
   state,
   action,
   createdAt = Date.now(),
+  farmId = 0,
```
spencerdezartsmith: Does this make sense? I don't think we should initialise the farmId.  I think it would make more sense to check for existence and throw if not.

src/features/island/fisherman/WaterTrapSpot.tsx
```diff
@@ -214,15 +238,33 @@ export const WaterTrapSpot: React.FC<Props> = ({ id }) => {
...
         <WaterTrapModal
           waterTrap={waterTrap}
           onPlace={place}
-          onPickup={pickup}
-          onCollect={collect}
-          onClose={() => setShowModal(false)}
+          onClose={() => setShowChumSelectionModal(false)}
         />
       </Modal>
+      {collectedCatch && (
```
spencerdezartsmith: I think this is causing the modal close animation to be removed. So what is happening is the modal shows and then it just snaps away without animating.

eliasSFL (approved):
LGTM!

spencerdezartsmith (approved):
LGTM

PR #6954, opened 2026-02-12
[FIX] Preserve upgraded resource equivalents on farm upgrades

jcads:
# Description

Players with upgraded resources could be treated as under-minimum because only base resource counts were checked. This caused incorrect top-ups during when upgrading. This PR ensures we only top up missing base resource amount when total equivalents are below minimum, preserving upgraded resources like Ancient Tree

- Upgrade a resource node while in desert
- Upgrade to volcano
-  Ensure you don't get extra base resource from the upgraded node

# Checklist:


eliasSFL (dismissed):
LGTM!

PR #6990, opened 2026-02-25
[FIX] Remove Potion Ticket from Hoarder Limit

c-py:
# Description

Removes hoarding limit for Potion Tickets

Manual inspection only

# Checklist:


eliasSFL (changes requested):
Will need to add it to offchain items as well, mostly for hoarding checks

eliasSFL (approved):

PR #7010, opened 2026-03-05
[FEAT] Pet Fetch Resource Boost Changes

eliasSFL:
# Description

This PR changes how fetch resource is fundamentally calculated.

Instead of a x% chance to get +1 fetch, it would give a flat +0.x yield instead.

For example, if it was a 10% chance to get +1 fetch previously, it would now give +0.10 yield

Fossil Shell is not affected by the decimal as it requires it to be an integer to open the loot box.

## Reason for change
There have been some issues with fetch resource prng in the past and how it can be exploited to get extra resources, or some players had bugs where the amount given wasn't what it showed in the UI. While most of the issues have been ironed out, changing it to a flat +0.x amount would remove that altogether

Fixes #issue

- Run FE + BE
- Level up your pet to get some of these fetch bonuses
- Fetch resource
- Ensure you get the correct amount of fetch resource

# Checklist:

### DIFF
--- a/src/features/game/events/pets/fetchPet.test.ts
+++ b/src/features/game/events/pets/fetchPet.test.ts
@@ -266,6 +266,28 @@ describe("fetchPet", () => {
     expect(state.farmActivity["Acorn Fetched"]).toBe(1);
   });
 
+  it("gives no Native boost at level 14 for non-Fossil resource", () => {
+    const state = fetchPet({
+      state: {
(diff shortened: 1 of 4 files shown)


brian-man:
Please exhaustively validate that all the logic flows are avoiding float so that we don't end up with 1.0000000000000002 or 0.9999999999998 yield junk going on.

PR #7031, opened 2026-03-17
[FIX] Feed when Main Bumpkin is placed

c-py:
# Description

The feed currently does not allow notifications to open the player modal if the `bumpkin` has coordinates set, i.e. the bumpkin is placed. This is due to a bug, where a listener on the playerNPCModal bus is not registered.

This PR updates the `PlacedBumpkin` component to use `PlayerNPC` instead of `NPCPlaceable`, which results in the PlayerNPC setting the correct listeners up.

## How to Test

On **main**
1. Have a bumpkin and at least one farm hand
2. Place the farm hand
3. Select the farm hand, and click `Set as Main Bumpkin`
4. Refresh the page
5. Open the feed and try open any notification. The feed should close without showing the Player Modal.
6. Test again on this PR and ensure the Player modal can open.

# Checklist:

### DIFF
--- a/src/features/island/bumpkin/components/PlacedBumpkin.tsx
+++ b/src/features/island/bumpkin/components/PlacedBumpkin.tsx
@@ -1,11 +1,10 @@
-import React, { useContext, useState } from "react";
+import React, { useContext } from "react";
 import { NPCPlaceable } from "./NPC";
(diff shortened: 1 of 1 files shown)


eliasSFL (approved):
LGTM! Verified that I can open player modal after placing bumpkin down

PR #7064, opened 2026-03-30
[FIX] Hoarder limit for Mutants of Crabs and Traps Chapter

vitt0:
# Pull request description

I added a hoarding limit for mutants in the Crabs and Traps chapter to prevent an AS-001 error when players exceed the limit, and to show that the cap has been reached instead.

---

### PR and scope


eliasSFL (approved):
LGTM!

PR #7112, opened 2026-04-16
[FIX] Digging up Pet House with pets inside

jcads:
## Summary

When visiting a farm, pets inside the Pet House no longer grant social XP or count as helpers if the Pet House building isn't placed on the farm. Pets placed directly on the farm or in the home are unaffected.

## How to test
  - Place a Pet House with a pet inside, then dig up the Pet House
  - Visit that farm as another player and verify:
    - HUD scroll count excludes the pet-house pet (shows 5 instead of 6)
    - Visitor Guide modal does not show the "In the pet house" section
    - Other pets placed on the farm / in the home are still clickable and award XP as before
  - Place the Pet House back and verify pet-house pets are clickable and award 5 XP again (up
   to the 50/day cap)


eliasSFL (changes requested):
Preventing players from digging up their pet house when there are pets inside would introduce alot of friction. If the concern was with social farming, if pet house is not present on the farm when visiting, the farm should exclude all pets in pet house for helping and giving of XP

eliasSFL (approved):
LGTM! Was able to verify that the pets in pet house didn't get XP when the pet house wasn't placed.

PR #7117, opened 2026-04-17
[FIX] April Fool Rugs Collision and Trading

vitt0:
# Pull request description

I added the rugs to the NON_COLLIDING_OBJECTS list so we can place things over them, I also added them to be tradable since they were missing.

## Checklist


eliasSFL (approved):
LGTM! Turns out the collisionDetection was missing already in the FE, BE already has it. It's only the withdrawables that need to be updated

PR #5880, opened 2025-07-25
[FEAT] Social Feed

spencerdezartsmith:
# Description

This PR introduces a new social feed system that allows players to see real-time interactions and activities from other farmers. The feed can be accessed both locally (showing nearby players) and globally (world-wide activity).
Key Changes:
New Feed Component: Added a unified Feed component that handles both local and world feeds with smooth animations
Enhanced HUD: Integrated feed button into the main HUD with responsive design for mobile/desktop
Player Interactions: Updated player modals to support the new social features
UI Improvements: Added proper loading states, infinite scroll, and following indicators

Fixes #issue

# What needs to be tested by the reviewer?

Please describe how this can be tested.

# Checklist:

- [ ] Title of the PR is relevant and is prefixed with [FEAT], [CHORE] or [FIX]
- [ ] Screenshot if it includes any UI changes
- [ ] I have read the contributing guidelines and agree to the T&Cs
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #7124, opened 2026-04-19
[CHORE] Social Feed Header Update

alexandrdragin:
**What changed**

Added a daily help counter to the Feed header, displayed as remaining/total
Added a compact Cheers badge to the same header, showing icon + available count

Implementation notes

Main UI update: [Feed.tsx]
Visit-state sync: [VisitingHud.tsx]
New counter logic in file: [useHelpCounter.ts]

Outcome
The social panel now surfaces both:

how many help actions the player still has available today
how many cheers remain available
in a single, contextually relevant location.

litte bit messy preview from local server

image


eliasSFL (changes requested):
Hey! Thanks for the PR — the daily help counter in the Feed is a nice addition. A few things to address before we merge:

1. Please drop the MarketplaceHotNow.tsx changes — they look unrelated to the help-counter feature and should go out in their own PR.

2. Avoid the full loadGameStateForVisit call on own-farm — useHelpCounter.ts:79

totalHelpedToday isn't in the machine context when the player is on their own farm, so the hook falls back to fetching the entire visit payload (both visitor + visited game states through makeGame) just to read one number. Since Feed is always mounted on the HUD, this fires on every own-farm session.

Options:

Include totalHelpedToday in the main game load response, or
Add a lightweight endpoint that returns just the counter.
3. Don't silently swallow the fetch error — useHelpCounter.ts:89-91

The empty .catch(() => {}) will make a broken endpoint look like "0 helps used" forever. Please at least console.error it, or surface it so we notice in telemetry.

4. Pick one owner for the localStorage cache

Both useSyncHelpCounterCache (in VisitingHud) and useHelpCounter (inside Feed) write the same value to localStorage via near-identical effects. When visiting, they both run and duplicate the work. Either:

Drop useSyncHelpCounterCache and let useHelpCounter own caching, or
Make useHelpCounter read-only and let the sync hook own writes.
5. Verify the counter doesn't lag after a help action

context.totalHelpedToday only updates on autosave response — the reducers that run on cheer/pet/garbage don't increment it locally. Please test in-browser: click help, watch the x/y label. If it stays stale until autosave fires, we'll need to increment totalHelpedToday locally in the action handlers.

6. Use the Label's icon prop instead of the custom stack — Feed.tsx:277-296

The hand-rolled absolute-positioned disc + drag icons with paddingLeft: "24px" are fragile and inconsistent with the neighboring cheer label. If a two-icon composite is intentional, extract a small component; otherwise collapse to a single icon={...} like the others.

Nice to have

Show —/5 (or a small spinner) while the fetch is in-flight instead of the ?? 0 fallback, so it doesn't briefly look like the user already used 0 helps.
The cachedUsedToday memo only depends on ownFarmId, so a session crossing UTC midnight won't refresh the cache key. Low risk since other sources usually cover it, but worth a look.
Let me know if any of this is unclear — happy to pair on #2 or #5 if needed.

alexandrdragin:
@eliasSFL Sorry for a mess: key changes here [Fixes after eliasSFL comment](https://github.com/sunflower-land/sunflower-land/pull/7124/commits/be335aa1452865de234e9ade4e2d3fa5802e2720)

**Avoid the full loadGameStateForVisit** — Removed from useHelpCounter.ts.
**Don't silently swallow the fetch error** — removing that fetch path entirely.
**Pick one owner for the localStorage cache** — Made useHelpCounter.ts the single localStorage cache owner
**Verify the counter doesn't lag after a help action** — Added local totalHelpedToday increment in gameMachine.ts, so visit-help actions update the counter immediately after successful local processing
**Use the Label's icon prop instead of the custom stack** — Replaced > Label icon={SUNNYSIDE.icons.drag}.
**Show —/5**  — Added --/total fallback
**session crossing UTC midnight won't refresh the cache** — refresh by rotating the cache key after UTC midnight.

eliasSFL (approved):

alexandrdragin:
@eliasSFL our friend @Dionis404 show me that cause 
image


I think my pr has little conflict with https://github.com/sunflower-land/sunflower-land/pull/5880
By my opinion simplest way to fix this its to do server name converter

sunflorea_oasis -> Bazaar
sunflorea_brazil -> Brazil
sunflorea_bliss -> Bliss
sunflorea_dream -> Dream
sunflorea_magic -> Magic
sunflorea_kale -> Kale
sunflorea_flower -> Flower

please approve this a approach and tell my is this a way to do it directly this pr or should i do new?

PR #7125, opened 2026-04-20
[FIX] Rebalance Salt Sculpture craft cost and upgrade ingredients

Jamboymagic:
Update ingredient costs to use new fermentation bait items instead of worms, reduce Refined Salt quantities, and lower Greenhouse item requirements to match backend changes.


eliasSFL (approved):
LGTM! ✅

PR #7144, opened 2026-04-21
[FIX] Aging Skill Icons Replace

vitt0:
# Pull request description

I replace 2 icons images that there were wrong and I also added the minus symbol in one the description of the skills that was missing.

## Checklist


eliasSFL (approved):
LGTM! Thanks for the icons as always!

PR #7155, opened 2026-04-22
[CHORE] Update InteractableModals.tsx (Clubhouse modal)

alexandrdragin:
# Pull request description

Update Button at InteractableModals which appears when you try to enter to Clubhouse with out Bud

Current state 
image

## Summary

Just changed Opensea to Marketplace and insert link directly to Buds

### DIFF
--- a/src/features/world/ui/InteractableModals.tsx
+++ b/src/features/world/ui/InteractableModals.tsx
@@ -699,7 +699,7 @@ export const InteractableModals: React.FC<Props> = ({ id, scene }) => {
                   },
                 },
                 {
-                  text: t("Marketplace"),
+                  text: t("marketplace"),
                   cb: () => {
                     window.open(
                       "https://sunflower-land.com/play/#/marketplace/collection?filters=buds",
--- a/src/features/world/ui/InteractableModals.tsx
+++ b/src/features/world/ui/InteractableModals.tsx
@@ -52,8 +52,6 @@ import { ChapterRaffles } from "./chapterRaffles/ChapterRaffles";
 import { FreeTrial } from "./FreeTrial";
 import { useNavigate } from "react-router";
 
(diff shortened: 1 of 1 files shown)


src/features/world/ui/InteractableModals.tsx
```diff
@@ -699,10 +699,10 @@ export const InteractableModals: React.FC<Props> = ({ id, scene }) => {
...
                 },
                 {
-                  text: t("opensea"),
+                  text: t("marketplace"),
                   cb: () => {
                     window.open(
-                      "https://opensea.io/collection/sunflower-land-buds",
+                      "https://sunflower-land.com/play/#/marketplace/collection?filters=buds",
                       "_blank",
                     );
```
eliasSFL: Thanks for the initiative!! However since we are referencing our own ingame marketplace, I think we can do something different instead. 
In the component above you will need to import useNavigate
then assign it to a const:

```ts
import { useNavigate } from "react-router";
...
  const navigate = useNavigate();
```

After that apply it to the button
```suggestion
                    navigate("/world/marketplace/collection?filters=buds");
                    closeModal();
```

using useNavigate will have better integration with our existing code and provide a smoother transition to the marketplace, as compared to using javascript's native window.open() command, which will redirect the player to that link and may cause longer waiting times

eliasSFL (changes requested):
Hey mate! Thanks for the initiative. I left a comment on how we can better make the flow a little bit smoother

src/features/world/ui/InteractableModals.tsx
alexandrdragin: Done

eliasSFL (approved):
LGTM! However just a minor change that I made for you: the useNavigate should be called in the React Component, instead of outside

PR #7171, opened 2026-04-24
[CHORE] Rework metadata image pipeline (WebP + auto-regen manifest)

eliasSFL:
## Summary

- Rebuilt the collectible metadata image pipeline to output **lossless WebP** (animated sources preserve all frames + delays), with filenames derived from `KNOWN_IDS[name]` instead of a per-entry `image` field — removes a class of bugs where distinct items (House/Manor/Mansion, etc.) clobbered each other's output.
- Added a per-item **manifest** (`public/erc1155/images/.manifest.json`) that fingerprints source bytes + boost status + pipeline version, so subsequent CI runs only regenerate items whose sprite or boost status actually changed.
- Fixed **boost detection** to handle bare-identifier keys (`Blossombeard`) and double-quoted keys containing apostrophes (`"Fisher's Hourglass"`) — earlier regex silently dropped both.
- Items are now centered at ~1/6 area of the 1920x1920 canvas, trimmed (static only) for consistent visual sizing, **nearest-neighbor** scaled to preserve pixel-art crispness.
- Metadata JSONs now have a **stable key order** (by explicit object construction) so diffs stay minimal across regenerations.
- Wearable JSONs: URL extension now detected from disk (`.webp` if present, else `.png`) — fixes a historical mismatch where some JSONs pointed at `.webp` URLs for files that were actually uploaded as `.png`.
- Workflow `.github/workflows/metadata.yml` now triggers on asset/script/composite-asset paths, supports `workflow_dispatch`, runs `yarn images` after `yarn metadata`, and commits both the WebPs and the manifest.

## Test plan

- [ ] Inspect `public/erc1155/images/462.webp` (Basic Scarecrow) — crisp, centered, boost overlay visible.
- [ ] Inspect `public/erc1155/images/414.webp` (Fountain) — should animate; every frame has the green bg behind it.
- [ ] Inspect `public/erc1155/images/421.webp` (Kuebiko) — 7-frame animation; no "too large for WebP" error regression.
- [ ] Inspect `public/erc1155/images/2077.webp` (Fisher's Hourglass) — boost overlay present (previously missed due to apostrophe regex bug).
- [ ] Sample JSON (e.g. `public/erc1155/462.json`) points at `.../462.webp`.
- [ ] Sample wearable JSON points at an extension that matches the file on disk.
- [ ] Load an affected token on OpenSea testnet and verify the image renders; spot-check an animated one (Fountain) to confirm frames play.
- [ ] Run `yarn images` locally — all 1393 items report "unchanged" when nothing has moved.
- [ ] Bump a single sprite in `src/assets`, run `yarn images` — only that item regenerates.


adamhannigan (approved):
Epic work 💪

PR #7172, opened 2026-04-24
[FEAT] Add greenhouse exterior activity indicator

alexandrdragin:
# Pull request description

Add an exterior activity indicator to the Greenhouse went something grow inside.

# Context / motivation

Players could not tell from the outside whether the Greenhouse was in use, which made it easy to forget planted grapes, olives, or rice and leave the building idle for long periods. This change reuses the existing smoke animation from the Greenhouse oil machine and shows it on the exterior while greenhouse crops are actively growing.

As a result:
- players had no quick visual reminder that the Greenhouse was currently occupied
- the building looked inactive even when production was underway

# Solution

Reuse the existing `SUNNYSIDE.building.smoke` animation on the exterior Greenhouse sprite whenever at least one greenhouse pot contains a planted crop.

The smoke was positioned and layered behind the Greenhouse art so it reads as part of the building rather than a UI overlay.

# Implementation

- adds a selector to detect whether any Greenhouse pot currently has an active plant
- renders the existing smoke animation on the Greenhouse exterior when active
- keeps the existing ready-to-harvest produce icons unchanged
- reuses an existing animation asset already used in the Greenhouse interior
- no smoke when there are no crops 
- no changes inside greenhouse
- no smoke when crops are ripe 

# Files

- `src/features/island/buildings/components/building/greenhouse/Greenhouse.tsx`

# Screenshots or screen recording

image
image
image

### DIFF
--- a/src/features/island/buildings/components/building/greenhouse/Greenhouse.tsx
+++ b/src/features/island/buildings/components/building/greenhouse/Greenhouse.tsx
@@ -13,6 +13,7 @@ import { GreenHouseCropName } from "features/game/types/crops";
 import { GreenHouseFruitName } from "features/game/types/fruits";
 import { ITEM_DETAILS } from "features/game/types/images";
 import { GREENHOUSE_VARIANTS } from "features/island/lib/alternateArt";
+import { SUNNYSIDE } from "assets/sunnyside";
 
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):
Oooh I love this 🏆

PR #7179, opened 2026-04-28
[FEAT] Pixel movements

adamhannigan:
# Pull request description

Allow players to move items by smaller pixel amounts. This will help them align + create new creative designs.

Screenshot 2026-04-28 at 3 05 15 pm

We have opted for new fields offset X (oX) and offset Y (oY) so we don't break existing functional logic that expects integer based on coordinates.

_This is just enabled for collectibles, buds, pets & bumpkins for now._

## How to test?

1. Open landscaping mode
2. Click on a collectible, then click the new pixel perfect icon
3. Click the arrows or use the WASD keys to move it to perfect location


c-py (changes requested):
Faced a few issues moving collectibles using cursor:

1. The up arrow is awkwardly behind this div that contains the pixel perfect icon. Its blocking my click on the up arrow:

Image

2. It seems when the collectible is at `oY` is at `-8`, when I try to move the collectible back up, I can only go up `8` pixels to `oY` 0. I should be able move up 16 pixels.

3. It seems like the save doesn't trigger unless I move the collectible all `8` pixels - haven't been able to reliably figure out what causes the save to actually send, but this can be replicated by moving a collectible down one pixel only by clicking the arrow icons.

PR #7190, opened 2026-04-29
[FIX] obsidian shrine default tab

alexandrdragin:
# Pull request description

@eliasSFL 

Hi, I need your opinion. Today, working with Obsidian Shrine, I have stumbled over a default state that seems to not exist.

telegram-cloud-photo-size-2-5444970598422484092-y

 As I understood, this is due to the fact that the Fertilize tab will only run with a new chapter. As Fix this state I made the simplest fallback , but I'm not very sure if it's worth fixing it if, in my opinion, this case will no longer be there next week. 
 
 What do you say? 

## Summary
- Fix Obsidian Shrine opening with no active tab when crops are planted but not ready to harvest.
- Only default to the `Fertilise` tab when the bulk fertilise feature is actually available.
- Fall back to `Harvest` so players see the existing “No crops ready to harvest” state instead of an empty panel.

## Problem
When all crop plots are occupied by growing crops, there are no ready crops and no available plots. The shrine previously selected `fertilise` as the default tab in that state. If the fertilise feature flag was disabled, that tab was not rendered, leaving no visible active tab.

## Summary by CodeRabbit

* **Improvements**
  * Obsidian Shrine bulk fertilisation feature is now universally accessible to all users.

### DIFF
--- a/src/features/island/collectibles/components/ObsidianShrine.tsx
+++ b/src/features/island/collectibles/components/ObsidianShrine.tsx
@@ -236,7 +236,13 @@ export const ObsidianShrine: React.FC<CollectibleProps> = ({
     );
     setShow(true);
     setActiveTab(
-      hasReadyCrops ? "harvest" : hasAvailablePlots ? "plant" : "fertilise",
+      hasReadyCrops
+        ? "harvest"
+        : hasAvailablePlots
+          ? "plant"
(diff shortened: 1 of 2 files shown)


eliasSFL (approved):
Thanks for catching the bug! I've updated to release the fertilise tab for the public

PR #7229, opened 2026-05-05
[FIX] Correct marketplace pet card levels

alexandrdragin:
## Root cause

`ListViewCard` renders pet level from `details.experience`. In the general marketplace collection flow, pet rows already pass `item.experience`, so levels render correctly. In `My Collection` and local trade/profile contexts, `getTradeableDisplay` was called without `experience`, so the card fell back to `getPetLevel(0)` and always displayed level 1.

The detail page looked correct because it uses `tradeable.experience` from the detail payload.

image

image

## Summary

- Fixes marketplace pet cards showing `Level: 1` in profile collection/trade contexts when the pet has a higher real level.
- Adds a fallback in `getTradeableDisplay` so pet cards use the owned pet XP from `gameState.pets.nfts[id].experience` when no explicit marketplace `experience` is provided.
- Keeps marketplace API experience as the primary source when it is available.
- Adds unit coverage for both fallback behavior and explicit marketplace experience precedence.

## Impact

Players now see the correct pet level directly on marketplace pet cards in profile collection and trade-related views, without opening the detail page.

## Validation

- `npx eslint --quiet src/features/marketplace/lib/tradeables.ts src/features/marketplace/lib/tradeables.test.ts`
- `git diff --check`

Note: local `jest` could not run because this checkout is missing `esbuild-runner/jest` in `node_modules`; the added test is included for CI where dependencies are installed normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected marketplace pet display logic so experience shown uses an explicitly set marketplace value when present and falls back to the pet's stored experience when not.

* **Tests**
  * Added test coverage validating both explicit marketplace experience prioritization and proper fallback to stored pet experience.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sunflower-land/sunflower-land/pull/7229)


eliasSFL (approved):
LGTM. Pushed a follow-up to make the new test suite runnable:

- Added global mocks in `test/setup.ts` for `useSound`, `SeasonalSeeds`, and `metadata/metadata` so importing `getTradeableDisplay` no longer drags in the SOUNDS / GameProvider / wagmi / generated-metadata chain.
- Strengthened the fallback assertion to check `display.experience === experience` (exact XP), in addition to the level comparison.
- Full suite: 270/270 passing.

PR #7231, opened 2026-05-06
[CHORE] Adjust trade limit for salt

Xeko-92:
For 3 wood (1 salt rake) people are currently getting up to 19 salt. With 6 nodes that's over the trade limit for salt in 1 set of recharges. The limit for salt should be at least the same as the limit for wood.

The salt price also keeps dropping (which is fine and expected), so the price for a maximum listing of salt is 1,8 FLOWER.

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the maximum Salt trade limit from 100 to 500 units, enabling larger trading quantities and providing greater flexibility in managing in-game resources.


eliasSFL (approved):
LGTM. Mirrored on BE in tradeLimits.ts (Salt: 100 → 500).

PR #7273, opened 2026-05-19
[FIX] Correct beehive honey readiness UI

alexandrdragin:
## Summary
Fixes a beehive UI desync where hives could appear visually full and show the Bee Swarm dialog even when the actual harvested honey amount was partial. The UI now syncs beehive readiness from the latest hive state and only shows the Bee Swarm modal when the swarm was actually consumed by the harvest event.

## Context / motivation
I reported case where a beehive looked full, but harvesting returned less than one full honey jar, for example `+0.43`. In the same flow, the Bee Swarm modal could appear even though crop plots were not pollinated.

telegram-cloud-photo-size-2-5217442609972846254-y
telegram-cloud-photo-size-2-5217442609972846255-y
telegram-cloud-photo-size-2-5217442609972846256-y

The game event logic is already correct: `harvestBeehive` only applies swarm pollination when `honeyProduced >= 1`. The issue was in the local beehive UI state machine. Once it entered `honeyReady`, it did not handle `UPDATE_HIVE`, so the visual ready state could stay stale after the underlying hive data changed.

## How to test
1. Open a farm with an active beehive that has a swarm and partial honey production.
2. Confirm the hive does not visually appear as full when synced hive data is below one full honey jar.
3. Harvest partial honey and confirm the Bee Swarm modal does not appear.
4. Harvest a genuinely full hive with a swarm and confirm the Bee Swarm modal still appears and crop plots receive the swarm state.
5. Run `npm run build -- --mode development`.

Note: the focused Jest command could not run in this local checkout because `esbuild-runner/jest` is missing from `node_modules`. A regression test was added for the beehive state-machine sync path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed harvest flow so the honey modal closes immediately and swarm alerts appear only when a swarm actually ends.

* **Refactor**
  * Improved beehive state synchronization and update flow to keep hive, honey production, and attached-flower status consistent.

* **Tests**
  * Expanded unit tests to cover multiple beehive state scenarios and transitions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sunflower-land/sunflower-land/pull/7273?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/features/game/expansion/components/resources/beehive/Beehive.tsx
+++ b/src/features/game/expansion/components/resources/beehive/Beehive.tsx
@@ -117,15 +117,18 @@ export const Beehive: React.FC<Props> = ({ id }) => {
   }, [honeyReady, beehiveService]);
 
   const handleHarvestHoney = () => {
-    if (hive.swarm && honeyReady) {
+    setShowHoneyLevelModal(false);
+
+    const hadSwarm = hive.swarm;
+    const state = gameService.send("beehive.harvested", { id });
+    const updatedHive = state.context.state.beehives[id];
+
+    if (hadSwarm && updatedHive?.swarm === false) {
       setShowSwarmModal(true);
     }
 
-    setShowHoneyLevelModal(false);
-
-    const state = gameService.send("beehive.harvested", { id });
     beehiveService.send("HARVEST_HONEY", {
-      updatedHive: state.context.state.beehives[id],
+      updatedHive,
     });
   };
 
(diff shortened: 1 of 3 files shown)


eliasSFL (approved):
Audited the PR: root cause correctly identified (UI state-machine, not the reducer), fix matches the reducer's contract on swarm consumption, and the latent staleness in updateHive is also addressed. Merge is clean against main. LGTM.

PR #7288, opened 2026-05-25
fix(BulkSellModal): prevent 50% button from generating invalid sell state

Radovenchyk:
Fix 50% button silently disabling Sell when itemAmount rounds to zero after precision clamp, unify disabled logic with safeHalf computation, fix coinAmount formatting and invalid Tailwind class.

## Summary by CodeRabbit

* **Bug Fixes**
  * The 50% quick-select button is now disabled when choosing it would produce an invalid or too-small amount.
  * Displayed "you will receive" amounts are now formatted with improved precision for clearer receipts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sunflower-land/sunflower-land/pull/7288?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/ui/BulkSellModal.tsx
+++ b/src/components/ui/BulkSellModal.tsx
@@ -44,7 +44,15 @@ export const BulkSellModal: React.FC<BulkSellProps> = ({
     maxDecimalPlaces,
   );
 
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):
Good catch 👏 Thanks for the contribution

PR #7291, opened 2026-05-26
[FEAT] Add marketplace favorites

alexandrdragin:
## What changed

Adds a client-side `Favorites` section to the marketplace so players can build a personal list of items they care about without backend changes.

## Logic

- Favorites are stored locally per host and farm id, with an in-memory fallback for environments where `localStorage` is unavailable.
- Item detail pages can toggle the current marketplace item in or out of Favorites.
- The toggle is only available on item detail pages, not on compact collection cards.
- The marketplace sidebar now includes a `Favorites` route with the active heart icon.
- The Favorites route reuses existing marketplace collection data and filters it by the local favorite keys.
- Merged marketplace rows are deduplicated by collection/economy/id so resources do not appear twice when both `resources` and `collectibles` are loaded.
- Favorites are ordered by category rank, then by display name: Resources, Limited, Power ups, Cosmetics, BUD NFTs, Pets.
- Favorites cards use an isolated visual variant so the mixed Favorites grid can display resources, wearables, BUDs, and Pets more consistently without changing the rest of the marketplace.

## Screenshots

image
image
image
image

## UI notes

- The detail-page toggle uses the existing button treatment with a protruding heart icon.
- Visible copy is shortened to `Add to` / `Remove from`, with the heart carrying the Favorites meaning.
- The full `add/remove from favorites` copy remains available as the button title.
- The Cosmetics marketplace sidebar icon was changed from the heart to Basic Bear to avoid conflicting with Favorites.

## Linked pr
https://github.com/sunflower-land/sunflower-land/pull/7191

## Validation

- `npx tsc --noEmit`
- Focused ESLint on changed marketplace components during implementation
- Husky pre-commit checks (`eslint --fix`, `prettier -u --write`)

## Summary by CodeRabbit

* **New Features**
  * Marketplace favorites: toggle favorites, persistent storage, and a Favorites-only view with deduplication, custom ranking, and adjusted card sizing.

* **User Experience**
  * Favorites filter and route, heart-icon empty state with conditional “browse marketplace” CTA, and favorite-styled cards/images across item panels.

* **Localization**
  * Added i18n strings for Favorites.

* **Accessibility**
  * Buttons now accept aria-labels for improved screen-reader support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sunflower-land/sunflower-land/pull/7291?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #7314, opened 2026-06-04
[microFIX] Keep marketplace favorite toggle below header

alexandrdragin:
## Summary

telegram-cloud-photo-size-2-5265128729917201278-y

Fixes the marketplace item detail favourite heart overlapping the mobile back/header panel while the page is scrolled.

## Context / motivation
The favourite toggle added in the Marketplace Favorites PR was positioned inside the tradeable image panel with the same z-index level as the sticky item header. On mobile, when the detail page scrolls, the heart icon could visually sit above the back/navigation panel instead of staying below it with the card content.

## Implementation
- Removed the elevated z-index from the favourite heart button inside TradeableInfo.
- Kept the temporary favourite feedback label above the image locally with a low z-index value so it still appears over the image but remains below sticky navigation/header layers.
- No marketplace favorites logic changed; this is a layering-only UI fix.

## How to test
1. Open Marketplace on mobile viewport.
2. Open any item detail page, for example a collectible.
3. Ensure the favourite heart appears in the image panel.
4. Scroll the page so the sticky back/header panel overlaps the image area.
5. Confirm the favourite heart no longer covers the back/header panel.
6. Toggle the favourite state and confirm the feedback label still appears over the image.

Required for this visual fix. Add the provided mobile screenshot as the before state and a new after screenshot after verification.

## Related issues
Follow-up to #7291.

## Summary by CodeRabbit

* **Style**
  * Adjusted UI layering for the marketplace favorite button and feedback label to ensure correct visual stacking order.

### DIFF
--- a/src/features/marketplace/components/TradeableInfo.tsx
+++ b/src/features/marketplace/components/TradeableInfo.tsx
@@ -206,7 +206,7 @@ export const TradeableImage: React.FC<{
             type="button"
             aria-label={favoriteLabel}
             title={favoriteLabel}
-            className="absolute top-4 right-4 z-10 h-8 w-8 cursor-pointer border-0 bg-transparent p-0 hover:img-highlight"
+            className="absolute top-4 right-4 h-8 w-8 cursor-pointer border-0 bg-transparent p-0 hover:img-highlight"
             onClick={(event) => {
               event.stopPropagation();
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):

adamhannigan:
Thanks @alexandrdragin 👏 

Just a heads up, please use [FIX] - this helps with some internal categorisation

PR #7262, opened 2026-05-13
[FIX] Flower grouping inside new quick drag panel

alexandrdragin:
## Summary

This is a follow-up fix for the flower chest grouping introduced in https://github.com/sunflower-land/sunflower-land/pull/7198.

Same logic for new quick drag-and-drop panel

image

## Changes

— Common flower-group at helper: inventory.ts (line 50)
— Chest now use that helper: Chest.tsx (line 367)
— Inside quick drag-and-drop create panel Flowers, and Decorations и Boosts no longer include flowers: LandscapingQuickPanel.tsx (line 371)

## Testing

- Checked the branch diff is isolated to the inventory grouping and quick drag panel files.
- Ran git diff --check successfully.
- Ran yarn tsc; it only fails on an existing unrelated BeachScene.ts replaceAll lib error on current main, with no new errors from this change.
- Mirrored the change into the local DevStand and verified the dev preset loads with flower preview items for visual checking.

## Support requests
https://discord.com/channels/880987707214544966/1497003604257673380/1501778099963170896

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated "Flowers" tab to the Landscaping quick panel for browsing and placing flowers.
  * Chest now provides a unified flower list used by the UI.

* **Improvements**
  * Improved flower selection and sorting to prevent duplicates across groups and ensure consistent ordering.
  * Added localized empty-state text for the flowers panel.

PR #7198, opened 2026-04-30
[CHORE] Improve chest flower grouping

alexandrdragin:
# Pull request description

I think it's a nice addition to this PR https://github.com/sunflower-land/sunflower-land/pull/7195/changes

## Problem

After flowers became placeable decorations, they appeared mixed into the general Decorations section. Since there are many flower variants, a separate label makes the chest easier to scan without changing placement or inventory logic.

image

## Solution
- Adds a dedicated Flowers group in the chest inventory for placeable flower collectibles.
- Keeps the existing Decorations group, then renders Flowers after it for easier navigation.
- Uses the existing FLOWERS registry and seedling label icon, so flower-themed SFT decorations such as Flower Cart/Rug remain in Decorations.

## badge preview

image

## Summary by CodeRabbit

* **New Features**
  * Chest inventory now includes a dedicated "Flowers" category with its own icon for easier discovery.

* **Improvements**
  * Flower collectibles are removed from decoration and boost groups and shown separately in a consistent, prioritized order with locale-aware fallback for sorting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sunflower-land/sunflower-land/pull/7198)

PR #7316, opened 2026-06-05
[FIX] Boosted flower separate from decoration 

alexandrdragin:
Fix to https://github.com/sunflower-land/sunflower-land/pull/7198 and https://github.com/sunflower-land/sunflower-land/pull/7262

## Problem
After flower grouping was added to the chest and quick drag-and-drop panel, flowers with gameplay boosts were grouped under Flowers instead of Boosts.

## Solution
- Keep using the existing boost source of truth: COLLECTIBLE_BUFF_LABELS.
- Exclude boosted flowers from the Flowers group.
- Let those same items flow into the existing Boosts group logic in both the chest and quick drag panel.

This avoids maintaining a separate manual list of boosted flowers and keeps future boosted flowers categorized automatically when they receive buff labels.

## Testing
- git diff --check
- yarn tsc

## Summary by CodeRabbit

* **Refactor**
  * Centralized inventory item categorization logic to ensure consistent classification of boosts across the island interface. Items are now consistently identified as boosts or flowers based on a shared determination method.

### DIFF
--- a/src/features/island/hud/components/LandscapingQuickPanel.tsx
+++ b/src/features/island/hud/components/LandscapingQuickPanel.tsx
@@ -17,7 +17,6 @@ import {
   placeEvent,
 } from "features/game/expansion/placeable/landscapingMachine";
 import type { PlaceableLocation } from "features/game/types/collectibles";
-import type { InventoryItemName } from "features/game/types/game";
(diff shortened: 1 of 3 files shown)


eliasSFL (approved):
LGTM! Thanks @alexandrdragin

PR #7334, opened 2026-06-12
[FEAT] UX improve island viewport behavior after inside scenes

alexandrdragin:
## Problem

Currently, when a player exits a building with an inside scene, the island viewport is centered back on the middle of the island via `scrollIntoView(Section.GenesisBlock, "auto")`.

This is a small but noticeable friction point when buildings are placed near the edges of the island: after leaving the inside scene, the player is no longer looking at the place they opened the building from.

It is especially frustrating during friend cleanup flows. For example, if a friend has a Pet House near a small island with gutter, opening the Pet House for cleanup and then exiting sends the camera back to the island center instead of preserving the cleanup area.

## What changed

This PR saves the island viewport position before entering inside scenes and restores it when returning to the island. That makes the interaction feel smoother because the player returns to the same viewport position they had before opening the building.

The behavior works for both:

- the player’s own main island
- visiting/friend cleanup flows, including opening a friend’s Pet House

Scroll positions are stored separately for the player’s own island and visited farms, so viewport state does not leak between different islands.

## Tested offline

Tested on dev farm configuration with a Pet House with `Barkley` placed inside

https://github.com/user-attachments/assets/7f75274f-fc49-4f6d-bab1-dab04dc3ee36

## Validation

- `npx eslint --quiet ...`
- `curl -I http://localhost:3000/` -> `HTTP/1.1 200 OK`

### DIFF
--- a/src/features/game/expansion/Land.tsx
+++ b/src/features/game/expansion/Land.tsx
@@ -44,6 +44,7 @@ import {
   getSortedResourcePositions,
   getSortedCollectiblePositions,
 } from "./lib/utils";
+import { restoreIslandScrollPosition } from "./lib/islandScroll";
 import { Clutter } from "features/island/clutter/Clutter";
 import { PetNFT } from "features/island/pets/PetNFT";
 import { WaterTrapSpot } from "features/island/fisherman/WaterTrapSpot";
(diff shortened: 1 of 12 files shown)


alexandrdragin:
Superseded by #7335 with the corrected branch name.

PR #7337, opened 2026-06-13
[FIX] Keep Karkinos boost label visible

alexandrdragin:
## Summary
Fixes Karkinos boost communication so it behaves consistently with other collectibles whose effects can be overridden by stronger collectibles. Karkinos now always displays its boost label and its conflict warning, keeping it grouped as a Boost item instead of falling back into Decorations when Cabbage Boy is active.

## Context / motivation

Karkinos still has an active gameplay effect: it gives `+0.1 Cabbage` when placed and ready, unless Cabbage Boy is active. The issue was UI communication. When Cabbage Boy was present, Karkinos returned an empty buff-label array, so players could see only the flavor description and the chest/quick panel could classify it as a decoration rather than a boost item.

image

it's affected marketplace description
image

Other collectibles with overlapping effects do not hide all buff labels. For example, Scarecrow, Nancy, Woody the Beaver, and Apprentice Beaver continue to show their boost information plus a warning such as `Disabled if Kuebiko Active` or `Disabled if Foreman Beaver Active`. This PR brings Karkinos in line with that existing pattern.

image

image

## What changed

- Karkinos now always shows `+0.1 Cabbage`.
- Karkinos now always shows `Disabled if Cabbage Boy Active`.
- The gameplay logic is unchanged: Cabbage Boy still takes priority over Karkinos during cabbage harvests.
- Because Karkinos now always has buff labels, it remains in Boosts sections instead of being grouped with Decorations when Cabbage Boy is active.

## How to test

1. Open the chest or landscaping quick panel with Karkinos in inventory and no Cabbage Boy placed.
2. Confirm Karkinos appears under Boosts and shows `+0.1 Cabbage` plus the warning label.
3. Place/activate Cabbage Boy.
4. Confirm Karkinos still appears under Boosts and still shows the warning.
5. Harvest cabbage with both active and confirm gameplay still follows the existing priority: Cabbage Boy applies instead of Karkinos.

## Summary by CodeRabbit

* **Bug Fixes**
  * Karkinos collectible item now consistently displays its power-up and warning information with the boosted item icon, independent of other collectibles in your collection.

### DIFF
--- a/src/features/game/types/collectibleItemBuffs.ts
+++ b/src/features/game/types/collectibleItemBuffs.ts
@@ -295,21 +295,17 @@ export const COLLECTIBLE_BUFF_LABELS: Partial<
       boostedItemIcon: CROP_LIFECYCLE["Basic Biome"].Cabbage.crop,
     },
   ],
-  Karkinos: (game) => [
-    ...(isCollectibleBuilt({ game, name: "Cabbage Boy" })
-      ? []
-      : ([
-          {
-            shortDescription: translate("description.Karkinos.boost"),
-            labelType: "success",
-            boostTypeIcon: powerup,
-            boostedItemIcon: CROP_LIFECYCLE["Basic Biome"].Cabbage.crop,
-          },
-          {
-            shortDescription: translate("description.Karkinos.warning"),
-            labelType: "danger",
-          },
-        ] as BuffLabel[])),
+  Karkinos: () => [
+    {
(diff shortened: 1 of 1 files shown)


Dionis404:
It seems to me that the logic is very correct, if Cabbage Boy is installed on the farm, then Karkinos becomes a decoration and should not be in the bonuses section, because it does not give any bonuses.

c-py (approved):
LGTM! Thanks @alexandrdragin !

PR #7340, opened 2026-06-16
[FEAT] Reel pack pricing: linear ramp then double

eliasSFL:
## What

Reworks the gem cost of fishing reel packs (extra daily casts).

The first four reel packs bought each day now ramp **linearly** — 5, 10, 15, 20 Gems — then **double** for every pack after (40, 80, 160, 320…). Previously every pack doubled from a base of 10 (10, 20, 40, 80…), making early extra reels noticeably more expensive.

| Pack # | Old | New |
|---|---|---|
| 1 | 10 | 5 |
| 2 | 20 | 10 |
| 3 | 40 | 15 |
| 4 | 80 | 20 |
| 5 | 160 | 40 |
| 6 | 320 | 80 |
| 7 | 640 | 160 |

## How

- Extracts a pure `getReelPackPrice(timesBought)` curve from `getReelsPackGemPrice`; the aggregator still sums the curve across packs bought together.
- Mirrors the change on the backend (sunflower-land-api PR to follow).
- Updates the existing gem-balance assertions and adds tests for the linear ramp, the double past the fourth pack, and the full daily cumulative (2570 for all 10 packs).

The fisherman UI and i18n need no changes — the displayed price is already fully dynamic.

## Summary by CodeRabbit

* **Updates**
  * Updated reel-pack gem pricing to use a new model: linear scaling for the first four purchases, then exponential doubling for additional packs.
  * Pricing now correctly reflects the day of the cast action when calculating pack costs.
* **Tests**
  * Expanded and adjusted unit tests to validate the updated reel-pack pricing curve and multi-pack purchase totals.

### DIFF
--- a/src/features/game/events/landExpansion/castRod.test.ts
+++ b/src/features/game/events/landExpansion/castRod.test.ts
@@ -3,9 +3,9 @@ import {
   INITIAL_FARM,
   TEST_FARM,
 } from "features/game/lib/constants";
-import { castRod, getReelsPackGemPrice } from "./castRod";
+import { castRod, getReelPackPrice, getReelsPackGemPrice } from "./castRod";
 import Decimal from "decimal.js-light";
-import type { Bumpkin } from "features/game/types/game";
+import type { Bumpkin, GameState } from "features/game/types/game";
 import { type Chum, getDailyFishingLimit } from "features/game/types/fishing";
 import { CHAPTERS } from "features/game/types/chapters";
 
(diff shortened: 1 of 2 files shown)


eliasSFL:
Thanks for the reviews. Addressed:

- **CodeRabbit [Critical] — `getReelsPackGemPrice` missing `createdAt`:** Fixed in 17a2eb787. `castRod` derives `today` from `createdAt`, but the price lookup was defaulting to `Date.now()`, so `timesBought[today]` could read the wrong day. The backend already passed `createdAt`, so this just brings FE in line.

Deliberately not changing:
- **[Low] defensive integer guard in `getReelPackPrice`:** `timesBought` always comes from controlled state (`timesBought[today] ?? 0` plus a loop index), and event helpers here don't add such guards — keeping it consistent with the surrounding code.
- **[Medium] ship FE + BE together:** Intentional — backend mirror is sunflower-land/sunflower-land-api#3434; they should land/deploy together.
- **[Low] `getReelPackPrice` singular naming:** Intentional — it's the price of *one* pack, vs `getReelsPackGemPrice` for the summed total of N packs.

PR #7347, opened 2026-06-17
[FIX] Airdrop missing resources (incl. sunstones) on expansion

eliasSFL:
Mirrors the server-side `getRewards` "missing nodes" airdrop to the client. This logic has been **API-only since Feb 2024** and was never ported here, so the client's optimistic state diverged from the server on expansion.

Companion to the API fix: sunflower-land/sunflower-land-api#3440.

## What changed

On expansion, `getRewards` now grants the resource nodes a player is missing from previous expansions — e.g. nodes added to a layout after they already expanded past it — matching the API exactly (`getExpectedResources` − inventory, airdropped as `missing-resources-<n>`).

## Sunstones

Sunstones are finite (a rock is deleted from inventory once `minesLeft` hits 0), so a naive `expected − inventory` would re-grant rocks a player mined to depletion. Instead of skipping sunstones entirely (which denied them to players who never received any — the originally reported bug), we subtract the rocks actually consumed:

```ts
const depleted = Math.floor((game.farmActivity?.["Sunstone Mined"] ?? 0) / 10);
missing -= depleted;
```

`floor(Sunstone Mined / 10)` is a safe upper bound on depleted rocks (each rock holds 10 mines). Never-mined → full grant; depleted rocks are never re-granted; partial mining (<10) doesn't reduce the grant. `Sunstone Mined` is complete since sunstones shipped (and reconciled across the bumpkin→farm activity move), so the input is reliable.

## Tests

31 passing in `revealLand.test.ts`:
- never-mined player gets a positive sunstone airdrop
- 20 mines → 2 fewer rocks; 9 mines → no reduction
- desert 24→25 airdrops the full 6 missing sunstones
- updated the existing "no more rewards" case to hold full expected nodes (it now correctly produces no missing-resources airdrop)

## Summary by CodeRabbit

* **Bug Fixes**
  * Land expansion now grants missing resources via “missing-resources” airdrops when expanding into island types whose requirements haven’t been mined yet, including sunstone scenarios.
  * Reward calculations now avoid duplicate grants by accounting for already-depleted rocks and any pending, unclaimed missing-resource airdrops (without impacting forged node rewards).

* **Tests**
  * Added regression helpers and coverage for never-mined players, sunstone depletion behavior, a desert expansion transition case, and deduplication with pending missing-resource airdrops.

### DIFF
--- a/.github/workflows/codex.yml
+++ b/.github/workflows/codex.yml
@@ -1,7 +1,7 @@
 name: Perform a code review when a pull request is created
 on:
   pull_request:
-    types: [opened]
+    types: [opened, synchronize]
 
 jobs:
   codex:
--- a/.github/workflows/codex.yml
+++ b/.github/workflows/codex.yml
@@ -1,7 +1,7 @@
 name: Perform a code review when a pull request is created
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [opened]
 
 jobs:
   codex:
--- a/src/features/game/events/landExpansion/revealLand.test.ts
+++ b/src/features/game/events/landExpansion/revealLand.test.ts
@@ -857,7 +857,7 @@ describe("revealLand", () => {
       );
 
     const baseline = grant({}, 0);
-    expect(baseline).toBeGreaterThan(0);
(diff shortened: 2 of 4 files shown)


eliasSFL:
@coderabbitai review

PR #7359, opened 2026-06-20
[FEAT] Add marketplace collection filters

alexandrdragin:
## Summary
This PR adds a dedicated filter toggle from Basket to the Marketplace `My Collection` view and introduces category chips for faster navigation through a player's collection. The user-facing result is a much shorter path to the relevant part of a large collection without changing the underlying collection data model.

image

## Context / motivation
The collection view can grow large very quickly, especially for active players with buds, pets, wearables, cosmetics, buildings, and boost items all mixed into one long page. The goal here was to make browsing and searching easier by reusing the existing filter interaction pattern from the basket/chest flow and exposing the same category structure directly inside Marketplace `My Collection`.

!!!

**During implementation we also explored a `Not for sale` split, but did not ship it. The reason is architectural rather than visual: tradability data currently comes from multiple inconsistent sources, so the UI cannot reliably derive a single correct `Not for sale` status for every item in the collection list.**

**Current tradability inputs are split across:
- local release windows such as `tradeAt` / `withdrawAt` in `src/features/game/types/withdrawables.ts`
- marketplace API activity data
- separate detail-page logic in `src/features/marketplace/components/TradeableInfo.tsx`
- collection grouping built from local farm state rather than one normalized tradability source**

**Because of that, a stable `Not for sale` category needs architectural cleanup in the data layer, not another UI-only patch.**

## Summary by CodeRabbit

* **New Features**
  * Added multi-select category filtering for collection items
  * New cosmetics and buildings sections in collection view
  * Removed time-based restrictions on pet and wearable availability

* **Improvements**
  * Collection items now sorted alphabetically
  * Updated filter interface for better usability

### DIFF
--- a/src/features/island/hud/components/inventory/InventoryFilters.tsx
+++ b/src/features/island/hud/components/inventory/InventoryFilters.tsx
@@ -76,19 +76,17 @@ export const InventoryFilters: React.FC<Props> = ({
           >
             {t("inventory.all")}
           </Chip>
-          {categories
-            .sort((a, b) => a.label.localeCompare(b.label))
-            .map((category) => (
-              <Chip
-                key={category.id}
-                selected={activeCategories.includes(category.id)}
-                icon={category.icon}
-                onClick={() => onToggleCategory(category.id)}
-                className="whitespace-nowrap sm:whitespace-normal mb-1 sm:mb-0"
-              >
-                {category.label}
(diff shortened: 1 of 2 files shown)


adamhannigan (approved):
Nice addition @alexandrdragin 👏 Much easier now to find items

PR #7320, opened 2026-06-06
[CHORE] Show cheered players in social feed

alexandrdragin:
image

 Summary
Adds a visual Rocket indicator to Social Feed cheer events when the player(you) has already cheered that farm today. 

##Context / Motivation
Help events already show a hand icon when the player has also helped that farm today, making reciprocal help easy to spot. Cheer events did not have an equivalent indicator, so players could not quickly tell which farms they had already cheered back from the feed.

What Changed

Reads today’s socialFarming.cheersGiven.farms from game state.
Shows the cheer icon on cheer feed rows when the sender’s farm ID is in today’s cheered farms list.
Refreshes feed data after farm.cheered succeeds, including both own-farm and visiting-farm cheer flows.
Keeps the icon non-interactive and prevents clicks on it from opening the player details modal.

## Summary by CodeRabbit

* **New Features**
  * Feed interactions now display a visual indicator when you've cheered that farm, helping you track your daily support activities.

PR #7369, opened 2026-06-24
[FIX] Show cheer feed icon for today only

alexandrdragin:
Just my bad 

after this pr https://github.com/sunflower-land/sunflower-land/pull/7320
image

## Problem
Cheer activity in the Social Feed was shown on older feed entries for the same player, even when the cheer happened today. This made historical feed rows look like they also had a recent cheer event, which was misleading.

## Why this differs from Help
Help currently uses helpedThemToday as a “today relationship status” badge. That logic was left unchanged.
Cheer is different because the feed already has a real interaction type: type: "cheer". A cheer should be treated as an activity event with its own timestamp, not as a reusable player status across older rows.

## Fix
The feed **now filters cheer interactions by the current UTC date before rendering**. This keeps today’s cheers visible, but prevents cheer activity from appearing on previous days’ feed entries. Other interaction types, including Help, are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Feed items now show more accurately based on the current UTC day for “cheer” activity.
  * The empty-state message now appears only when there are no visible items, not just no raw feed items.
  * Pagination now updates correctly when filtered feed items change.

### DIFF
--- a/src/features/social/Feed.tsx
+++ b/src/features/social/Feed.tsx
@@ -9,6 +9,7 @@ import React, {
   useContext,
   useEffect,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
 } from "react";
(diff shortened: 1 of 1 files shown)


eliasSFL (approved):
LGTM! Ahh I saw this and was about to fix it

PR #7373, opened 2026-06-25
[FIX] marketplace profile section layout

alexandrdragin:
## Summary
- Let Active Trades offers and listings fill the available marketplace profile space and split it evenly when both sections are present.
telegram-cloud-photo-size-2-5325749672154439030-y

- Keep existing offer/listing layouts unchanged outside the full-height marketplace route.
- Make the profile Collection route use an internal scroll container so collection cards stay inside the marketplace window.
telegram-cloud-photo-size-2-5325749672154439033-y

## Problem
After the profile marketplace cleanup, the Active Trades tab still used the previous capped offer/listing panel heights. This left unused vertical space even though other profile content had been removed. The Collection tab also rendered long content beyond the marketplace bounds without a working internal scroll area.

## Solution
- Add an optional full-height mode to MyOffers, MyListings, and MyCollection.
- Use that mode from the marketplace profile routes only.
- Rework MyTrades into a flex column so offers/listings can share the available height instead of relying on fixed max heights.

## Summary by CodeRabbit

* **New Features**
  * Improved the profile collection, listings, and offers views with a full-height layout option for a better fit on larger screens.
  * Updated the trades page to display offers and listings only when they’re available, creating a cleaner, more relevant view.

* **Bug Fixes**
  * Refined scrolling and spacing in marketplace profile panels for a more consistent browsing experience.

### DIFF
--- a/src/features/marketplace/components/home/MarketplaceHome.tsx
+++ b/src/features/marketplace/components/home/MarketplaceHome.tsx
@@ -245,7 +245,7 @@ export const MarketplaceNavigation: React.FC = () => {
               />
               <Route
                 path="/profile/:id/collection"
-                element={<MyCollection />}
+                element={<MyCollection fullHeight />}
               />
               <Route path="/profile/rewards" element={<MarketplaceRewards />} />
               {/* default to hot */}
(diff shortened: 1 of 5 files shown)


eliasSFL (approved):
LGTM! Thanks for your contributions!

PR #7377, opened 2026-06-26
[FIX] Restore search results when navigating back from marketplace item

Dionis404:
## Summary

- **Bug**: After searching for an item in the marketplace, opening it, and pressing Back — the player was returned to the main marketplace screen instead of search results.
- **Root cause**: Search state was stored in React `useState`. The `onNavigated` callback cleared the search immediately on item click, so by the time `Tradeable` navigated back, the search was gone.
- **Fix**: Moved search state into the URL param `?q=`, so the `backRoute` captured by `Collection` naturally includes the query string. Navigating back restores the search automatically.
- **Bonus**: Added a clear (✕) button to the search field by wiring up the existing `onCancel` prop on `TextInput`.

## Changes

- `MarketplaceHome.tsx` — replaced `useState` for search with `useSearchParams` (`?q=` param); removed the now-unnecessary `onNavigated` callback.
- `MarketplaceSearch.tsx` — pass `onCancel` to `TextInput` to show the clear button when input is non-empty.

https://github.com/user-attachments/assets/c3980668-60c0-4da9-b578-211f538f20ff
## Test plan

- [ ] Search for an item in the marketplace
- [ ] Open an item from search results
- [ ] Press Back — verify search results are restored with the original query
- [ ] Verify the ✕ button appears while typing and clears the field on click
- [ ] Verify normal marketplace navigation (tabs, filters) is unaffected

### DIFF
--- a/src/features/marketplace/components/MarketplaceSearch.tsx
+++ b/src/features/marketplace/components/MarketplaceSearch.tsx
@@ -14,6 +14,7 @@ export const MarketplaceSearch: React.FC<{
           icon={SUNNYSIDE.icons.search}
           value={search}
           onValueChange={setSearch}
+          onCancel={() => setSearch("")}
         />
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


eliasSFL (approved):
LGTM!

PR #7386, opened 2026-06-29
[FIX] Route delivery Old Salty and Pharaoh to Digging

alexandrdragin:
## Summary
Updates the delivery order location for Old Salty and Pharaoh from Beach to Digging. Their order cards now show Digging and the travel button opens the Digging entry point instead of the regular Beach spawn.

## Context / motivation
A new Digging transition point exists on the world map, but delivery orders for nearby NPCs still sent players to Beach. This made the CTA misleading for orders tied to digging-focused characters.

## How to test
1. Open the deliveries panel with an active Old Salty or Pharaoh order.
2. Confirm the location badge shows Digging.
3. Click the travel button.
4. Confirm the world opens through the Digging entry point rather than the default Beach spawn.

image
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Orders involving digging NPCs now show the correct location name and navigation behavior.
  * Russian text now uses the proper preposition for the digging location.
  * Completing these orders now routes users to the correct beach scene before continuing.

* **Style**
  * Minor formatting cleanup was applied to a location constant.

### DIFF
--- a/src/features/island/delivery/components/Orders.tsx
+++ b/src/features/island/delivery/components/Orders.tsx
@@ -80,9 +80,9 @@ export const BEACH_BUMPKINS: NPCName[] = [
   "finn",
   "finley",
   "miranda",
-  "old salty",
-  "pharaoh",
 ];
+
+export const DIGGING_BUMPKINS: NPCName[] = ["old salty", "pharaoh"];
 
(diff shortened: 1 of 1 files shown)


eliasSFL (approved):
LGTM!

PR #7413, opened 2026-07-08
[FEAT] Bulk fetch: spread energy use evenly across pets

eliasSFL:
## What

Bulk fetch now **spreads energy use evenly across pets** instead of concentrating on the highest-yield ones.

Before, the per-fetch pet pick preferred higher yield, so an order drained the two highest-yield pets while lower-level pets sat idle. Now, among equally-specialist pets, each fetch goes to the pet with the **most energy remaining**, so all pets drain down together and the idle ones get used.

## Change

One tie-break flip in `planBulkFetch.ts` — "most energy left" is now ranked above "higher yield". Everything else is unchanged:
- **Scarcity ordering** (rarest resource first) — unchanged.
- **Specialist preservation** (rare resources still go to the pets that can make them) — unchanged.
- Yield is now only a tie-break; a stable id breaks the final tie.

## Trade-off

Spreading onto lower-yield pets means slightly **more total fetches/energy**, scaling with the yield gap between pets — negligible on a balanced roster, noticeable only when mixing e.g. a lvl-1 and a lvl-100 pet.

## Testing

- `planBulkFetch.test.ts` — 8/8, including a new "spreads the load across pets by energy" case (both pets used instead of one); the scarcity/specialist and boosted-yield tests still hold.
- tsc clean, ESLint + Prettier clean.

FE-only — `pets.bulkFetch` just applies whichever list the planner produces, so no BE change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource-fetch planning when multiple pets can contribute, distributing work more fairly based on remaining energy.
  * Refined the selection order for eligible pets to better balance assignments and outcomes in shared scenarios.
* **Tests**
  * Added new coverage for bulk-fetch allocation, including mixed-resource and single-source edge cases, to prevent regressions.

### DIFF
--- a/src/features/island/buildings/components/building/petHouse/planBulkFetch.test.ts
+++ b/src/features/island/buildings/components/building/petHouse/planBulkFetch.test.ts
@@ -186,4 +186,40 @@ describe("planBulkFetch", () => {
     });
     expect(plan.shortfall).toEqual({});
   });
+
+  it("still drains a pet heavily when it is the only source of a resource", () => {
+    // Barkley (Dog) is the only pet that can make chewed bone. Even with
+    // even-spread on, it takes the full chewed-bone load (using far more
+    // energy than Meowchi), and the shared acorn goes to the pet with energy
+    // to spare. That lopsided drain is acceptable when there is no alternative.
+    const activePets: ActivePets = [
+      ["Barkley", makePet("Barkley", { experience: LEVEL_3_XP, energy: 1000 })],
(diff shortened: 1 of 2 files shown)


eliasSFL:
Closing — keeping main's energy-efficient (yield-first) model. Because energy cost per fetch is fixed, concentrating on the highest-yield pets minimises total energy; even-spread trades that away for balanced drain, which only pays off on same-level rosters and wastes energy on mixed ones. Idle pets are the accepted cost of the efficient model.

PR #7417, opened 2026-07-08
[FEAT] Add Bulk Feed food exclusions to Pet House

Dionis404:
## Summary
- Adds a preferences panel (gear icon) in the Pet House manage view where players can mark foods to be excluded from Bulk Feed
- Exclusions are stored per-player in `GameState.pets.bulkFeedExclusions` via a new `pets.bulkFeedExclusionsUpdated` event, so they persist in the save
- Bulk Feed's food-selection algorithm now skips excluded foods when auto-picking what to feed

## Rollout note
Recommend rolling this out to beta testers first before a full release, to validate the UX and gather feedback before it reaches all players.

## Demo

https://github.com/user-attachments/assets/a1f291a4-7b76-4e54-bfed-42b26544093d

## Test plan
- [ ] Build/place a Pet House with pets that have active food requests
- [ ] Open Pet House → gear icon → toggle a food off
- [ ] Confirm Bulk Feed no longer auto-selects that food, while other requests are still selected normally
- [ ] Toggle it back on and confirm it's selected again
- [ ] Reload the game and confirm the exclusion persisted

## Summary by CodeRabbit

- **New Features**
  - Added bulk-feed preferences for searching and filtering available pet foods.
  - Players can enable or exclude specific foods from bulk feeding.
  - Preferences are saved between sessions.
  - Added visual indicators for excluded foods on pet cards and food selections.
  - Added translated labels, descriptions, filters, and empty-state messaging.

- **Bug Fixes**
  - Corrected translation file formatting to ensure language resources load reliably.


eliasSFL (commented):
I would suggest storing this on local storage instead actually, since this is really a FE only thing. it would mean that between devices they would need to set again but it saves data storage in the server

Dionis404:
> I would suggest storing this on local storage instead actually, since this is really a FE only thing. it would mean that between devices they would need to set again but it saves data storage in the server

I was starting from the fact that a player who constantly plays on a PC and has food selection settings on it may one day log in from his phone and think that the setting works and make a bulk feeding.  Which will lead to a negative experience.

eliasSFL (changes requested):
Hey @Dionis404 I love this idea! However the button styles to set the option doesn't really match our design system. I suggest using something similar to our old deposit UI, where clicking on a box would move that food from include to exclude, or viceversa as seen in the wireframe below:

Image

Dionis404:
@eliasSFL  Okay, I'll take it to work this week or next week.

Dionis404:
Closing in favour of #7533, which supersedes this — addresses Elias's design review (click-to-toggle Included/Excluded board instead of the settings-gear list), moves exclusions to localStorage per his suggestion, and folds this into the broader Pets/Feed/Guides tab redesign.

PR #7424, opened 2026-07-09
[FIX] Finn and Timmy delivery icon layering

alexandrdragin:
Problem: Delivery-ready box icons for Finn and Timmy were rendered inside the NPC label/container, so top map layers and nearby buildings could draw over the icon.

telegram-cloud-photo-size-2-5355282863714146749-x
telegram-cloud-photo-size-2-5364059556369079286-x

Solution: Allow labels to render an optional floating icon at an explicit scene depth, and use it for delivery icons so the box stays above building layers without lifting the whole NPC.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved delivery icon rendering on NPC labels so the icon appears on the correct layer and remains properly aligned as the scene updates.
  * Fixed floating icon visibility and positioning behavior, ensuring icons stay in sync when labels or their parent containers are shown, hidden, or moved.

### DIFF
--- a/src/features/world/containers/BumpkinContainer.ts
+++ b/src/features/world/containers/BumpkinContainer.ts
@@ -26,6 +26,7 @@ const NAME_ALIASES: Partial<Record<NPCName, string>> = {
   "pumpkin' pete": "pete",
   "hammerin harry": "auctions",
 };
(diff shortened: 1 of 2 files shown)


adamhannigan (approved):
LGTM ✅ This always bothered me!

Issue #7387, opened 2026-06-29
[BUG] Pet icon showing but no animals are ready to be pet


FunixG:
### Search existing issues

- [x] I have searched the existing issues

### Summary of Bug

The check for the pet available is wrong for barn and hen house. We need to rewrite the code to handle properly the pet time with maybe a variable pet_at and next_pet_at to handle the pet available or not.

### Steps to Reproduce

1. Pet animals
2. Wait
3. Check after if the icon shows but there is no animals to be pet

### Screenshots

<img width="865" height="362" alt="Image" src="https://github.com/user-attachments/assets/b0d8026f-d7e9-40d5-bbf5-d7ae1d890107" />

<img width="844" height="475" alt="Image" src="https://github.com/user-attachments/assets/c9f92b5c-d357-4838-9049-04852cb67308" />

<img width="550" height="410" alt="Image" src="https://github.com/user-attachments/assets/01974188-749e-4e8c-9bb0-ed268632af30" />

### Device Type

None

FunixG:
To add new variable into objects, do we need changes on the BE ? For example if I need to add pet_at and next_pet_at variables Dates objects, do I need to ask BE changes ?

alexandrdragin:
https://github.com/sunflower-land/sunflower-land/pull/7426

PR #7426, opened 2026-07-09
[FIX] Hide stale animal care status icons

alexandrdragin:
[Pull request description
](https://github.com/sunflower-land/sunflower-land/issues/7387)

## Summary
Fixes stale Barn and Hen House exterior care indicators by only showing the care icon while an animal is still asleep and actually eligible for love.

## Context / motivation
After the animal status icons were added, the exterior care icon could appear after animals woke up and started asking for food. The root cause was that isAnimalNeedingLove only checks whether the love cooldown gate has opened; it does not mean the animal is still asleep. Once awakeAt has passed, feeding is the next required action, so the outside building indicator should not advertise care.

## How to test
1. Put Barn or Hen House animals into a state where the love window has opened but awakeAt has passed.
2. Confirm the exterior building shows the feeding indicator, not the care indicator.
3. Put animals into a sleeping state where the love window is open before awakeAt.
4. Confirm the exterior care indicator still appears.
5. Run `yarn test src/features/game/events/landExpansion/loveAnimal.test.ts --runInBand`.

## Summary

* **New Features**
  * Love alert indicators for animals now show only after the love window opens and while the animal is still asleep.
  * Barn and hen house love icons now follow this updated timing.

* **Bug Fixes**
  * Corrected the love status timing so the alert no longer stays visible after an animal wakes up.
  * Added test coverage to ensure the alert appears at the correct moment and disappears immediately after waking.

### DIFF
--- a/src/features/game/events/landExpansion/loveAnimal.test.ts
+++ b/src/features/game/events/landExpansion/loveAnimal.test.ts
@@ -2,6 +2,7 @@ import { INITIAL_FARM } from "features/game/lib/constants";
 import {
   getNextLoveAvailableAt,
   isAnimalNeedingLove,
+  isAnimalReadyForLove,
   loveAnimal,
 } from "./loveAnimal";
 import Decimal from "decimal.js-light";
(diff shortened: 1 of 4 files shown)


eliasSFL (approved):
LGTM!

PR #7430, opened 2026-07-13
[FEAT]  Claim airdrops and marketplace sales from mailbox Rewards tab

Dionis404:
## Summary
- Adds a "Rewards" tab to the in-game mailbox that lists pending airdrops (Telegram quest rewards, dev giveaways) and fulfilled marketplace sales together, with a single "Claim All" button, plus a manual "Check for Rewards" refresh — so players no longer need to hard-refresh the page or claim rewards one-by-one.
- The same combined list now auto-shows right after the daily reward is acknowledged (reusing the existing notification queue order in the game state machine), replacing the old separate one-by-one AirdropPopup / MarketplaceSalesPopup.
- Fixes a game state machine bug where `purchase.claimed` was not handled while in the `airdrop` state (and `airdrop.claimed` was not handled while in `marketplaceSale`), which silently dropped the claim for whichever reward type didn't match the current state.

https://github.com/user-attachments/assets/96d13944-eb85-4bb9-94ca-ca0a6e61e788

## ⚠️ Status
This still needs further testing and review before merge — I've only verified it locally in ART_MODE with manually seeded test data (airdrops, marketplace listings). It has **not** been tested against a real backend/session yet. Please hold off merging until that verification is done.

## Test plan
- [ ] Verify against a real backend session (not just ART_MODE)
- [ ] Confirm airdrops with negative item amounts (e.g. admin corrections) still claim correctly through the new list
- [ ] Confirm on-chain (signed) marketplace listings correctly trigger the RESET flow after claiming
- [ ] Confirm the Rewards tab and auto-popup behave correctly when there are zero pending rewards
- [ ] Confirm multiple queued airdrops all claim correctly via Claim All

## Summary by CodeRabbit

* **New Features**
  * Added a **Rewards** tab to the mailbox, with a “check rewards” flow to refresh pending airdrops and marketplace sale proceeds.
  * Introduced a **Rewards** popup that replaces separate airdrop and marketplace sale modals, including reward breakdown, loading/empty states, and a **Claim all** action.
  * Added new localized strings for the rewards tab, checking state, empty state, and claim actions.
* **Bug Fixes**
  * Prevented the shared rewards modal from getting stuck open after rewards are claimed (auto-closes when nothing is left to claim).


adamhannigan:
@Dionis404 A lot of important problems you are solving here.

I am going to close this PR and ask that you address some of these separately. We won't introduce a new rewards tab into the news centre - typically empty reward sections give new players lots of questions (how do I get rewards 😅) and adds more noise to this section of the game.

1. `Fixes a game state machine bug where purchase.claimed...` - Nice find ✅  Please open this fix in a different PR. It's best practice to split out small changes so we can review independently and ensure we don't miss something
2. My understanding is that this PR addresses the problem of having to accept lots of different trades/airdrops etc. I would suggest keeping these as popups during the game load, but building a UI which lets you claim all of them at once

Building the new UI is quite a huge task - it will be about simplifying the amount of rewards/trades that have happened while you were away. The current large buttons + panels would not work well here. 

We often get reports of players not receiving FLOWER from a listing - when in fact they just skipped through the claim too fast. I can imagine them scrolling quickly to get through a huge/long lists would make this worse as they miss certain events.

I would suggest trying to condense the airdrops/trades (one per row) and building out more of a table like structure. In these scenarios, players are quickly scanning for important details so structure + simplicity helps out.

This will be a tricky change IMO as it will depend a lot on the UI/UX design

PR #7432, opened 2026-07-13
[FEAT] Buy All Seeds button in the seed market

Dionis404:
## Summary
- Adds a "Buy All Seeds" button to the seed market (`SeasonalSeeds.tsx`) that computes the maximum affordable amount of every unlocked seed for the current season (plus Crop Machine seeds and Full Moon seeds when applicable), spending from a single shared coin pool.
- Skips seeds that are locked by level, out of stock, would exceed the inventory limit, or whose required planting spot (Crop Plot / Fruit Patch / Flower Bed / Greenhouse) isn't owned — mirrors the exact validation `seedBought.ts` already enforces, so nothing skipped by the plan could have failed the purchase anyway.
- Shows a confirmation modal with the full per-seed breakdown and total coin cost before spending anything; cancel is a no-op.
- New pure helper `planSeedPurchases.ts` (no UI dependencies) computes the plan, so it's easy to unit test independently of the component.
image
image

## Status
Manually tested locally (ART_MODE, via a temporary dynamic test farm) across: basic/spring/desert islands, low bumpkin level (locked seeds correctly skipped), missing Fruit Patch/Flower Bed (correctly skipped), insufficient coins (partial plan, rest skipped), Full Moon event active (full moon seeds included when stocked), and Kuebiko built (free seeds, plan capped only by stock/inventory limit). 

## Test plan
- [ ] Verify against a real backend session (not just ART_MODE)
- [ ] Confirm the confirmation modal's total matches the actual coins deducted after purchase
- [ ] Confirm locked/out-of-stock/inventory-full seeds are silently skipped without breaking the rest of the batch
- [ ] Confirm Crop Machine seed inclusion respects owned Crop Extension Module skills

## Summary by CodeRabbit

* **New Features**
  * Added a bulk **“Buy All Seeds”** option in the seasonal seed market that plans purchases based on available eligibility and affordability.
  * Includes a confirmation dialog with total coin cost, per-item breakdown, and partial-failure messaging, with VIP gating and expansion requirements.
* **Bug Fixes**
  * Improved planting-spot validation so missing inventory entries are treated as zero, preventing incorrect skips.
* **Tests**
  * Added/expanded coverage for bulk purchase planning, eligibility/skip rules, and full-moon + chapter-crop-week timing.
  * Updated test fixtures to ensure required planting-spot inventory keys are present.


adamhannigan:
Nice initiative @Dionis404 👏 

My one concern here is that the "Buy All Seeds" is the biggest call to action on the page. For new players this makes it a bit more confusing and draws their attention away from purchasing seeds.

1. Do we need to show this for all players? (is this something that players find useful later in the game)
2. Would this still work if the "Buy all seeds" buttons was placed below all of the seeds.

I would also put on my greedy business hat and suggest that this is a VIP only feature - the quality of life improvements usually fall into things we'd like to put behind and reward the VIP players with.

Dionis404:
@adamhannigan  I can add settings for the display of this button and the selection of seeds, similar to what I suggest here for the tools: https://github.com/sunflower-land/sunflower-land/pull/7435

Dionis404:
Regarding this option for VIP only. 
My opinion is that such functions should be for VIP only 
The more VIP features we have, the more attractive it becomes to purchase it.  

But the final decision is always yours.

adamhannigan:
@Dionis404 I don't think a custom settings is needed for the seeds.

The following changes would suffice from a UX perspective:

1. Move the button to the bottom of the seeds
2. Only show for players on spring island and further (so tutorial players don't see it)
3. Make the feature VIP only (show the VIP crown on the button) and then some blocking logic in the confirmation screen if they do not have VIP 👑

Dionis404:
@adamhannigan 

Without VIP


https://github.com/user-attachments/assets/1695e323-a9e4-4b70-9d8b-59968f2501da


With VIP


https://github.com/user-attachments/assets/927afc6d-3ac2-4e40-816d-a4eede7354ca

adamhannigan:
@Dionis404 Nice - this is super close.

I would just suggest that even without VIP they can open the modal. This helps them understand what the button actually does.

You can then disable the "Buy all seeds" confirmaton button and put a red "VIP Required" label on it to make it a bit clearer.

After that we are good to go ✅ Cheers!

Dionis404:
@adamhannigan  
image
image

adamhannigan (approved):
Great work 👏 @Dionis404

PR #7433, opened 2026-07-13
[FEAT] Preserve farm layout across island ascension

eliasSFL:
## What

Reworks island **ascension** (behind `SWAMP_ASCENSION`) so it no longer wipes the farm.

- **Volcano → Swamp (first ascension):** keeps the player's arrangement exactly in place — skips `removeAll` + `placeInitialLand` — since the land is identical (both 30 expansions) and a maxed volcano already meets the swamp node floor. The arrangement is saved as a **protected "Ascension Layout"** (`SavedLayout.auto`): it can't be deleted, renamed or overwritten, and doesn't count against the manual `MAX_SAVED_LAYOUTS` cap.
- **Swamp → future ascensions:** re-applies that saved Ascension Layout via `applyFarmLayout` (reset to the canonical arrangement). Legacy farms with no saved layout fall back to the old `placeInitialLand` behaviour.
- **Ascension rewards → side-island chest:** the Ascension Crystals a player is owed (most players won't have any yet) plus any node shortfall are delivered via a reward chest (airdrop) on the side island instead of straight to inventory. The chest reuses `revealLand`'s `missing-resources` back-pay accounting + dedup (so crystals are never double-granted), is keyed on `ascensionLevel` for a unique id (incl. the infinite `marble→marble` loop), and picks a free side-island tile so un-collected chests don't stack.

## Why

Players lost their carefully-built farm arrangement on every ascension. Now the arrangement is preserved and the ascension rewards are surfaced as a collectable chest.

## Notes

- Gated behind `SWAMP_ASCENSION` (testnet + team usernames); the saved-layout UI is behind `SAVED_LAYOUTS`.
- `ascensionUpgrade` no longer tops resource inventory up to the floor — any shortfall flows to the reward chest instead.
- Mirrors the backend change 1:1: sunflower-land/sunflower-land-api#3486.

## Test plan

- `yarn jest src/features/game/events/landExpansion/upgradeFarm.test.ts src/features/game/events/landExpansion/ascensionCrystalGrants.test.ts src/features/game/events/landExpansion/saveLayout.test.ts src/features/game/events/landExpansion/renameLayout.test.ts src/features/game/events/landExpansion/deleteLayout.test.ts src/features/game/events/landExpansion/applyLayout.test.ts` — all green.
- New coverage: arrangement preserved across ascension, growth timers untouched (no instant growth), reward chest holds owed crystals + node shortfall (not inventory), Ascension Layout saved once and protected (cap-exempt / un-deletable), future-ascension layout reuse, unique chest ids + no stacking.
- `tsc --noEmit` clean.

## Summary by CodeRabbit

- **New Features**
  - Preserved island farm arrangements across ascensions using an auto-managed “Ascension Layout.”
  - Ascension rewards now deliver via side-island reward chests, including missing-resources handling.
- **Bug Fixes**
  - Protected “Ascension Layout” can’t be deleted, renamed, or overwritten; auto layouts don’t count toward the saved-layout limit.
  - Improved layout restoration by unplacing conflicting instances and restoring intended placements precisely.
  - Updated reward/back-pay reconciliation to use shared missing-resource calculations and place safe, non-colliding chests.
- **Tests**
  - Expanded unit coverage for protected layouts, ascension transitions, reward chest delivery, and restoration edge cases.

### DIFF
--- a/src/features/game/events/landExpansion/lib/layouts.ts
+++ b/src/features/game/events/landExpansion/lib/layouts.ts
@@ -857,6 +857,33 @@ export function applyFarmLayout(
     });
   }
 
+  // Exact restore: applying a layout yields exactly the saved snapshot, so lift
+  // every placed item the layout does NOT represent too — its instance (with any
+  // timers) is kept; it simply returns to inventory. Resources are lifted in full
+  // above; mirror that for the name/id-keyed categories. The player's own Bumpkin
+  // is intentionally left alone — it is the avatar and is always captured on-farm.
+  getObjectEntries(state.collectibles).forEach(([name, items]) => {
+    if (layout.collectibles[name]) return;
+    items?.forEach((item) => {
+      item.coordinates = undefined;
+    });
+  });
+  getObjectEntries(state.buildings).forEach(([name, items]) => {
+    if (layout.buildings[name]) return;
+    items?.forEach((item) => {
+      item.coordinates = undefined;
+    });
+  });
+  Object.entries(state.buds ?? {}).forEach(([id, bud]) => {
+    if (!layout.buds?.[id]) bud.coordinates = undefined;
+  });
+  Object.entries(state.pets?.nfts ?? {}).forEach(([id, pet]) => {
+    if (!layout.petNFTs?.[id]) pet.coordinates = undefined;
+  });
+  Object.entries(state.farmHands?.bumpkins ?? {}).forEach(([id, fh]) => {
+    if (!layout.farmHands?.[id]) fh.coordinates = undefined;
+  });
+
   // Everything is now lifted. Place each slot in turn against the live draft
   // (non-layout items + already-placed slots). A blocked slot stays unplaced
   // for now; reused instances get a best-effort restore pass below.
--- a/src/features/game/events/landExpansion/upgradeFarm.ts
+++ b/src/features/game/events/landExpansion/upgradeFarm.ts
@@ -1086,6 +1086,15 @@ function buildAscensionRewardBundle(
 ): Partial<Record<InventoryItemName, number>> {
   const bundle: Partial<Record<InventoryItemName, number>> = {};
 
+  // Items already promised by an un-collected reward chest count as owned — the
+  // player will receive them on claim. Ignoring them would re-issue the same
+  // crystals/nodes on the next ascension (both chests stay claimable).
+  const pending = (item: InventoryItemName): number =>
+    (game.airdrops ?? []).reduce(
+      (total, airdrop) => total + (airdrop.items[item] ?? 0),
+      0,
+    );
+
   // Node floor for the starting island (base row), excluding the sunstone bonus —
   // matching `ascensionUpgrade`'s minimum. After `applySetup` this is already met,
   // so the shortfall is only positive for legacy / under-provisioned accounts.
@@ -1098,16 +1107,18 @@ function buildAscensionRewardBundle(
     "Sunstone Rock": 0,
   };
   getObjectEntries(floor).forEach(([resource, amount]) => {
-    const shortfall =
-      (amount ?? 0) - getTotalBaseResourceEquivalents(game, resource);
+    const owned =
+      getTotalBaseResourceEquivalents(game, resource) + pending(resource);
+    const shortfall = (amount ?? 0) - owned;
     if (shortfall > 0) {
       bundle[resource] = (bundle[resource] ?? 0) + shortfall;
     }
   });
 
   // Ascension Crystals owed for this ascension: the expected cumulative count
-  // minus what the player already owns (inventory + placed) and has mined
-  // (single-use). This replaces the +1 inventory grant on the ascension path.
+  // minus what the player already owns (inventory + placed), has mined
+  // (single-use), or has pending in an un-collected chest. This replaces the +1
+  // inventory grant on the ascension path.
   const expectedCrystals = getExpectedAscensionCrystals({
     islandType: target,
     ascensionLevel: game.island.ascensionLevel ?? 0,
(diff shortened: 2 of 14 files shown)


src/features/game/events/landExpansion/upgradeFarm.ts
```diff
@@ -1086,11 +1069,113 @@ const ISLAND_SETUP: Record<UpgradeTarget, IslandSetup> = {
...
+  // (single-use), or has pending in an un-collected chest. This replaces the +1
+  // inventory grant on the ascension path.
+  const expectedCrystals = getExpectedAscensionCrystals({
+    islandType: target,
+    ascensionLevel: game.island.ascensionLevel ?? 0,
+    basicLand: game.inventory["Basic Land"]?.toNumber() ?? 0,
+  });
+  const ownedCrystals =
+    (game.inventory["Ascension Crystal"]?.toNumber() ?? 0) +
+    Object.keys(game.ascensionCrystals ?? {}).length;
```
eliasSFL: **[Correctness — crystal accounting] `ownedCrystals` double-counts placed Ascension Crystals.**

`placeAscensionCrystal` computes `available = inventory["Ascension Crystal"] − placedCount`, so (like every resource) `inventory["Ascension Crystal"]` is the **total owned, already including placed nodes**. Adding `Object.keys(game.ascensionCrystals).length` here counts placed crystals a second time.

Effect: a player who has any Ascension Crystal nodes placed (or sitting removed-but-present in the bucket) when they ascend gets `ownedCrystals` inflated → `crystalShortfall` too low → the reward chest **under-delivers** crystals. It self-heals later via `revealLand` back-pay (which correctly uses `expected − inventory − mined`, no placed term), so it is the *safe* direction — not the over-grant you were worried about — but the two paths use inconsistent models and the chest is wrong.

Fix: drop the `+ Object.keys(...)` term so it mirrors `revealLand`:
```suggestion
  const ownedCrystals =
    game.inventory["Ascension Crystal"]?.toNumber() ?? 0;
```
Also mirrored in the BE `src/domain/game/events/upgradeFarm.ts`. Worth a regression test that seeds a *placed* crystal (the current tests only seed inventory).

eliasSFL: Fixed in 3616c5934 (BE mirror sunflower-land/sunflower-land-api@7221e49d5). `ownedCrystals` now uses inventory only — `expected − inventory − mined − pending`, identical to `revealLand`. Added a placed-crystal regression test (existing tests only seeded inventory) that was proven red→green: it fails on the old line (`Expected: 1, Received: undefined`) and passes after.

src/features/game/events/landExpansion/upgradeFarm.ts
```diff
@@ -1202,26 +1229,80 @@ function transitionToIsland({
...
+      farmId,
+      createdAt,
+      initialLandCoordinates: setup.initialCoordinates,
+    });
+  }
+
+  // Capture the arrangement as the protected, auto-managed "Ascension Layout" the
+  // first time the player ascends (volcano→swamp); it is reused on every later
+  // ascension. Appended once (idempotent guard) and exempt from the manual cap.
+  if (isFirstAscension && !game.layouts?.some((layout) => layout.auto)) {
```
eliasSFL, line 1256: **[Layout-preservation gap for existing swamp players]** `auto: true` is only ever set here, and only when `isFirstAscension` (source island `=== "volcano"`). Players who already ascended volcano→swamp *before this ships* have no auto-layout, so every future ascension (swamp→spooky→…) falls through to the wipe path (`removeAll` + `placeInitialLand`) **and** never captures one either (the guard requires the source to be volcano). They're permanently excluded from the feature and their arrangement is wiped on each ascension, with no migration to backfill it.

No resource leak — items lift back to inventory and the chest back-pay is conservation-safe — but worth confirming this is intended vs. capturing on the first ascension *from any island* when no auto-layout exists yet.

src/features/game/events/landExpansion/upgradeFarm.ts
```diff
@@ -1086,11 +1069,41 @@ const ISLAND_SETUP: Record<UpgradeTarget, IslandSetup> = {
...
 
+/**
+ * A free side-island tile for the ascension reward chest. The side island sits
+ * off the main land, so `pickEmptyPosition`/`detectCollision` (which treats
+ * off-land tiles as water) can't be used — instead scan tiles just below the
+ * mushroom spawn rows and return the first not already holding an airdrop. The
+ * scan walks down unbounded rows so a genuinely free tile is always found, no
+ * matter how many un-collected chests have piled up (e.g. marble→marble).
+ */
+function pickAscensionChestPosition(
```
eliasSFL, line 1080: **[Reward-chest placement]** `pickAscensionChestPosition` only treats tiles held by *other airdrops* as taken. Mushrooms and relocated social-farming clutter on the side island can sit at `y ≥ 7`, so the scan can place the chest directly on top of them. It also anchors on `getIslandAnchorX(setup.startingExpansions)` (a fixed 30 for ascension); as the player later expands the island the side island re-anchors leftward while this already-dropped airdrop stays put, so an un-collected chest can end up in open water / over the main land and be awkward to reach.

Not a resource-conservation issue (the items are correct), purely claimability/visual — flagging in case chest reachability matters.

src/features/game/events/landExpansion/lib/layouts.ts
```diff
@@ -857,6 +857,33 @@ export function applyFarmLayout(
     });
   }
 
+  // Exact restore: applying a layout yields exactly the saved snapshot, so lift
+  // every placed item the layout does NOT represent too — its instance (with any
+  // timers) is kept; it simply returns to inventory. Resources are lifted in full
+  // above; mirror that for the name/id-keyed categories. The player's own Bumpkin
+  // is intentionally left alone — it is the avatar and is always captured on-farm.
+  getObjectEntries(state.collectibles).forEach(([name, items]) => {
```
eliasSFL, line 865: **[Shared `applyFarmLayout` semantics change]** This exact-restore block now unplaces *every* collectible / building / bud / pet / farmHand not represented in the layout. That's required for the ascension reuse path, but `applyFarmLayout` is also the engine behind the user-facing `applyLayout` (saved layouts): applying a manual layout now lifts all *unrelated* placements back to inventory instead of leaving them in place. The tests were updated to match, so this reads as intentional — flagging so the saved-layouts UX change is a deliberate call rather than an accidental side effect of the ascension work.

PR #7443, opened 2026-07-15
[FEAT] Compost ascension skill upgrades

eliasSFL:
Adds rank 1-3 ascension upgrades for **10 of the 13 Compost skills**, mirroring the 8 merged trees. Rank 1 reproduces the current live magnitude for every skill, so nothing changes for players who don't rank up.

Values transcribed from the **v5 sheet**. No sheet/code disagreements found.

| Skill | Effect kind | Ranks 1/2/3 |
|---|---|---|
| Efficient Bin | `additiveYield` | +5 / +7 / +9 Sprout Mix |
| Turbo Charged | `additiveYield` | +5 / +7 / +9 Fruitful Blend |
| Wormy Treat | `additiveYield` | +1 / +2 / +3 Worms |
| Feathery Business | `costMultiplier` | 2x / 1.5x / 1x Feather cost |
| Premium Worms | `additiveYield` | +10 / +15 / +20 Rapid Root |
| Fruitful Bounty | `multiplier` | 2x / 3x / 4x Fruitful Blend effect |
| Swift Decomposer | `growthMultiplier` | 0.9 / 0.875 / 0.85 compost time |
| Composting Bonanza | `flatTimeBonus` **(new)** | +1hr / +1.5hr / +2hr |
| Composting Overhaul | `additiveYield` | +2 / +5 / +8 Worms |
| Composting Revamp | `yieldWithDebuff` | +5/+8/+10 fertilisers, −2/−3/−4 Worms |

**Skipped:** Sprout Surge, Blend-tastic, Root Rocket — all Power Skills, all `Rank=No` in the sheet.

### Notes for reviewers

- **New effect kind:** only `flatTimeBonus` (ms shaved off an in-flight production) for Composting Bonanza. Everything else reuses existing kinds. The `costMultiplier` comment was broadened from "coin cost" to "coin or resource cost", since Feathery Business multiplies a Feather cost.
- **No in-flight composter snapshotting was needed** — the existing shape already handles it, and this PR matches it rather than adding a mechanism:
  - Produce amounts (Efficient Bin / Turbo Charged / Premium Worms / Composting Revamp buff) are already written into `producing.items` at start, so they are snapshotted by value. A test asserts a mid-production rank-up does not retroactively change output.
  - Swift Decomposer folds into `readyAt` at start — snapshotted the same way.
  - Worm boosts (Wormy Treat / Composting Overhaul / Composting Revamp debuff) roll live at collect **by existing design** — there is an explicit comment in `startComposter.ts` stating boosts equipped any time before collect apply. That behaviour is preserved; ranks read live.
  - Feathery Business / Composting Bonanza / Fruitful Bounty resolve instantaneously at the accelerate/fertilise event.
- **One UI label had to become rank-aware.** `collectibleItemBuffs.ts` hard-coded `"+0.1 Fruit Patch Yield from Fruitful Bounty Skill"`, which is only correct at rank 1 (+0.2/+0.3 at ranks 2/3). It is now value-driven via a new `.ranked` key. To avoid the import cycle codex flagged on the Machinery PR, the shared `FRUITFUL_BLEND_YIELD` constant moved to a new zero-import leaf, `types/fertilisers.ts`. The old non-ranked key is now orphaned in `dictionary.json` but left in place, since the locale files are bot-owned.
- **i18n:** new keys in `dictionary.json` only; `en.json` untouched.

`SKILL_RANKS` goes 94 → **104**, matching the BE PR exactly.

Backend: sunflower-land-api#3493

## Summary by CodeRabbit

* **New Features**
  * Skill bonuses and penalties now scale by rank across composting, fertilising, and worm bait systems, affecting output amount, speed/cost, and completion-time modifiers (including in-progress timing).
  * Added rank-based, parameterized localization and updated skill rank descriptions for composting skills and Fruitful Blend/Fruitful Bounty.

* **Bug Fixes**
  * Fixed compost and bait calculations to use per-rank scaling instead of fixed/boolean effects.
  * Ensured ongoing compost production isn’t retroactively changed when skills are upgraded mid-run, and worm ranges remain safely clamped to non-negative values.


src/features/game/events/landExpansion/fertiliseFruitPatch.ts
```diff
@@ -70,12 +72,26 @@ type Options = {
...
+  const fruitfulBountyLevel = getSkillLevel(
+    state.bumpkin?.skills ?? {},
+    "Fruitful Bounty",
+  );
+  if (fruitfulBountyLevel) {
+    const v = SKILL_RANKS["Fruitful Bounty"].ranks[fruitfulBountyLevel - 1];
+    // The skill multiplies Fruitful Blend's base effect, so the marginal gain
+    // is (multiplier - 1) x base.
+    const added = fruitfulBlendBuff * (v - 1);
+    fruitfulBlendBuff *= v;
```
eliasSFL: **Float drift: rank 2 returns `0.30000000000000004`, not `0.3`.**

`fruitfulBlendBuff *= v` with `multiplier` ranks `[2,3,4]` on a `0.1` base is exact at ranks 1 and 3, but `0.1 * 3 = 0.30000000000000004`. The `added` value was rounded for the boost label, but `fruitfulBlendBuff` itself — the `amount` actually returned — was not. It flows into `getFruitYield`, which accumulates into a plain `number`, so the drift reaches the player's yield.

Fruitful Bounty is the only remaining `multiplier` over a **fractional** base — `Pear Turbocharge` and `Leak-Proof Tank` multiply integers, where drift is impossible.

**Resolved:** computed in `Decimal` (already imported in this file), which keeps the v5 sheet's multiplier semantics, both boost entries, and every i18n string unchanged — and lets both `Math.round` guards go:

```ts
const base = new Decimal(FRUITFUL_BLEND_YIELD);
const added = base.mul(v - 1);      // 0.1 / 0.2 / 0.3 exactly
fruitfulBlendBuff = base.mul(v);    // 0.2 / 0.3 / 0.4 exactly
// ...
return { amount: fruitfulBlendBuff.toNumber(), boostsUsed };
```

> **Correction to an earlier revision of this comment:** it suggested converting to `additiveYield [0.2, 0.3, 0.4]` by analogy with Loyal Macaw. Two things were wrong. Loyal Macaw's actual ranks are **`[0.2, 0.25, 0.3]`** (a v5 rebalance moved them and left its code comment saying "doubled/tripled/quadrupled"). And its shape only works because it *subsumes* its base in an if/else — storing the marginal here would not help, since `0.1 + 0.2` drifts to `0.30000000000000004` just the same. Decimal was the chosen fix.

src/features/game/events/landExpansion/fertiliseFruitPatch.test.ts
```diff
@@ -236,3 +237,22 @@ describe("fertiliseFruitPatch", () => {
     });
   });
 });
+
+describe("Fruitful Bounty skill ranks", () => {
+  it.each([
+    [1, 0.2],
+    [2, 0.30000000000000004],
```
eliasSFL: **This test hardcodes the float drift as the expected value.**

`0.30000000000000004` is not the intended boost — rank 2 is meant to be `+0.3`. The test enshrines the bug rather than catching it, and the BE test for identical logic asserts `0.3`:

```ts
// sunflower-land-api/src/domain/game/events/fertiliseFruitPatch.test.ts
it.each([
  [1, 0.2],
  [2, 0.3],
  [3, 0.4],
])
```

Both pass only because `toBeCloseTo(expected, 10)` tolerates a 5.5e-17 difference — so this row could have read `0.3` and stayed green either way. As written it reads as documentation that `0.30000000000000004` is correct, and the two repos now disagree on what rank 2 is worth.

Once the source fix lands (see my comment on `fertiliseFruitPatch.ts`), this becomes `[2, 0.3]` and matches the BE.

```suggestion
    [2, 0.3],
```

src/lib/i18n/dictionaries/dictionary.json
```diff
@@ -9246,6 +9257,7 @@
   "description.sprout.mix.boost": "+0.2 Crop Plot Yield",
   "description.fruitful.blend.boost": "+0.1 Fruit Patch Yield",
   "description.fruitful.bounty.skill.boost": "+0.1 Fruit Patch Yield from Fruitful Bounty Skill",
```
eliasSFL: **Orphaned key.** `description.fruitful.bounty.skill.boost` no longer has any caller — its only reference (`collectibleItemBuffs.ts`) was switched to `description.fruitful.bounty.skill.boost.ranked` in this PR. It also states a now-wrong value ("+0.1"), which only holds at rank 1.

Nothing catches this: tsc and jest do not verify i18n key usage, so it will sit here and get propagated to every locale by the translation bot. Worth deleting in this PR.

PR #7445, opened 2026-07-15
[FIX] Pet House quick-placement panel offers items it can't place there

Dionis404:
## Summary
- Bug: opening the landscaping quick-drag panel while inside the Pet House showed every collectible category (decorations, boosts, buildings, banners, etc.), but the Pet House only accepts pets - attempting to place anything else there doesn't work.
- Root cause: `Chest.tsx` already restricts its item list to pet collectibles (and hides buds/farm hands) when `location === "petHouse"`, but that same gating was never ported to `LandscapingQuickPanel.tsx` when it was added - the two components independently build their item lists from the same `getChestCategories` helper, and only one of them filtered its input.
- Fix: `LandscapingQuickPanel.tsx` now applies the identical `location === "petHouse"` gating as `Chest.tsx` before computing categories, so only the "Pets" tab (plus Pet NFTs) shows up, and empty tabs are hidden automatically by the panel's existing `hasItems` filter - no new UI/tab logic needed.

image

## Test plan
- [x] Full suite: 304/304 test suites, 4464 tests passing
- [x] `tsc --noEmit` and `eslint` clean (only pre-existing, unrelated `any` warnings)
- [x] Verified with a targeted data-level check against the real `getChestCategories` categorizer: on the farm, a mixed inventory (a pet, a boost, a building, a decoration) produces 6 non-empty tabs; applying the same `petHouse` filter this PR adds narrows it to exactly `["pets"]`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the pet house panel to display only relevant pet items.
  * Prevented non-pet collectibles from appearing when viewing the pet house.
  * Updated item filtering when switching between panel locations.

### DIFF
--- a/src/features/island/hud/components/LandscapingQuickPanel.tsx
+++ b/src/features/island/hud/components/LandscapingQuickPanel.tsx
@@ -35,7 +35,7 @@ import {
   BUILDINGS_DIMENSIONS,
   type BuildingName,
 } from "features/game/types/buildings";
-import { isPetNFTRevealed } from "features/game/types/pets";
+import { isPetNFTRevealed, PET_TYPES } from "features/game/types/pets";
 import { Box, type BoxProps } from "components/ui/Box";
 import { ITEM_DETAILS } from "features/game/types/images";
 import { SUNNYSIDE } from "assets/sunnyside";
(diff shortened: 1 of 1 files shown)


c-py (approved):
LGTM! Thanks @Dionis404 !

PR #7446, opened 2026-07-15
[FEAT] Add "New Day" separator to the social Feed

Dionis404:
## Summary
- Inserts a "New Day" divider between Feed entries whenever the UTC calendar day changes, so cheer/help notifications near the daily reset are no longer ambiguous between "today" and "yesterday".
- Reuses the existing `getUTCDateKey` helper already used for the cheer visibility filter — no backend/data changes needed, since every feed item already carries a millisecond `createdAt`.
- Addresses #7041.

image

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` passes (no new warnings)
- [x] Manually verified in a running dev build with mock multi-day feed data: separator renders full-width matching notification cards, calendar icon renders crisp (no sub-pixel text ghosting), correct placement between day boundaries

## Summary by CodeRabbit

* **New Features**
  * Added a translated “Older posts” separator to the feed, automatically shown when older interactions start.

* **Bug Fixes**
  * Improved how feed items are grouped across date boundaries to ensure older sections appear at the correct times.

* **Chores**
  * Added the “Older posts” localization string and updated dictionary formatting.


eliasSFL:
@Dionis404 Thanks for the suggestion! However I wouldn't use a button there as it would make it seem like it's clickable. Instead a simple paragraph field is probably good enough to convey. Also probably good enough to have events from a day ago say something like "older posts" or something for anything over a day ago. don't need to have one for every calendar day

Dionis404:
> @Dionis404 Thanks for the suggestion! However I wouldn't use a button there as it would make it seem like it's clickable. Instead a simple paragraph field is probably good enough to convey. Also probably good enough to have events from a day ago say something like "older posts" or something for anything over a day ago. don't need to have one for every calendar day

@eliasSFL I've updated the display for the new day, is this a better option?

eliasSFL (approved):
LGTM!

PR #7449, opened 2026-07-15
[FIX] Kitchen recipe panel height mismatch

Dionis404:
## Summary
- Cooking building modals (Fire Pit, Kitchen, Bakery, Deli, Smoothie Shack) show a recipe list panel (left) and a recipe details panel (right) side by side. The two panels sized independently, so whichever side had less content (e.g. left panel with few unlocked recipes vs. a right panel stretched by VIP/boost text) left visible empty space instead of the two columns lining up.
- Added an opt-in `matchPanelHeight` prop to `SplitScreenView` that measures the panel (right) column's live rendered height via `ResizeObserver` and caps the content (left) column to match it on desktop, so both columns' borders always align regardless of how much content renders on either side.
- Wired it up only in `Recipes.tsx` (the shared component behind all cooking building modals) — the other 30+ consumers of `SplitScreenView` are unaffected since the prop defaults to off.
image

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` clean on changed files
- [x] Manually verified in-browser (ART_MODE) across Fire Pit, Kitchen, Bakery, Deli, Smoothie Shack at various content lengths (many recipes / few recipes, with and without boosts/VIP) — panels consistently line up on desktop
- [x] Verified no regression on mobile (panels stack, unaffected by the desktop-only height matching)
- [x] Verified no regression in the Crafting Box "Craft" tab (also uses `SplitScreenView`, without the new prop)

## Summary by CodeRabbit

- **New Features**
  - Added an option to match split-screen header and content panel heights on desktop.
  - The height matching updates automatically when the header size changes.
  - Enabled this improved alignment in the Recipes view for a more consistent layout.

### DIFF
--- a/src/components/ui/SplitScreenView.tsx
+++ b/src/components/ui/SplitScreenView.tsx
@@ -1,4 +1,4 @@
-import React, { type JSX } from "react";
+import React, { type JSX, useEffect, useRef, useState } from "react";
 import { InnerPanel } from "components/ui/Panel";
 import classNames from "classnames";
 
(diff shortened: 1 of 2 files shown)


adamhannigan (approved):
LGTM 👍 Nice polish :shipit:

PR #7454, opened 2026-07-18
[FIX] Show cheer feed entries for all days

Dionis404:
## What

Fixes a regression from #7369: cheer rows in the social feed were being hidden whenever they were older than the current UTC day, while help/chat/follow rows stayed visible regardless of age. Removes that date filter so cheer interactions render like every other interaction type - visible for as long as they're in the feed, not just for today.

Also adds a small follow-up: the Following list now shows a cheer icon (mirroring the existing help icon) when you've cheered that farm today, computed client-side from `socialFarming.cheersGiven.farms` - no backend change required.

## Why

#7369 added `shouldShowInteraction()` to hide cheer entries outside the current UTC day, intended to keep cheer activity "fresh," but it also meant a legitimate past cheer would vanish from the feed the next day - inconsistent with how help/chat/follow behave, and confusing since the same information stays visible elsewhere (e.g. `PlayerDetails`).

## Changes

- `Feed.tsx` - removed `shouldShowInteraction`, `getUTCDateKey`, `getNextUTCDateKeyRefreshAt`, the `todayKey` state, and the midnight-UTC refresh timer. The feed list renders directly with no date filter.
- `FollowList.tsx` - reads `socialFarming.cheersGiven.farms` for today via `gameService`/`useSelector` and passes `cheeredThemToday` per row.
- `FollowDetailPanel.tsx` - renders the cheer icon next to the existing help icons when `cheeredThemToday` is true.

## Left untouched

- The per-notification rocket icon on individual cheer rows (`interaction.cheeredThemToday`, from #7379) - still driven by the backend flag, unaffected by this change.
- Help icon logic in the feed - was never date-filtered, unchanged.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` clean (pre-existing `no-explicit-any` warning unrelated to this change)
- [x] Manually verified locally with mock feed/follow data: cheer rows from previous days now render; Following list shows the cheer icon only for farms cheered today; help icons and rocket reciprocity icon behave as before

## Summary by CodeRabbit

* **Bug Fixes**
  * Social feed interactions are now shown consistently without being limited by the current UTC day, improving loading/pagination and the empty-state experience.
  * Follow details now correctly indicate whether you have cheered the viewed player today, including correct “previous interaction” separation for older posts.
* **New Features**
  * Added a “cheered” indicator to the follow detail panel when applicable.

### DIFF
--- a/src/features/social/Feed.tsx
+++ b/src/features/social/Feed.tsx
@@ -113,6 +113,9 @@ const mergeUpdates = (
   ];
 };
 
+const getUTCDateKey = (timestamp: number) =>
+  new Date(timestamp).toISOString().split("T")[0];
+
 export type FeedFilter = "all" | "help" | "chat" | "cheer" | "follow";
 export type FeedFilterOption = {
   value: FeedFilter;
@@ -580,6 +583,8 @@ const FeedContent: React.FC<FeedContentProps> = ({
   const { t } = useAppTranslation();
   const [canPaginate, setCanPaginate] = useState(false);
   const [hasMeasuredScroll, setHasMeasuredScroll] = useState(false);
+  const now = useNow({ live: true });
+  const todayKey = getUTCDateKey(now);
 
   // Intersection observer to load more interactions when the loader is in view
   const { ref: intersectionRef, inView } = useInView({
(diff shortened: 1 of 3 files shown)


eliasSFL (approved):
LGTM! There were some removal of vars in #7446 that caused tsc issues here. I've patched them up for you

PR #7385, opened 2026-06-29
[FEAT] Automatic Mixer Machine

alexandrdragin:
## Summary
Adds an **Auto Mixer** tab to the Feeder Machine so players can see the combined feed need for animals currently waiting in the opened building and mix only the missing feed in one action.

## Context / motivation
Animal feeding can require repeated manual counting: players need to inspect animals, compare requested feed against basket inventory, then mix the missing food item by item. This PR improves that flow without changing the core feeding model. The Auto Mixer summarizes the current building's feeding demand and turns the missing part into a single mix action.

## Helper logic
The new `getBulkMixRequirements` helper keeps the calculation isolated from the UI:

- works only for the active building context passed by the caller: **Barn** or **Hen House**;
- scans animals in that building and includes only animals that are awake, idle, feedable, and currently asking for food;
- excludes animals covered by free-feed collectibles such as `Gold Egg`, `Golden Cow`, or `Golden Sheep`;
- resolves each animal's current favorite food and boosted feed quantity through existing animal helpers;
- returns three separate values:
  - `requests`: total feed requested by animals waiting in this building;
  - `missingRequests`: requested feed minus feed already available in inventory;
  - `requirements`: raw ingredients and coins needed to mix only the missing feed.

The UI consumes these helper outputs directly, so the modal does not duplicate the calculation rules.

## What changed
- Barn and Hen House now open the Feeder Machine with their own building context.
- Added the **Auto Mixer** tab to the Feeder Machine modal.
- Added a two-column Auto Mixer layout:
  - left side shows `Animals waiting`, `You have`, and `Need to Mix`;
  - right side shows `Ingredients to mix`, `Missing ingredients`, and the `Mix to Feed All` action.
- Empty state now explains that there are no current feed requests and that the mixer can be used when animals need food.
- `Mix to Feed All` mixes only the missing feed for the current building.
- The action is disabled when:
  - no animals currently need feed;
  - the basket already covers all current requests;
  - raw ingredients or coins are insufficient.
- Added localized copy through `dictionary.json`.

## Checked cases
1. **Sleeping / no current requests**
   - No request labels are shown.
   - Empty state is shown.
   - `Mix to Feed All` is disabled.
   
telegram-cloud-photo-size-2-5357534663527831770-y

2. **Current feed exactly equals requests**
   - Requested feed is shown under `Animals waiting`.
   - Existing feed is shown under `You have`.
   - Nothing needs to be mixed.
   - Button stays disabled.
   
[...]

PR #7455, opened 2026-07-18
[FEAT] Feeder Machine: Mix N Needed button

Dionis404:
## Summary
Adds a **"Mix N Needed"** button to the Feeder Machine's existing per-item panel. For the currently selected food/medicine, it calculates exactly how much still needs to be mixed to fully feed the animals in this building that are waiting on it — right down to their next level-up — and mixes precisely that amount in one click.

https://github.com/user-attachments/assets/5d41bec5-e938-44a1-bb45-8d8321ca5554

## Why not #7385's Auto Mixer?
#7385 proposes a separate "Auto Mixer" tab that bulk-mixes every missing food across *all* animals in the building at once, in a single all-or-nothing action.

This PR takes a different approach that gives players more control and variety:

- **Per-food, not all-or-nothing.** Players choose which food to top up (same as today's manual flow), instead of being locked into mixing everything the building needs in one bulk action. If ingredients are short for one food but plentiful for another, the player can still act on what they *can* afford — #7385's bulk button blocks entirely until every ingredient for every food is available.
- **Projects real demand, not just the next feed.** The needed amount accounts for every feed required until the animal's next level-up (not just one round), so the number shown is the amount that actually gets the animal there — not an undercount that needs repeated re-mixing.
- **Omnifeed stays a deliberate reserve.** Omnifeed is never substituted into a regular food's calculation (a player's Omnifeed stash isn't silently "spent" as a planning assumption). It only appears as its own fallback option, and even then only for the portion of demand the player's regular favourite-food stock doesn't already cover — so it never suggests replacing food the player already has on hand.
- **No new UI surface.** Reuses the existing Feeder Machine panel and interaction model instead of introducing a new tab, new bulk-requirements dictionary, or new confirmation flow.

## What changed
- `FeederMachine` / `FeederMachineModal` now take a `building: "Hen House" | "Barn"` prop (passed from `BarnInside` / `HenHouseInside`) so the mixer knows which animals to calculate demand from.
- New `getNeededFeedAmount.ts`:
  - `getNeededFeedAmount` — total quantity of a given food/medicine needed to feed every eligible animal in the building through their next level-up (or, for Barn Delight, every sick animal). Omnifeed's own total is reduced by whatever the animals' actual favourite foods already have in stock.
  - `getFeedShortfall` — the above minus what's already in inventory; floors at 0.
- New "Mix N Needed" button under "Mix 10" — mixes `getFeedShortfall` for the selected item. Hidden entirely (not just disabled) when nothing is needed; disabled when funds/ingredients fall short of that amount.
- New localisation key: `mix.needed`.

## Test plan
- [x] `getNeededFeedAmount.test.ts` — 18 unit tests covering: favourite-food matching, sick/asleep/ready exclusion, until-level-up projection, Barn Delight, and Omnifeed's stock-aware fallback.
- [x] `npx tsc --noEmit` clean.
- [x] Manual: Barn and Hen House, animals in every state (idle/happy/sad/ready/sick/asleep), confirm the button's suggested amount and that it disappears once nothing is owed.

## Summary by CodeRabbit

* **New Features**
  * Added building-specific feed requirement and shortage calculations for the Hen House and Barn.
  * Feeder UI now conditionally shows a “mix needed” action with the required quantity.
  * Added localized messaging for the required mix amount.
* **Bug Fixes**
  * Improved feed planning by accounting for current inventory, upcoming level-ups, and correct handling of Omnifeed and Barn Delight demand.
* **Tests**
  * Added unit tests for feed-quantity and shortfall calculations across key scenarios.

### DIFF
--- a/src/features/barn/BarnInside.tsx
+++ b/src/features/barn/BarnInside.tsx
@@ -231,7 +231,7 @@ export const BarnInside: React.FC = () => {
                   transform: "translateX(-50%)",
                 }}
               >
-                <FeederMachine />
+                <FeederMachine building="Barn" />
               </div>
 
               <MapPlacement
--- a/src/features/feederMachine/FeederMachine.tsx
+++ b/src/features/feederMachine/FeederMachine.tsx
@@ -4,7 +4,11 @@ import { useAppTranslation } from "lib/i18n/useAppTranslations";
 import React, { useState } from "react";
 import { FeederMachineModal } from "./FeederMachineModal";
 
-export const FeederMachine: React.FC = () => {
+interface Props {
+  building: "Hen House" | "Barn";
+}
+
+export const FeederMachine: React.FC<Props> = ({ building }) => {
   const { t } = useAppTranslation();
   const [showFeederMachineModal, setFeederMachineModal] = useState(false);
   const feederMachineImage = SUNNYSIDE.building.feederMachine;
@@ -35,6 +39,7 @@ export const FeederMachine: React.FC = () => {
       <FeederMachineModal
         show={showFeederMachineModal}
         onClose={() => setFeederMachineModal(false)}
+        building={building}
       />
     </>
   );
--- a/src/features/feederMachine/FeederMachineModal.tsx
+++ b/src/features/feederMachine/FeederMachineModal.tsx
@@ -22,10 +22,12 @@ import {
 } from "features/game/types/animals";
 import { Label } from "components/ui/Label";
 import { getIngredients } from "./feedMixed";
+import { getFeedShortfall } from "./getNeededFeedAmount";
 
 interface Props {
   show: boolean;
   onClose: () => void;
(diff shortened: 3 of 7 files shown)


alexandrdragin:
> Projects real demand, not just the next feed. The needed amount accounts for every feed required until the animal's next level-up (not just one round), so the number shown is the amount that actually gets the animal there — not an undercount that needs repeated re-mixing.

strange point and basically no true — automixer count all request until next level-up

Dionis404:
> > Projects real demand, not just the next feed. The needed amount accounts for every feed required until the animal's next level-up (not just one round), so the number shown is the amount that actually gets the animal there — not an undercount that needs repeated re-mixing.
> 
> strange point and basically no true — automixer count all request until next level-up

When I tested your code, I only had the opportunity to mix food for one feeding.

PR #7339, opened 2026-06-15
[FIX] Add farm hand line to new interiors

alexandrdragin:
## Summary
Adds the farm hand / Bumpkin line to the new interior surfaces and anchors it to the actual interior layout bounds instead of hard-coded canvas coordinates. This keeps the line aligned across different island interior shapes and also supports the second floor.

## Context / motivation
The legacy /home interior exposes the farm hand line outside the placeable area so players can access farm hands, outfits, and related placement actions. The new interior experience did not expose that layer, so the same functionality was missing even though beds and farm hands are supported by the underlying placement logic.

## What changed
- Added a reusable interior layout bounds helper based on the active valid tile set.
- Rendered InteriorBumpkins inside a dynamic container around the active interior bounds for the ground floor.
- Rendered the same farm hand line on level_one using the active home expansion bounds.
- Extended InteriorBumpkins so it can target home, interior, or level_one without changing the legacy /home default.
- Restricted navigation to the second floor to volcano island only:
  - HUD floor navigation disables the up button outside volcano.
  - The in-world second-floor arrow is only rendered on volcano.

## How to test
1. Open DevStand with new interiors enabled.
2. Visit /interior on spring, desert, and volcano island presets.
3. Confirm the farm hand line is aligned above the interior bounds and does not overlap the interior art.
4. Confirm second-floor navigation is only available on volcano.
5. On volcano, open /level_one and confirm the farm hand line is aligned above the level-one layout.

## Validation
- Ran npx tsc --noEmit --pretty false.

## Screenshots or screen recording
spring
telegram-cloud-photo-size-2-5296544452887387390-y

desert
telegram-cloud-photo-size-2-5296544452887387391-y

volcano floor 1
image

## Summary

* **New Features**
  * Added Swamp biome support across island rendering, interior backgrounds, placement/anchoring, and metadata (including a new Swamp Biome entry and the Brasil Jersey wearable).
  * Introduced unified news rendering for Community and Discord announcements.

* **Bug Fixes**
  * Level One interior unlocks are now restricted to volcano islands.
  * Interior bumpkins and land-expansion anchoring (including mushroom relocation) now use correct computed bounds/positions.
  * Updated reel-pack gem pricing logic to match the intended “today” rules.

* **Other Changes**
  * Community tab is now always available.

PR #7457, opened 2026-07-20
[FIX] Farm hand line overlapping the Upgrade button on interior floors

Dionis404:
## Summary
- Moved the interior farm-hand/bumpkin line (`InteriorBumpkins`) out of the gameboard canvas and into the HUD (`Hud.tsx` / `LandscapingHud.tsx`), fixed to the viewport instead of the room. Previously the row's container width was derived from the current floor's placeable tile bounds (`bumpkinLayerBounds`), which shrink drastically on early home-expansion tiers (e.g. `level-one-start` is only ~10 tiles wide). With no wrap, a full roster of farm hands overflowed that narrow box and visually crowded on top of the in-world Upgrade button.
- `InteriorBumpkins`: the avatar row now wraps (`flex-wrap`) instead of overflowing, and the row + "Add Farm Hand" button share one wrapping flex container (dropped `justify-between`, which used to fling the button to the far edge of the screen once the row spilled onto a second line).
- `Interior.tsx` / `LevelOne.tsx`: removed the now-dead `bumpkinLayerBounds` positioning code and its layout-lookup imports.
- `Interior.tsx` / `LevelOne.tsx`: the "Import items" button is now anchored *above* the room background's top edge instead of a few pixels *into* it — the mansion/level-one roof art runs flush to the top of its canvas with no built-in headroom, so the old offset landed the button on top of the roof.
- The Upgrade button and the `interior.upgrade` event handler both hard-gated on `island.type === "volcano"`, so ascending to swamp/spooky/etc. after volcano silently took away the ability to unlock further home-expansion tiers. Home expansions unlock at volcano and should carry forward through every later ascension island, matching the existing floor-navigation gate (`hasRequiredIslandExpansion(island.type, "volcano")`) — applied the same check here instead of an exact island match.

https://github.com/user-attachments/assets/e65f3540-4547-4cc6-80bd-9784f13bd24a

## Context
Follow-up to #7339, which added the farm-hand line to the new `/interior` and `/level_one` surfaces but positioned it inside the room itself.

## Test plan
- [ ] Open `/interior` and `/level_one` with a large farm-hand roster (10+) and confirm the row wraps cleanly without overlapping the Upgrade button or the "Add Farm Hand" button.
- [ ] Confirm "Import items" no longer overlaps the room's roof art on `/interior` and on `/level_one` at tiers whose roof runs to the canvas edge.
- [ ] Confirm floor navigation and normal (non-landscaping) placement still work on both floors.
- [x] Added a unit test asserting `interior.upgrade` succeeds on `swamp`; existing pre-volcano rejection test (`spring`) still passes.

### DIFF
--- a/src/features/home/components/InteriorBumpkins.tsx
+++ b/src/features/home/components/InteriorBumpkins.tsx
@@ -102,8 +102,11 @@ export const InteriorBumpkins: React.FC<Props> = ({ location = "home" }) => {
 
   return (
     <>
-      <div className="flex justify-between items-end">
-        <div className="flex">
+      <div className="flex flex-wrap items-end gap-2 max-w-[70vw]">
+        <div
+          className="flex flex-wrap"
+          style={{ rowGap: `${32 * PIXEL_SCALE}px` }}
+        >
           {(!isLandscaping ||
             !bumpkin.coordinates ||
             bumpkin.location !== location) && (
--- a/src/features/interior/Interior.tsx
+++ b/src/features/interior/Interior.tsx
@@ -22,11 +22,7 @@ import {
   NON_COLLIDING_OBJECTS,
   FURNITURE_OBJECTS,
 } from "features/game/expansion/placeable/lib/collisionDetection";
-import {
-  getInteriorLayoutBounds,
-  INTERIOR_CANVAS,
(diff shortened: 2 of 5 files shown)


adamhannigan (approved):
LGTM

PR #7459, opened 2026-07-20
[FIX] Allow home-expansion upgrades on ascension islands past volcano

Dionis404:
## Summary
The Upgrade button and the `interior.upgrade` event handler both hard-gated on `island.type === "volcano"`, so ascending to swamp/spooky/etc. after volcano silently took away the ability to unlock further home-expansion tiers. Home expansions unlock at volcano and should carry forward through every later ascension island, matching the existing floor-navigation gate (`hasRequiredIslandExpansion(island.type, "volcano")`) — applied the same check here instead of an exact island match.

## Context
Split out from #7457 — that commit landed in main before this follow-up commit was pushed, so it never made it into the merge. Cherry-picked from the same branch.

## Test plan
- [x] Added a unit test asserting `interior.upgrade` succeeds on `swamp`; existing pre-volcano rejection test (`spring`) still passes.
- [ ] Confirm the in-world Upgrade button appears and works on swamp (and later ascension islands).

## Summary by CodeRabbit

* **New Features**
  * Interior upgrades are now available starting from volcano island and remain available on later ascension islands.
* **Bug Fixes**
  * Fixed upgrade eligibility so valid later-stage islands are no longer incorrectly blocked.
* **UI Updates**
  * The interior upgrade button now unlocks based on expansion progression (volcano onward), matching the updated availability messaging.
* **Tests**
  * Added coverage for interior upgrades on eligible ascension island types after volcano.

### DIFF
--- a/src/features/game/events/landExpansion/upgradeInterior.test.ts
+++ b/src/features/game/events/landExpansion/upgradeInterior.test.ts
@@ -83,13 +83,21 @@ describe("upgradeInterior (interior.upgrade)", () => {
     ).toHaveLength(1);
   });
 
-  it("rejects upgrade when not on volcano island", () => {
+  it("rejects upgrade when not yet on volcano island", () => {
     expect(() =>
       upgradeInterior({
         state: { ...richVolcanoState(), island: { type: "spring" } },
         action: { type: "interior.upgrade" },
       }),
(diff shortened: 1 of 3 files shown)


eliasSFL (approved):
LGTM! Nice catch!

PR #7412, opened 2026-07-07
[FIX] Improve sick animal sell confirmation

alexandrdragin:
## Summary
Improves the sick animal bounty confirmation flow so reduced rewards are harder to miss and accidental confirmation clicks are blocked briefly.

## Context / motivation

telegram-cloud-photo-size-2-5348113949116803048-y

Sick animals only pay 75% of the normal bounty, but the previous confirmation UI made that discount easy to miss. Players could confirm the sale while focusing on the reward badge and miss that the payout had been reduced. There is also a stale-state corner case: if an animal becomes sick after the modal opens, the player can be looking at old expectations unless the UI re-checks the live animal state before selling.

## What changed
- Adds a danger warning inside the sick animal sell confirmation modal.
- Splits the reward display into two labels: the full bounty is shown struck through, and the actual reduced bounty is shown separately.
- Disables the Confirm button for one second when the confirmation modal opens to reduce accidental taps/clicks.

!!!
- Re-checks the selected animal against the live game state before sending `animal.sold`.
- If animal health changes after the modal opens, the sale is blocked, the displayed bounty is refreshed, and the player must confirm again after the short delay.
- Applies the same confirmation flow to Barn and Hen House by passing the selected animal id into the shared confirmation modal.

image

## How to test
1. Open Barn or Hen House with an animal bounty available.
2. Select a sick animal for sale.
3. Confirm the modal shows the sick-animal warning, the full reward struck through, and the reduced reward separately.
4. Click Confirm immediately after opening the modal and verify the sale does not go through.
5. Wait one second, click Confirm again, and verify the animal is sold for the reduced bounty.
6. For the state-change guard, simulate the selected animal becoming sick after the modal opens and confirm the first click refreshes the modal instead of selling.

## Summary by CodeRabbit

* **New Features**
  * Improved animal sale flow with a brief confirmation step before finalizing deals.
  * Added clearer bounty displays for sick animals, including reduced rewards and original-vs-updated amount comparisons.

* **Bug Fixes**
  * Animal selection now stays tied to the chosen animal’s ID, making sale dialogs more reliable.
  * If an animal’s state changes before selling, the sale view now warns you and recalculates the reward using the latest state.

* **Documentation**
  * Added new on-screen messages for reduced bounty amounts and state-change warnings.

PR #7467, opened 2026-07-22
[FIX] Scope sick animal confirm lock

alexandrdragin:
## Summary

My bad — fix for https://github.com/sunflower-land/sunflower-land/pull/7412

Fixes the sell confirmation regression where the one-second Confirm lock applied to every animal sale instead of only sick animal confirmations.

## Context / motivation
The misclick protection was added for the sick animal reduced-bounty flow, where the player needs a moment to notice the warning and reduced payout. Healthy animal confirmations should keep the original immediate Confirm behavior.

## What changed
- Keeps the one-second Confirm delay only in the sick animal confirmation branch.
- Restores the normal immediate Confirm button for healthy animal sales.
- Leaves the stale-state protection intact: if an animal becomes sick after the modal opens, the UI switches into the sick confirmation flow and the delay applies there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the animal sale confirmation flow so the confirm button can be selected immediately without being blocked by a countdown.

### DIFF
--- a/src/features/barn/components/AnimalBounties.tsx
+++ b/src/features/barn/components/AnimalBounties.tsx
@@ -427,9 +427,7 @@ export const AnimalDeal: React.FC<{
             <Button className="mr-1" onClick={onClose}>
               {t("cancel")}
             </Button>
-            <ConfirmButton key={confirmationKey} onClick={sell}>
-              {t("confirm")}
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):
:shipit:

PR #7191, opened 2026-04-29
[CHORE] Improve marketplace profile navigation

alexandrdragin:
# Pull request description

@adamhannigan Dear sir, its new iteration that idea https://github.com/sunflower-land/sunflower-land/pull/7105
I am really looking forward to your feedback and am open to any ideas

## Summary
This PR restructures the Marketplace profile navigation to create the shortest path to active deals for active traders.

The main problem this solves is that users who actively trade need a shortest route to their current offers and listings, without first landing in a mixed profile page or scanning through unrelated information.

## Changes
— Makes click to My Profile open Active Trades first when the player has active offers or listings.
— Renames the profile trade entry to Active Trades and routes it to the active offers/listings screen.
— Falls back to Stats when there are no active trades, so the first profile click does not open an empty screen.
— Splits trade history into a dedicated Sales History profile route.
— Splits collection into a dedicated profile route.
— Hides Active Trades from the profile submenu when there are no active offers/listings.
— Hides Sales History from the profile submenu when the profile has no trade history.
— Adds empty-state fallbacks for direct URLs, so old links or manual navigation remain safe.

Gives active traders the fastest possible path to their current marketplace work.
Makes the profile section easier to scan by separating active trades, stats, sales history, collection, and rewards.
Avoids showing empty navigation targets when there is nothing useful behind 

## Screensots

Active

Empty state
image

Active state with history 
image

## Release Notes

* **New Features**
  * Added switchable profile views (stats and history) in marketplace profile
  * New dedicated collections and sale history sections accessible in user profiles

* **Improvements**
  * Enhanced filter options with dynamic detection of active trades and history
  * Improved mobile navigation behavior for profile section

PR #7472, opened 2026-07-22
[FIX] "View trades" from a player's profile now opens their sale history

Dionis404:
## Summary
PR #7191 split the marketplace profile into separate routes: the bare `/profile/:id` now defaults to the `stats` view, and `/profile/:id/history` shows sale history.

`PlayerModal`'s "View trades" link was never updated and still navigates to the bare `/profile/:id` route. Clicking it from another player's profile now opens their stats screen instead of their trade/sale history, and there's no way to reach history from there since `MarketplaceUser` only switches view based on the route, not from any in-page control.

## Fix
Point the link at `/profile/:id/history` instead.

## Test plan
- [x] `tsc --noEmit` passes
- [ ] Click "View trades" on another player's profile modal and confirm it opens their Sale History (not Stats)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the “View Trades” link to open the player’s trade history instead of their profile.

### DIFF
--- a/src/features/social/PlayerModal.tsx
+++ b/src/features/social/PlayerModal.tsx
@@ -294,7 +294,9 @@ export const PlayerModal: React.FC<Props> = ({
                 className="text-xxs underline cursor-pointer"
                 onClick={() => {
                   closeModal();
-                  navigate(`${marketplaceBase}/profile/${currentPlayerId}`);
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):
Good catch 👌

PR #7476, opened 2026-07-23
[CHORE] Remove unused i18n translation keys (block 1/10)

Dionis404:
## Summary

This is **block 1 of 10** in a series of PRs removing translation keys from the i18n dictionary that are confirmed unused anywhere in the game's source code. The full dictionary (`dictionary.json`) has 9,484 keys; an initial broad scan flagged ~1,824 unused candidates, which were then independently re-verified down to a confirmed list of ~1,556 truly-unused keys, split into 10 batches of ~155 keys each for safer, incremental review.

This PR removes **155 keys** from `dictionary.json` (the i18n source of truth) and all 11 locale files (`en`, `de`, `es`, `fr`, `id`, `it`, `ja`, `pt-BR`, `ru`, `tr`, `zh-CN`) in lockstep, keeping them in sync — `dictionary.json` is what `TranslationKeys` is derived from (`keyof typeof dictionary.json` in `src/lib/i18n/dictionaries/types.ts`), and `en.json`/other locales are regenerated from it by `_scripts/translate.ts` / the `translate.yml` CI workflow, so all locale files must be edited together.

## How unused keys were found and verified

1. **Initial scan**: scanned every `.ts`/`.tsx`/`.js`/`.jsx` file under `src/` (excluding the dictionaries themselves) for literal-string and template-literal arguments to `t(...)` / `translate(...)` calls (the two entry points into the i18n system — `useAppTranslation()` and the standalone `translate()` helper). Cross-referenced the resulting "used" set against every key in `dictionary.json`.

2. **Independent re-verification pass** (to guard against false positives from dynamic/indirect key usage that a literal-only scan misses): for every candidate flagged unused, searched the entire `src/` tree for the key as an **exact quoted string** in any file that also calls `t()`/`translate()` anywhere in that file. This is deliberately permissive — a coincidental string match in an i18n-touching file is treated as "used" rather than risk a false deletion.
   - This pass caught and corrected **~268 false positives** out of an initial 1,824 candidates, including:
     - `dailyReset.loading.one` through `.four` — used via an indexed array lookup (`LOADING_MESSAGES[loadingMessageIndex % LOADING_MESSAGES.length]` in `DailyResetModal.tsx`), which a plain literal scan missed entirely.
     - Several `linkedAccounts.*` / `gameOptions.blockchainSettings.*` keys selected via ternary expressions (`condition ? "key.a" : "key.b"`) before being passed to `t()` a few lines later.
     - Keys referenced through typed lookup objects like `Record<X, TranslationKeys>` (e.g. `VIP_NAME`, `FOOD_TYPE_TERMS`, `CATEGORY_META[...].titleKey`).
   - Remaining ambiguous matches (coincidental substring/string matches with no real i18n signal — e.g. `"villageProjects"` as a category `id`, `"skip"` as a URL search param, `"chicken"` as an NPC name literal) were manually inspected file-by-file to confirm they are genuinely unrelated to translation lookups before being kept on the removal list.

3. **Post-deletion safety net**: after removing keys, ran `npx tsc --noEmit`. Since `TranslationKeys` is a TypeScript union type derived directly from `dictionary.json`, any surviving `t("removed.key")` call anywhere in the codebase would fail to compile — this is the strongest guarantee available here, and it isn't just theoretical: it caught a **real miss** in this very batch. `"villageProjects"` was initially removed, but is actually used via a `satisfies readonly { id: TranslationKeys; icon: string }[]` array in `chestCategories.ts` (`CHEST_CATEGORIES_META`), consumed as `t(category.id)` in `Chest.tsx`. It was restored before this PR was finalized, and `tsc --noEmit` now passes clean with 0 errors.

4. **Test suite**: ran the full Jest suite after the change — **311/311 test suites passing, 4,636/4,646 tests passing** (7 skipped, 3 todo — pre-existing, unrelated to this change).

## Why in batches of 10

Given the scale (~1,556 keys across 12 files), splitting into 10 smaller, independently-reviewable and independently-tested batches makes each PR easier to review and reduces blast radius if something is later found wrong with one batch — later batches aren't blocked waiting on a large single review.

## Test plan

- [x] All 12 dictionary files (`dictionary.json` + 11 locales) have identical, reduced key counts (9,484 → 9,329 keys each)
- [x] `npx tsc --noEmit` passes with 0 errors
- [x] `yarn test` — 311/311 suites, 4,636 tests passing
- [ ] Reviewer spot-check of the removed key list (diff) for anything that looks like it shouldn't be removed

## Summary by CodeRabbit

* **Localization**
  * Refreshed translations across supported languages for gameplay, menus, onboarding, NPC dialogues, marketplace, trading, delivery, cooking, farming, fishing, and events.
  * Added clearer messaging for quick selection, auction raffles, confirmations, chat status, language changes, and connection issues.
  * Updated item, animal, equipment, and feature descriptions, including seasonal and bonus content.
  * Removed outdated or unused labels and revised village project and feeder-related text.


Dionis404:
Closing to recreate from an up-to-date upstream/main base (the previous branch was based on an older commit and had merge conflicts after new features landed). Will reopen as a fresh PR.

PR #7477, opened 2026-07-23
[CHORE] Remove unused i18n translation keys (block 1/10)

Dionis404:
## Summary

This is **block 1 of 10** in a series of PRs removing translation keys from the i18n dictionary that are confirmed unused anywhere in the game's source code. The full dictionary (`dictionary.json`) currently has ~9,535 keys; an initial broad scan flagged ~1,824 unused candidates, which were then independently re-verified down to a confirmed list of ~1,556 truly-unused keys, split into 10 batches of ~155 keys each for safer, incremental review.

(Note: this replaces an earlier version of this PR, which was based on an older `main` commit and hit merge conflicts once several new features landed upstream. This version is rebased onto current `main` and all 155 keys were re-verified as still unused against the new code before re-applying.)

This PR removes **155 keys** from `dictionary.json` (the i18n source of truth) and all 11 locale files (`en`, `de`, `es`, `fr`, `id`, `it`, `ja`, `pt-BR`, `ru`, `tr`, `zh-CN`) in lockstep, keeping them in sync — `dictionary.json` is what `TranslationKeys` is derived from (`keyof typeof dictionary.json` in `src/lib/i18n/dictionaries/types.ts`), and `en.json`/other locales are regenerated from it by `_scripts/translate.ts` / the `translate.yml` CI workflow, so all locale files must be edited together.

## How unused keys were found and verified

1. **Initial scan**: scanned every `.ts`/`.tsx`/`.js`/`.jsx` file under `src/` (excluding the dictionaries themselves) for literal-string and template-literal arguments to `t(...)` / `translate(...)` calls (the two entry points into the i18n system — `useAppTranslation()` and the standalone `translate()` helper). Cross-referenced the resulting "used" set against every key in `dictionary.json`.

2. **Independent re-verification pass** (to guard against false positives from dynamic/indirect key usage that a literal-only scan misses): for every candidate flagged unused, searched the entire `src/` tree for the key as an **exact quoted string** in any file that also calls `t()`/`translate()` anywhere in that file. This is deliberately permissive — a coincidental string match in an i18n-touching file is treated as "used" rather than risk a false deletion.
   - This pass caught and corrected **~268 false positives** out of an initial 1,824 candidates, including:
     - `dailyReset.loading.one` through `.four` — used via an indexed array lookup (`LOADING_MESSAGES[loadingMessageIndex % LOADING_MESSAGES.length]` in `DailyResetModal.tsx`), which a plain literal scan missed entirely.
     - Several `linkedAccounts.*` / `gameOptions.blockchainSettings.*` keys selected via ternary expressions (`condition ? "key.a" : "key.b"`) before being passed to `t()` a few lines later.
     - Keys referenced through typed lookup objects like `Record<X, TranslationKeys>` (e.g. `VIP_NAME`, `FOOD_TYPE_TERMS`, `CATEGORY_META[...].titleKey`).
   - Remaining ambiguous matches (coincidental substring/string matches with no real i18n signal — e.g. `"skip"` as a URL search param, `"chicken"` as an NPC name literal) were manually inspected file-by-file to confirm they are genuinely unrelated to translation lookups before being kept on the removal list.

3. **Post-deletion safety net**: after removing keys, ran `npx tsc --noEmit`. Since `TranslationKeys` is a TypeScript union type derived directly from `dictionary.json`, any surviving `t("removed.key")` call anywhere in the codebase would fail to compile — this is the strongest guarantee available here, and it isn't just theoretical: it caught a **real miss** in this batch. `"villageProjects"` was initially removed, but is actually used via a `satisfies readonly { id: TranslationKeys; icon: string }[]` array in `chestCategories.ts` (`CHEST_CATEGORIES_META`), consumed as `t(category.id)` in `Chest.tsx`. It was restored before finalizing the list.

4. **Rebase re-verification**: after rebasing onto the latest `main` (which added new features in the interim), re-checked all 155 keys in this batch against the new code to confirm none had become used — none had.

## Test plan

- [x] All 12 dictionary files (`dictionary.json` + 11 locales) have identical, reduced key counts (9,535 → 9,380 keys each)
- [x] `npx tsc --noEmit` passes with 0 errors
- [ ] `yarn test` locally hits an unrelated pre-existing failure in this environment (`SUNNYSIDE.tiles` resolving as `undefined` under the `assets/*` Jest module mock, introduced by the recent Scrabble tiles feature) that reproduces identically on a clean `main` with no changes from this PR — deferring to CI here since `main`'s CI is green
- [ ] Reviewer spot-check of the removed key list (diff) for anything that looks like it shouldn't be removed

## Summary by CodeRabbit

* **Localization**
  * Updated translations across supported languages for clearer, more consistent gameplay and interface text.
  * Refreshed onboarding and character dialogue, including faction, home, and town introductions.
  * Improved wording for trading, marketplace, auction, raffle, chat, and connection flows.
  * Added or revised guidance for cooking, deliveries, farming, fishing, animals, desert activities, and crop machines.
  * Updated confirmation messages, item descriptions, tutorials, and achievement-related notifications.


Dionis404:
@eliasSFL What do you think? Is it worth continuing this purge?

PR #7479, opened 2026-07-24
[FIX] Seed purchases permanently blocked by fractional inventory remainder

Dionis404:
## Summary

A player reported that a historical bug had left their "Apple Seed" inventory count non-integer, and as a result they could no longer buy that seed at all — they noted it was "the same bug as normal buy" (i.e. the same underlying issue also affects the new bulk-buy-seeds feature from #7432).

## Root cause

Three places compare the player's current seed inventory count against the whole-number inventory limit, but only round to 2 decimal places (or don't round at all) before doing so — so a stray fractional remainder in the inventory (seeds are always meant to be whole units) can permanently look like "no headroom left" even though the player is under one whole seed away from the real limit:

- `SeasonalSeeds.tsx` — `setPrecision(inventory[seed], 2)` leaves a fractional remainder that can still be `>= inventoryLimit`, permanently showing `restock.tooManySeeds` and hiding the buy buttons for that seed.
- `planSeedPurchases.ts` (the new bulk-buy planner from #7432) — the exact same pattern, inherited from the single-seed buy flow.
- `seedBought.ts` — the purchase event itself compares `oldAmount.add(amount)` against `inventoryLimit` with no rounding at all, so even if the UI allowed the attempt, the event would still throw `"Can't buy more seeds than the inventory limit"`.

I wasn't able to find a currently-live code path that produces a fractional seed amount in the first place — all quantity-affecting operations on seed inventory I found (`plant.ts`, `bulkPlant.ts`, `fruitPlanted.ts`, `plantGreenhouse.ts`, `supplyCropMachine.ts`, `seedBought.ts`) use whole-number amounts. This strongly suggests whatever historical bug originally wrote a fractional value is long gone from the codebase, but nothing in the current pipeline (there's no generic inventory sanitization on load — checked `processEvent.ts` and `gameMachine.ts`) ever cleans up a stray fractional value once it exists, so it just sits there and permanently blocks purchases via the three comparisons above.

## Fix

Round the inventory amount down to a whole number (0 decimal places, not 2) immediately before each of these three comparisons. Seeds are discrete units, so a fractional remainder below one full seed shouldn't count as "no headroom left" — the fix only changes the comparison, it does not modify or "clean up" the underlying persisted inventory value.

## Test plan

- [x] Added a regression test to `seedBought.test.ts`: buying 1 seed succeeds when inventory is `limit - 1 + 0.5` (a fractional remainder just under the limit), and the resulting inventory is `limit - 1 + 0.5 + 1` as expected.
- [x] Added a regression test to `planSeedPurchases.test.ts`: the bulk-buy planner still produces a purchase plan under the same fractional-remainder scenario.
- [x] Confirmed the existing test asserting a genuine over-limit purchase still throws `"Can't buy more seeds than the inventory limit"` continues to pass — the fix doesn't weaken the real limit check, only its tolerance for sub-1 fractional noise.
- [x] `yarn test` on both affected test files: 48/48 passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Seed purchases and purchasing plans now correctly handle fractional seed remnants, allowing players to buy available whole seeds up to the inventory limit.
  * Market headroom, “inventory full” behavior, and bulk-buy capacity calculations now consistently use whole-unit seed amounts.

* **Tests**
  * Added regression coverage for seed purchase execution and purchase planning when fractional inventory remains.

### DIFF
--- a/src/features/game/events/landExpansion/seedBought.test.ts
+++ b/src/features/game/events/landExpansion/seedBought.test.ts
@@ -141,6 +141,37 @@ describe("seedBought", () => {
     ).toThrow("Can't buy more seeds than the inventory limit");
   });
 
+  it("allows buying up to the limit even if inventory has a stray fractional remainder below it", () => {
+    const sunflowerLimit =
+      INVENTORY_LIMIT(GAME_STATE)["Sunflower Seed"] ?? new Decimal(0);
+
+    // Simulates a historical bug that left a fractional amount in the
+    // inventory (seeds are always meant to be whole units). The player is
+    // still one whole seed under the limit and should be able to buy.
+    const state = seedBought({
+      state: {
+        ...GAME_STATE,
+        inventory: {
+          ...GAME_STATE.inventory,
+          "Sunflower Seed": sunflowerLimit.minus(1).plus(0.5),
+        },
+        stock: {
+          "Sunflower Seed": new Decimal(1),
+        },
+        coins: 100,
+      },
+      action: {
+        type: "seed.bought",
+        item: "Sunflower Seed",
+        amount: 1,
+      },
+    });
+
+    expect(state.inventory["Sunflower Seed"]).toEqual(
+      sunflowerLimit.minus(1).plus(0.5).plus(1),
+    );
+  });
+
   it("subtracts the coins on purchase", () => {
     const coins = 1;
     const balance = new Decimal(1);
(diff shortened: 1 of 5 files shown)


Dionis404:
Good catch — tightened the assertion to `toBe(1)` in 1584efcee, since the fractional-remainder scenario (`limit - 1 + 0.5`) has exactly 1 unit of headroom, and `>= 1` wouldn't have caught a regression that over-allowed.

adamhannigan (approved):
Great fix - those players will be very grateful for you ❤️ I have just added the same patch to the API ✅

PR #7435, opened 2026-07-14
[FEAT] Buy All Tools button in the Workbench

Dionis404:
## Summary
- Adds a "Buy All Tools" button to the Workbench Tools tab that computes the max affordable amount of every unlocked land/water tool, spending from a single shared coin pool and shared ingredient inventory (`planToolPurchases`)
- Adds per-tool Buy All settings (block a tool from bulk-buying, cap max kept in inventory) plus a global on/off toggle for the whole feature, editable via a settings modal and persisted in `GameState.settings.toolShop`
- Confirmation modal shows the tools + total resources that will be spent before purchasing

https://github.com/user-attachments/assets/b937aba1-2eb3-4dbf-9285-3f7a600ec582

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `eslint --fix` clean on all touched files
- [x] Unit tests: `planToolPurchases.test.ts` (15 tests), `updateToolShopSettings.test.ts` (5 tests), existing `craftTool.test.ts` (30 tests) all pass
- [x] Manually verified in the browser (art mode): buying all tools, blocking individual tools, setting max-in-inventory caps, disabling the whole feature, and buttons/layout in the Tools tab and settings modal

## Summary by CodeRabbit

* **New Features**
  * Added batch buying for workbench tools, including customizable quantities, exclusions, and total cost previews.
  * Added an experimental setting to enable or disable batch buying.
  * Added a “craft maximum affordable” option for tools.
  * Added per-tool controls to block tools from batch purchases.
  * Added localized labels and messages for the new purchase options.

* **Bug Fixes**
  * Improved purchase planning to respect available coins, ingredients, stock, and tool requirements.
  * Fixed tutorial progress tracking for crafting multiple axes at once.

PR #7480, opened 2026-07-24
[FIX] Batch Buy modal cut off on mobile, and a stray "Craft 0" button

Dionis404:
## Summary

Three related issues/gaps found while manually testing the Buy All Tools feature (#7435) on a mobile viewport.

### 1. Batch Buy modal layout cut off on mobile

The Land Tools / Water Tools columns in `ToolBatchBuyModal` were both fixed at `w-80` (320px) and laid out side-by-side with no wrapping (`flex items-start justify-center space-x-4`), so on any viewport narrower than ~660px the second column ran off the edge of the screen and was inaccessible — exactly what shows up in the reported screenshot (the Water Tools column cut off on the right on a phone).

**Fix:** the columns now stack vertically below the `sm` (640px) breakpoint (`flex-col sm:flex-row`, `w-full sm:w-80`), matching the responsive pattern already used elsewhere in this codebase (see the seed market's buy buttons in `SeasonalSeeds.tsx`: `flex space-x-1 sm:space-x-0 sm:space-y-1 sm:flex-col`). Desktop layout is unchanged.

While verifying this fix in-browser, a second layout bug surfaced: each column also had its own `max-h-[420px] overflow-y-auto`. On mobile, once the columns stack instead of sitting side by side, this created a nested scroll container that hid the first tool in the list (Axe) above the fold and cut off the total-cost/confirm button beneath it — two independent scroll regions fighting each other. Scoped that inner scroll+height-cap to `sm:` only, so on mobile the whole modal now scrolls as a single unit (using the scroll behavior `Modal.tsx` already provides for the dialog) instead of trapping content inside an inner scroll region.

### 2. Stray "Craft 0" button

Separately, in `Tools.tsx`, the "craft all affordable" button was rendered whenever stock exceeded the standard bulk-craft amount (10), regardless of whether the player could actually afford any at all:

```tsx
{stock.greaterThan(bulkToolCraftAmount) &&
  (() => {
    const craftAllAmount = maxAffordableAmount();
    return (
      <Button disabled={craftAllAmount <= 0} onClick={...}>
        {t("craft")} {craftAllAmount}
      </Button>
    );
  })()}
```

When `maxAffordableAmount()` returns `0` (not enough coins/ingredients for even one), this rendered a permanently disabled "Craft 0" button — inconsistent with how the other buy-amount buttons in this same file already hide themselves entirely when not applicable (e.g. the bulk-10 button only renders `{bulkToolCraftAmount > 1 && ...}`). Now the button simply doesn't render when the affordable amount is 0.

### 3. No purchase confirmation for either bulk-buy path

Both bulk-purchase paths for tools skipped straight to spending coins and ingredients with no confirmation step, unlike the equivalent seed-buying flows in `SeasonalSeeds.tsx`, which already confirm large purchases:

- The Batch Buy modal's "Batch Buy" button executed the purchase immediately on click.
- The single-tool "Craft All" button (the third button, crafting the max affordable amount of one tool) did the same, with no floor — a player could accidentally spend a large amount of resources with one misclick.

Both now show a `ConfirmationModal` (same component/pattern as `SeasonalSeeds.tsx`) before committing:

- **Batch Buy** always confirms, listing how many tool types are being purchased.
- **Craft All** only confirms when the affordable amount exceeds 10 — matching the existing seed-buying threshold, where small amounts don't need confirmation but large ones do.

Both confirmations show the **full cost breakdown** (coins + every ingredient that will be spent), not just the coin total, since a single bulk purchase can consume several different resources at once and a player should be able to see all of it before committing — not just the coin count. Refactored `ToolBatchBuyModal`'s cost breakdown JSX into a shared `costBreakdown` value so the main form and the confirmation modal don't duplicate the same markup.

Added a new `confirmation.craftTools` translation key, and simplified the existing (previously unused) `confirmation.buyAllTools` key by dropping its inline coin amount, since that's now redundant with the cost breakdown shown alongside it in the confirmation body. Only `dictionary.json` was touched for translations, per this repo's i18n pipeline — `en.json` and the other 10 locales are regenerated from it by `_scripts/translate.ts` / the `translate.yml` CI workflow, not edited by hand.

## Test plan

- [x] `yarn test` on the affected area (`workBench`): 12/12 passing.
- [x] Manually verified all three fixes in the running app at a mobile viewport width:
  - Batch Buy modal scrolls as a single unit with Axe visible at the top of the list and the confirm button reachable at the bottom.
  - The "Craft 0" button no longer appears when funds/ingredients are insufficient to craft even one beyond the bulk-10 amount.
  - Both Batch Buy and Craft All (>10) now show a confirmation dialog with the full coin + ingredient breakdown before spending, and cancelling leaves the purchase unmade.

## Summary by CodeRabbit

* **New Features**
  * Added confirmation dialogs for “buy all tools” and bulk crafting actions, with totals shown and confirmation disabled when funds/resources are insufficient.
  * Improved batch-buy tool layout for better responsiveness, including mobile-stacked columns and smoother scrolling on smaller screens.

* **Bug Fixes**
  * Removed the “Craft All” action when no craftable quantity is available, preventing unusable UI.

* **Documentation**
  * Updated tool-related confirmation text in the i18n dictionary for clearer prompts.


Dionis404:
Closing — the mobile layout fix here (columns stacking, scroll containment) is now superseded by #7483, which rewrote `ToolBatchBuyModal.tsx` from scratch (new `ToolRow`/`ToolSection` components, `dvh`-based heights, unified scroll container) and solves the same responsiveness problem in a different way. That PR didn't touch the "Craft 0" button fix or the purchase confirmation modals added here, so those will follow in a fresh PR rebased on the new component structure.

PR #7484, opened 2026-07-27
[FEAT] Add purchase confirmation to Batch Buy and single-tool Craft All

Dionis404:
## Summary

Follow-up to #7483, which rewrote `ToolBatchBuyModal.tsx` (new `ToolRow`/`ToolSection` layout) to fix mobile responsiveness. That PR didn't touch purchase confirmation, so both bulk-purchase paths for tools still skipped straight to spending coins and ingredients with no confirmation step — unlike the equivalent seed-buying flows in `SeasonalSeeds.tsx`, which already confirm large purchases.

(This replaces #7480, which targeted the pre-#7483 `ToolBatchBuyModal.tsx` layout and was closed as superseded once #7483 rewrote that file from scratch. The confirmation-modal and stock-hint work from that PR is redone here against the current component structure.)

### 1. Purchase confirmation

- The Batch Buy modal's "Batch Buy" button executed the purchase immediately on click.
- The single-tool "Craft All" button (crafting the max affordable amount of one tool) did the same, with no floor.

Both now show a `ConfirmationModal` (same component/pattern as `SeasonalSeeds.tsx`) before committing:

- **Batch Buy** always confirms, listing how many tool types are being purchased.
- **Craft All** only confirms when the affordable amount exceeds 10 — matching the existing seed-buying threshold, where small amounts don't need confirmation but large ones do.

Both confirmations show the full cost breakdown (coins + every ingredient that will be spent), not just a coin count, since these purchases can consume several resources at once. Refactored `ToolBatchBuyModal`'s cost breakdown JSX into a shared `costBreakdown` value so the main form and the confirmation don't duplicate markup.

### 2. Stock-shortfall hints

Matching how `SeasonalSeeds.tsx` handles the analogous "shop has more stock than you can afford" case (`bulkSeedBuyAmount` is stock/inventory-limit-based, not money-based, with a "reaching inventory limit" hint underneath when stock exceeds it):

- **Craft All**'s button now always shows the real affordable amount, or the full stock amount (disabled) when nothing is affordable — instead of ever showing a bare "Craft 0" with no context. A hint appears underneath when the affordable amount falls short of the full stock: "Not enough coins or resources to craft all {N} in stock".
- The same per-tool hint now appears in each **Batch Buy** row when that tool's planner-computed max purchasable amount falls short of its full stock, so a player can see at a glance which tools they don't have enough coins/resources to fully restock. Required threading `state.stock` down into `ToolBatchBuyModal` as a new prop.

## Test plan

- [x] `yarn test` on the affected area (`workBench`): 12/12 passing.
- [x] Manually verified in the running app:
  - Batch Buy shows a confirmation with the full coin + ingredient breakdown before spending; cancelling leaves the purchase unmade.
  - Craft All (>10 affordable) shows the same style of confirmation; ≤10 crafts immediately as before.
  - After crafting to the point where funds run out, the Craft All button shows the remaining stock amount (disabled) instead of "Craft 0", with the shortfall hint underneath.
  - Batch Buy rows show the same shortfall hint per-tool when the shared coin/ingredient pool can't cover a tool's full stock.

## Summary by CodeRabbit

* **New Features**
  * Added a two-step confirmation flow for batch tool purchases.
  * Added confirmation for larger “craft all” batches, while smaller batches craft immediately.
* **Improvements**
  * Confirmation dialogs now show itemized coin and ingredient costs.
  * Added stock-aware messaging when funds or resources cannot cover the full quantity.
  * Purchase and crafting actions are disabled when unavailable.
  * Updated localized confirmation and fishing-puzzle guidance messages.


Dionis404:
Good catch — fixed in 703460b42. `buyAllTools()` was calling `onClose()` unconditionally right after `setFailures()`, so the failure Label never had a chance to render (pre-existing from #7435, not introduced here, but worth fixing while we're already touching this function). Now only closes on a fully successful purchase; a partial failure keeps the modal open with the failure message visible.

adamhannigan (approved):
Nice improvement - LGTM ✅ Thanks for the contribution

PR #7497, opened 2026-07-28
[FIX] Interior fixes

adamhannigan:
# Pull request description

**Bug 1: weather item renewed for free**

When you place an item, the reducer looks for an existing (coordinate-less) placement to move, but the list of locations it searched was hardcoded and incomplete: placing on farm only searched ["home", "petHouse"], and placing in home only searched ["farm", "petHouse"] — neither searched interior / level_one.

So moving an item out of the mansion interior fell through to "create a new placement", producing a fresh PlacedItem with a new id and no used flag, while the real used instance stayed parked (unplaced) in the interior., etc.

Fix: derive the search list from PLACEABLE_LOCATIONS minus the target, so every other location is always searched.

**Bug 2: Cooking Pots**
Different mechanism, same "free renewal on re-place" symptom.

Fix: only seed on a first-ever placement (no existing instance in any location).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed monument relocation so previously used monuments do not restart their associated village projects.
  * Improved collectible movement between interior areas, the home, and the farm without creating duplicates.
  * Preserved collectible usage status and identity when moving items.
  * Improved placement reliability when optional interior or pet-house areas are not initialized.

* **Tests**
  * Added coverage for monument relocation and collectible movement scenarios.

### DIFF
--- a/src/features/game/events/landExpansion/placeCollectible.test.ts
+++ b/src/features/game/events/landExpansion/placeCollectible.test.ts
@@ -193,6 +193,73 @@ describe("Place Collectible", () => {
     expect(state.collectibles["Big Orange"]).toHaveLength(1);
   });
 
+  it("does not restart a village project when re-placing a monument", () => {
+    const dateNow = Date.now();
+    const state = placeCollectible({
+      state: {
+        ...GAME_STATE,
+        inventory: {
+          "Basic Cooking Pot": new Decimal(1),
+        },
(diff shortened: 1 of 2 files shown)


eliasSFL:
@adamhannigan You might wanna check for shrines, hourglasses and totems as well, they fall under the sane category

PR #7502, opened 2026-07-29
[FEAT] Group Letter Tiles into their own Chest category

Dionis404:
## Summary

The 26 Letter Tiles (`Letter A Tile` through `Letter Z Tile`, used to spell decorative words like FARM) had no dedicated grouping in `getChestCategories`, so they fell into the catch-all "decorations" category alongside every other unclassified collectible — making them hard to find in a Chest full of decorations once a player owns several tiles.
image

## Changes

- Added a `LetterTileName` type and `LETTER_TILES` lookup object (`Record<LetterTileName, object>`, the same pattern already used for `DOLLS`/`PET_TYPES`/etc.) in `decorations.ts`.
- Added a new `letterTiles` category in `chestCategories.ts`, following the exact same pattern as the other dedicated categories (dolls, banners, weatherItems, ...):
  - Filtered out of `temporaryBoosts`, `boosts`, and the catch-all `decorations` category so tiles aren't double-counted.
  - Own icon (reusing the `Letter A Tile` artwork already in `ITEM_DETAILS`).
  - New `letterTiles` translation key ("Letter Tiles"), added to `dictionary.json` only, per this repo's i18n pipeline (`en.json` and the other 10 locales are regenerated from it by `_scripts/translate.ts`).

## Test plan

- [x] Added a test asserting letter tiles are grouped into their own category and excluded from `decorations`, following the existing test style in `chestCategories.test.ts`. All 4 tests pass.
- [x] Full test suite (`yarn test`): 318/318 suites passing.
- [x] Manually verified in the running app: with a mock inventory containing a few letter tiles plus an unrelated boosted decoration, the Chest now shows a separate "Letter Tiles" category and the tiles no longer appear under "Decorations".

### DIFF
--- a/src/features/game/types/decorations.ts
+++ b/src/features/game/types/decorations.ts
@@ -856,6 +856,63 @@ export const DECORATION_TEMPLATES = {
 
 export type TemplateDecorationName = keyof typeof DECORATION_TEMPLATES;
 
+export type LetterTileName =
+  | "Letter A Tile"
+  | "Letter B Tile"
+  | "Letter C Tile"
+  | "Letter D Tile"
+  | "Letter E Tile"
+  | "Letter F Tile"
+  | "Letter G Tile"
+  | "Letter H Tile"
+  | "Letter I Tile"
+  | "Letter J Tile"
+  | "Letter K Tile"
+  | "Letter L Tile"
+  | "Letter M Tile"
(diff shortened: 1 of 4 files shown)


eliasSFL (approved):
LGTM!

PR #7514, opened 2026-08-01
[FIX] Preserve auction list scroll position

alexandrdragin:
## Problem

Opening an auction or result replaces and unmounts its scrollable list. Returning to the list mounted a new container at the top, forcing players to find their previous place again.

The list now records its current `scrollTop` and restores it when the scroll container mounts again.

## What changed

- added a reusable scroll restoration hook for conditionally rendered scroll containers
- preserved the auction list position when returning from auction details
- applied the same behavior to the auction results list

image

## Player impact

Players return to the same place in the Auctions or Results list after viewing an item.

## Validation

- `yarn tsc --noEmit`
- targeted ESLint on the changed files
- Prettier check on the changed files
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Auction history and auction lists now restore their previous scroll position when revisited.
  * Scroll position is automatically saved as users browse auction content.


adamhannigan (approved):
Thank you! Losing my scroll position in SFL is so annoying 😅

PR #7527, opened 2026-08-05
[FEAT] Context-aware AI translation pipeline (trial)

eliasSFL:
## What

Replaces the per-string AWS Translate pass with a context-aware pipeline that sends each string alongside **its key, its call sites, already-translated neighbours, and a glossary**.

Behind a new `yarn translate:ai`. **`translate.yml` (AWS) is untouched and remains the production path** — nothing changes until we deliberately cut over.

## Why

The legacy pipeline translates one string at a time with zero context. Real output currently live in the dictionaries:

| Key | EN | Output | Problem |
|---|---|---|---|
| `plant` | Plant | ja `プラント` | Industrial plant, not 植える |
| `build` | Build | ja `ビルド` | Software build, not 建てる |
| `on` | On | de `Auf` | Preposition; a toggle needs `An` |
| `coins` | Coins | fr `Monnaies` | "Currencies", not `Pièces` |
| `gems` | Gems | fr `Pierres précieuses` | Gemstones, and overflows the HUD chip |
| `landTools` | Land Tools | es `Herramientas terrestres` | Read "Land" as terrain, not the Land tab |
| `ok` | OK | fr `OK.` | Invented a full stop |

Plus two mechanical bugs: **~30 strings across the locales silently lose a `{{placeholder}}`** (worst is `tr` at 8), and `it.json`/`ja.json` each ship a raw `[0@/$]` marker because the retry path in `_scripts/translate.ts` forgets to restore placeholders.

## How

- **`provider.ts` is the only vendor-specific file.** Structured JSON-schema output, constant schema across requests, retry with backoff.
- **Context**: 7,641 of 9,592 keys resolve to a call site. `t("on")` renders as `{isAudioMuted ? t("off") : t("on")}` — that's the signal that makes it a toggle state.
- **Glossary is derived**, not hand-maintained: it reads `ITEM_DETAILS.translatedName`, so it grows on its own (currently 479 terms for `ja`).
- **Placeholders are verified after the fact.** A string that loses one is retried alone, then falls back to English rather than shipping broken.
- **Serialisation matches the legacy script byte for byte**, so the eventual switchover diff contains only real translation changes.
- `ru` exclusion preserved — flipping that is a product decision, not a refactor.

## Workflow

`translate-ai.yml` is PR-scoped: it runs on each push to an open PR and commits back to that PR's branch, so translation diffs are reviewable before reaching main. A documented `CUTOVER TO PRODUCTION` block at the top explains the flip to push-on-main (uncomment one block, delete `translate.yml`).

The concurrency group is deliberately shared with `translate.yml` on main so the two can never race during a transition.

## ⚠️ Before merging

The `LANGUAGES` / `KEYS` env defaults in the translate step are **temporary** — they force a small `ja,de` run over the known-bad keys so the pipeline can be exercised here (`workflow_dispatch` isn't available until this file is on the default branch, and the dictionary delta is currently empty). Revert both to `${{ inputs.* }}` before merge. They're marked inline.

## Follow-ups (not in this PR)

- 1,058 of 1,541 `ITEM_DETAILS` entries have no `translatedName`; `getTranslatedItemName()` is used in only 18 places. Item-name localisation is a separate piece of work, and it must land **after** this pipeline — otherwise the AWS bot translates 1,058 item names context-free on the next push.

## Summary by CodeRabbit

* **New Features**
  * Added AI-assisted translation with context-aware results, glossary guidance, and language-specific style rules.
  * Added targeted translation, dry-run, result-limit, and alternate output options.
  * Added automated pull-request and manual workflows for updating language dictionaries.

* **Bug Fixes**
  * Improved handling of invalid translations with retries, placeholder validation, and safe English fallbacks.
  * Refined translations across German, Japanese, Spanish, French, Indonesian, Italian, Brazilian Portuguese, Turkish, and Simplified Chinese dictionaries.


eliasSFL:
Parking this for now — closing rather than merging. Branch `feat/ai-translation-pipeline` is left intact so it can be reopened as-is.

**State when parked:** pipeline works end to end. Four workflow runs, all green, each committing back to this branch. Latest run produced a 6-line diff across 5 languages.

**Bugs found by running it, all fixed here:**
- Key reordering — 15,116-line diffs for 3 changed strings
- `ru.json` wipe — the all-languages run replaced all 9,519 Russian strings with English; the excluded-language path overwrote every key instead of only new/changed ones
- Failed translations pinned to English and never retried (`en.json` advanced past them)
- Placeholder validation only checked one direction, so invented `{{placeholders}}` passed

20 tests across `validate.test.ts` and `merge.test.ts`, red-green verified.

**Must do before this ever merges:** remove the temporary `KEYS` override in `translate-ai.yml` — it forces retranslation of a fixed key set. Marked inline.

**Open questions for next time:**
- Translation churn on already-acceptable strings under forced runs (es `gems` Gemas→Gema, tr `coins` Jeton→Altın). Doesn't occur in normal delta-only operation.
- Call-site context is one source line; a 2-line window would give ~30% of bare call sites a JSX element or role prop to work from.
- Item-name localisation (1,058 `ITEM_DETAILS` entries missing `translatedName`) has to land *after* this pipeline, or the AWS bot translates them all context-free.

PR #7531, opened 2026-08-05
[FEAT] Show the new interior when visiting someone's house

eliasSFL:
Visiting a farm always landed on that farm's legacy `/home`, so a player who had opted into the interiors experiment still showed visitors their old room.

The visited farm's data was already on the client — `makeFarmForVisit` doesn't strip `interior` or `settings`, and `makeGame` hydrates both — so this is frontend-only. No BE change.

## Changes

**Routing**
- `getHomeRoute` reads `settings.interiorsEnabled` off the *visited* farm's state (which is what `context.state` holds while visiting), so the house a visitor walks into is the one its owner actually uses. Owners without the experiment still show `/home`, unchanged.
- New `/visit/:id/interior` and `/visit/:id/level_one` routes.
- New `interiorRoutes.ts` helper so every in-world link keeps the `/visit/:id` prefix — a bare `/interior` on a visited farm silently teleports the player into their own house.

**Visiting-safe interiors** — `VisitingHud` in place of the farming `Hud`, no upgrade button, stairs and empty-state buttons routed through the helper, and collectibles / buds / pet NFTs opted into `enableOnVisitClick` to match `/home`.

**Import prompt** — the import-items flow now opens itself the first time a player with un-migrated items walks into their house, instead of waiting to be found in the HUD. Once per page load, so taking the stairs between floors doesn't re-prompt. Still dismissible.

## Note on the route placement

The visiting interior routes are deliberately top-level rather than inside `Game`'s visiting `<Routes>`. That tree renders inside `LandExpansion`'s `ScrollContainer` and `GameBoard`, and `GameBoard` already applies the `ZoomContext` scale — `Interior`/`LevelOne` bring their own scroll container and their own scaled canvas, so nesting them there gives two scrollers and a doubled zoom. There's a comment in `Navigation.tsx` saying so.

## Drive-by fix

The visit farm-id parse took the last URL segment, so it read `"home"` (or `"interior"`) rather than the id and produced `NaN`. `/visit/:id/home` was already broken on reload before these routes existed. Happy to split this out if you'd rather.

## Testing

`getInteriorRoute` / `getInteriorExitRoute` / `getHomeRoute` have unit coverage (10 new cases, both the own-farm and visiting branches). The component layer does not — this repo has no React component tests — so the UI is verified only by `tsc`, eslint and `vite build`. **Worth checking by hand:** visiting someone with the experiment off still lands on their `/home`; the stairs both ways keep the `/visit/:id` prefix; the HUD is the visitor one; and a reload inside `/visit/:id/interior` (which exercises the parse fix).

## Not addressed

The `home.upgrade.*` copy still promises the old home is retired "on August 3rd" and that you can change your mind "before August 3rd". That date has passed and there is no code behind it — `interiorsEnabled` is only ever set by the player, and nothing auto-migrates. This PR makes the nudge harder to miss but doesn't make the deadline real; that needs a product call.

## Summary by CodeRabbit

* **New Features**
  * Added support for visiting another player’s ground-floor and upper-floor interiors.
  * Visitors can interact with collectibles, pets, buds, and stairs while exploring.
  * Added visitor-specific HUDs, navigation, and farm return paths.
  * Home navigation now respects whether interiors are enabled for the relevant farm.
  * Unmigrated home items now prompt players automatically once per page load.

* **Bug Fixes**
  * Corrected visit URL handling for interior subroutes.
  * Improved messaging and navigation when upper-floor interiors are unavailable.


adamhannigan:
Looks like some of the changes from another PR are being included here

adamhannigan (approved):
LGTM ✅

PR #7538, opened 2026-08-08
[Chore] Sell animals UX improvement

alexandrdragin:
https://github.com/user-attachments/assets/075fe6ba-d007-4bc9-97aa-6d083dd9672b

## Why

Selling several animals currently requires repeatedly reopening the bounty menu because selecting an animal moves the interaction back into the Barn or Hen House. Keeping the choice inside the bounty board makes repeated sales faster and makes animal level, XP progress, sickness, and upcoming mutants visible before the player commits.

## Summary by CodeRabbit

* **New Features**
  * Added an expanded, mobile-friendly animal bounty board with grouped rows and expandable bulk bounty cards.
  * Added animal selection with experience sorting, level and XP progress, and warnings for sick animals and upcoming mutant rewards.
  * Integrated bounty selection, weekly reset timing, confirmation prompts, and scrolling into barn and hen house panels.
  * Added dedicated buy, sell, and guide controls with responsive animal purchasing.

* **Bug Fixes**
  * Improved validation for invalid, completed, outdated, or changing bounty transactions.
  * Clarified unavailable and sick-animal sale states with adjusted reward messaging.


adamhannigan:
@alexandrdragin Nice idea to streamline the bounties 👏 

I'm always cautious of adding extra UI/workflows into the game as it adds to the maintenance we need to support.

You mentioned the issue was needing to keep reclicking and opening the bounties to sell. Is there an easier path where we can just reopen the panel after sales?

alexandrdragin:
@adamhannigan at my proposal this bounties window not close after each selling, and it the shortest path that i can find(

adamhannigan:
@alexandrdragin Is there a way that we can reopen the modal after each bounty sale & keep players where they were looking at

alexandrdragin:
@adamhannigan yes, but I don't like this option because we don't know for sure sale was last or not, and forcing  opening a window if the player doesn't need it is also not a good solution for me. 

I totally understand your position and support your consciousness, but basically it is the same flow that we have, but wrapped into one window instead of two. 

I can propose to test it on the beta testers (under beta flag), and after their feedback, we decide whether it should be spread in the whole auditory or not

adamhannigan (changes requested):
I understand the paint point and agree we need a solution for this.

For now I am rejecting these changes as the UI/UX doesn't simplify the process:

- Testing with 30+ animals was difficult to view all of them and select
- Duplicating the animal view + lost some of the positional awareness from the barn view

This is a tough UX challenge and I don't have a solution or specific design for this.

Consider exploring a UI that combines the selling into the main barn area - you could try something similar to how the quick landscaping section works 🙏

alexandrdragin:
@adamhannigan what do you think? Still need some polish, but the core idea, I think, is understandable

https://github.com/user-attachments/assets/4088f636-6f72-4c2a-a0cf-80ebc5fdc26c

alexandrdragin:
mobile


https://github.com/user-attachments/assets/354c1c41-ec6b-4602-9e73-b377506965b3

adamhannigan:
@alexandrdragin Definitely getting closer ❤️ 

I would suggest keeping the Buy/Guide the same modal as it was before. Tap the shop button and those show up in the modal it was before. We don't have the pattern in the repo where the entire modal pins to the bottom so I'd suggest not introducing that.

We can introduce a separate HUD action to activate the bounty selling.

For the bounty selling, it takes up too much space. Are we able to change that layout so less height is taken up? Ideally it should not be taller than the quick design section we have.

Screenshot 2026-08-20 at 11 47 55 am

alexandrdragin:
@adamhannigan Let me check one more time, that I fully understand your idea. Your propos to separate sell tab from the other two tabs and make for it a separate entry point from of the bag button, and from this separate entry would be open sell buttom hub panel

 right?

Dionis404:
@alexandrdragin Has the confirmation of sale been removed?

adamhannigan:
@alexandrdragin Yes that is correct 👍

alexandrdragin:
New entry points
image

New hud 
image

image


@adamhannigan still not sure about icons for selling or buying at the top, but a layout as you asked 

buy / guide as at prod

adamhannigan (approved):
Nice work @alexandrdragin 

I've gone ahead and committed a few small UI tweaks. I've also decided to keep the old tabs for now.

Since this is a mechanic players are familiar with, we will let them use the old selling workflow while they get used to the new one ❤️

PR #7543, opened 2026-08-11
[FIX] Improve bounty bonus claim UX

alexandrdragin:
## What changed

- added explicit in-progress, ready-to-claim, and claimed states to the Bounty Bonus panel
- changed the CTA copy to show the exact reward and replaced the ambiguous disabled `Claim bonus` state with `Bonus claimed`
- added progress and success feedback, including optional confetti after a successful claim
- localized the previously hard-coded Bounty Bonus title and the new state copy
- centralized the weekly claim check so the UI and game event use the same rule
- prevented empty bounty boards and same-timestamp repeat events from being treated as valid claims

## Why

`MegaBountyBoard` used a non-live `useNow()` snapshot and required `bonusClaimedAt` to be earlier than that snapshot. A claim created after the panel opened therefore stayed visually claimable until the panel was reopened. The UI also reused `All Bounties Completed!` after claiming and left the button text as `Claim bonus`, so players did not receive a clear confirmation that the one-time reward was collected.

The event guard also required a repeated action timestamp to be strictly newer than `bonusClaimedAt`, leaving an edge case where two actions with the same timestamp could pass the duplicate check.

## Player impact

- the button and badge now transition immediately after claiming
- players can see bounty progress before the reward is unlocked
- the ready CTA states the exact ticket amount
- the claimed state confirms that the tickets were added to inventory
- duplicate claim attempts are safely rejected at the game-event layer

image

## Summary by CodeRabbit

* **New Features**
  * Added a bounty completion bonus panel showing progress, reward amounts, claim status, and available actions.
  * Added English text for bonus rewards, progress, and claim states.

* **Bug Fixes**
  * Improved weekly bonus eligibility handling, including week boundaries and repeated claims.
  * Rechecks current bounty status before processing a bonus claim.


adamhannigan (dismissed):
Nice addition @alexandrdragin This makes it much clearer, especially for new players understanding how the bonus works

PR #7549, opened 2026-08-12
feat(analytics): tutorial funnel, identity ordering and experiment scaffolding

7hakurg:
Follow-up to the MoonForge analytics integration. This extends the existing instrumentation so onboarding and cohort questions can actually be answered from the data.

All changes are additive and sit behind the existing `src/lib/moonforgeAnalytics.ts` wrapper, which already guarantees analytics can never throw into game code. No game logic, state transitions, or return values are altered.

## Changes

**Tutorial funnel.** `tutorial_step_completed` was being emitted with a single step value in production (`expand_to_3_land`), so there was no funnel to analyse — "where do players drop off during onboarding, and does it differ by platform" had no answer. Steps now come from a fixed exported constant with a derived union type, so an invalid step name fails typecheck rather than silently creating a new funnel stage.

The step vocabulary is sourced from the existing `onboardingAnalytics.logEvent` call sites rather than invented, so it matches milestones already reasoned about in this codebase. Existing `onboardingAnalytics` and `gameAnalytics` calls are untouched.

**Identity ordering.** `mfIdentify`/`mfSetUser` now run first inside `initialiseAnalytics`, ahead of the other analytics providers, so events emitted in that action attach to a durable identifier rather than an anonymous one.

**Experiment scaffolding.** `mfExperiment(experimentId, variant)` emits an `experiment_assigned` event. No call sites yet — it exists so a future holdout test is measurable, which is considerably cheaper to add before an experiment runs than to retrofit afterwards.

## Two corrections made during review

The auth steps originally fired on the `SIGN_IN`/`SIGNUP` transitions — the moment the button is pressed, before the wallet connects or the account exists — despite the event being named `tutorial_step_completed`. They now fire on the success transition (`authorising.onDone`), where a token actually exists.

`connect_wallet` and `create_account` were also modelled as sequential funnel steps when they are mutually exclusive branches, which would have shown a ~100% drop-off at step two for every player. They are now a single `auth` step with a `method` discriminator.

## Known limitation

`session_start` is emitted synchronously by `MoonForgeAnalytics.init()` at app boot, before identification is possible — `mfIdentify` is gated on the async session round trip. Events in that window remain unattributed. Closing it requires a change in the MoonForge SDK (deferred emission or an alias signal) and is not in scope here.

## Tests

53 passing across `src/lib/moonforge`, `src/features/auth`, and `expandLand`. `npx tsc --noEmit` clean.

## Summary by CodeRabbit

* **New Features**
  * Added tutorial progress tracking for wallet connection and account creation after successful authentication.
  * Added tracking for login and land-expansion tutorial milestones.
  * Added analytics support for recording experiment assignments.
  * Improved analytics event delivery during identification and page transitions.
* **Bug Fixes**
  * Prevented failed authentication and analytics errors from recording misleading events.
* **Tests**
  * Expanded coverage for tutorial progress, authentication outcomes, experiment tracking, and event reliability.


adamhannigan (approved):
LGTM

PR #7551, opened 2026-08-12
[FEAT] Add Beta Features panel to Advanced settings

Dionis404:
## Summary
- Beta testers (Beta Pass holders / testnet) can now see what's currently in beta from a new "Beta Features" panel under Advanced settings, instead of discovering beta-gated features by chance.
- The panel lists feature names derived directly from `FEATURE_FLAGS` by identity match against `betaFeatureFlag` (new `BETA_FEATURE_NAMES` export in `flags.ts`), so the list can't drift from the flags it reports on — a new feature flagged with `betaFeatureFlag` automatically appears here with no extra bookkeeping.
- The panel entry point is only visible when `betaFeatureFlag` is true for the player.

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` passes on changed files
- [x] Verified locally on the dev server (testnet network, so `betaFeatureFlag` is true by default): Advanced → Beta Features shows the current beta-gated flags
- [x] Verified the panel entry point is hidden when `betaFeatureFlag` is forced false (temporary local override, reverted before commit)

## Summary by CodeRabbit

- **New Features**
  - Added a Beta Features section to Advanced settings for eligible users.
  - Displays available beta features with descriptions and an empty-state message when none are available.
  - Added a Custom Cursor experiment option under Experiments.
- **Localization**
  - Added English text for beta features, custom cursor settings, connection and offline states, FLOWER withdrawal details, and Pumpkin Pop giveaways.
  - Updated giveaway score formatting and corrected a withdrawal label.


adamhannigan (dismissed):
Nice addition ✅ Very helpful for testers

PR #7553, opened 2026-08-13
[FEAT] Show which boosts are giving chapter-ticket bonuses

Dionis404:
image
image

## Summary
- Chapter-ticket rewards from deliveries, chores, and bounties can be boosted by VIP Access and seasonal wearables/collectibles (`CHAPTER_TICKET_BOOST_ITEMS`), but the boost breakdown was either silently discarded by the UI or never computed at all — players saw a higher ticket count with no way to see why.
- `generateDeliveryTickets` already returned `boostsUsed`; wired it into the two UI spots that were discarding it (`BumpkinDelivery.tsx`, `Orders.tsx`).
- `generateChoreRewards` didn't track boosts at all — extended it the same way (`{ items, boostsUsed }`), updating its two call sites (`completeNPCChore.ts`, `ChoreBoard.tsx`).
- `generateBountyTicket` has 17 call sites relying on its plain-number return, so rather than changing its signature, added a sibling `getBountyTicketBoosts` that mirrors its boost logic for display only (`sellBounty.ts`, `MegaBountyBoard.tsx`).
- All three surfaces wire into the existing `BoostsDisplay` popup via `anchorRef` — the portal-based positioning already used by `CraftingRequirements.tsx` — which also fixes the popup clipping/mispositioning inside scrollable Codex panels.
- `AnimatedPanel`'s backdrop only handled `onClick`, which mobile Safari can drop for a non-interactive `div` (only the trigger itself could close the popup on mobile). Added `onTouchEnd` + `cursor-pointer`, matching the pattern already used in `Modal.tsx`.

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` passes on changed files
- [x] Full `jest` suite passes (328 suites, 4943 tests)
- [x] Verified locally: delivery ticket reward, chore ticket reward, and bounty ticket reward all show a clickable boost breakdown (VIP Access, seasonal wearables) when boosts are active, with no regression when there are none
- [x] Verified the popup no longer clips inside the Codex panel and closes on backdrop click on both desktop and mobile viewport emulation

## Summary by CodeRabbit

* **New Features**
  * Added visibility into boosts applied to chore, delivery, bounty, and ticket rewards.
  * Reward indicators can be selected to open a details popover showing boost names and values.
  * Chapter-ticket boosts are now reflected across relevant reward displays.
  * Boost details can also be opened using keyboard controls for improved accessibility.

* **Bug Fixes**
  * Improved touch interactions when dismissing panels by tapping the backdrop.
  * Preserved existing reward amounts while displaying applied boost information.

### DIFF
--- a/src/features/game/events/landExpansion/completeNPCChore.ts
+++ b/src/features/game/events/landExpansion/completeNPCChore.ts
@@ -1,5 +1,9 @@
 import Decimal from "decimal.js-light";
-import type { GameState, InventoryItemName } from "features/game/types/game";
+import type {
+  BoostName,
+  GameState,
+  InventoryItemName,
+} from "features/game/types/game";
 import {
   getChoreProgress,
   NPC_CHORES,
@@ -151,7 +155,7 @@ export function completeNPCChore({
     // Mark chore as completed
     chore.completedAt = createdAt;
 
-    const items = generateChoreRewards({
+    const { items } = generateChoreRewards({
       game: draft,
       chore,
       now: new Date(createdAt),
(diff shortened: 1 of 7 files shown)


adamhannigan (approved):
Much clearer - cheers!

PR #7558, opened 2026-08-16
[CHORE] Adjust trading limits for some smaller crops

Xeko-92:
The trading limit for some of these smaller crops have a max "stack" valued at less than 1 FLOWER.

## Summary by CodeRabbit

* **New Features**
  * Increased trading limits for Sunflower, Potato, and Rhubarb to 10,000 trades.
  * Increased trading limits for Pumpkin, Zucchini, and Carrot to 4,000 trades.

### DIFF
--- a/src/features/game/actions/tradeLimits.ts
+++ b/src/features/game/actions/tradeLimits.ts
@@ -49,12 +49,12 @@ export type TradeResource = Extract<
 
 export const TRADE_LIMITS: Record<TradeResource, number> = {
   // Crops
-  Sunflower: 4000,
-  Potato: 3000,
-  Rhubarb: 2000,
-  Pumpkin: 2000,
-  Zucchini: 2000,
-  Carrot: 2000,
+  Sunflower: 10000,
+  Potato: 10000,
+  Rhubarb: 10000,
+  Pumpkin: 4000,
+  Zucchini: 4000,
+  Carrot: 4000,
   Yam: 2000,
   Cabbage: 2000,
   Broccoli: 2000,


adamhannigan:
Thanks for the contribution @Xeko-92 

Could you provide more context + the goal of this change. Is this to avoid players doing 2 separate listings?

I would consider putting these higher limits behind a VIP flag to avoid any economic issues

Xeko-92:
> Thanks for the contribution @Xeko-92
> 
> Could you provide more context + the goal of this change. Is this to avoid players doing 2 separate listings?
> 
> I would consider putting these higher limits behind a VIP flag to avoid any economic issues

Well, let me first start by saying, the change here is a bandaid fix. I think the resource marketplace should be changed in 1 of 2 ways.

1) A system where you can buy partial listings. Ideally someone could put up a listing for 1 million Sunflowers, and someone could come along and buy 5,000 of them, then another person could buy 6,000 of them, etc.
2) Tourist has been pushing the idea of an LP for resources, similar I think to what niftyswap had

The change here is just because resource prices have dropped again and now a full stack of some of these crops is below 1 flower (some of the crops are close). And some people have A LOT of some of these smaller crops.

image

adamhannigan:
@Xeko-92 Thanks, that makes sense. We try not to be reactionary to changing prices (it could swing the other way as well).

In this case here I would be ok with those changes, as long as it is reserved behind VIP so the free to players/low reputation don't get access to selling more

Xeko-92:
> @Xeko-92 Thanks, that makes sense. We try not to be reactionary to changing prices (it could swing the other way as well).
> 
> In this case here I would be ok with those changes, as long as it is reserved behind VIP so the free to players/low reputation don't get access to selling more

Unfortunately that is above my paygrade, not sure how to do that. So I'll close this PR as it's not a pressing issue.

PR #7562, opened 2026-08-17
[FIX] Keep monument help icon upright when flipped

alexandrdragin:
## Problem

When a monument is flipped, the help action icon shown to a visiting player is also mirrored. This makes the help affordance appear backwards and is not part of the monument's intended visual flip behavior.

image

## Cause

Collectibles with `flipped` use the global `.flipped img` rule, which applies `scaleX(-1)` to every image nested inside the collectible. The monument's visitor help icon is rendered inside that same subtree.

## Change

Keep the monument sprite mirrored while excluding the visitor help action icon from the collectible image-flip rule. The help icon remains readable in both monument orientations.

## Validation

- `yarn tsc --noEmit`
- `git diff --check`
- Pre-commit ESLint and Prettier hooks

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the appearance and orientation of visitor action icons in monument interactions.
  * Prevented collectible-flipping effects from altering these icons.

### DIFF
--- a/src/features/island/collectibles/components/Monument.tsx
+++ b/src/features/island/collectibles/components/Monument.tsx
@@ -211,7 +211,7 @@ export const Monument: React.FC<MonumentProps> = (input) => {
                   }}
                 >
                   <div
-                    className="relative mr-2"
+                    className="monument-help-action relative mr-2"
                     style={{ width: `${PIXEL_SCALE * 20}px` }}
                   >
                     <img className="w-full" src={SUNNYSIDE.icons.disc} />
--- a/src/styles.css
+++ b/src/styles.css
@@ -1677,6 +1677,11 @@ th {
   transform: scaleX(-1) translateX(50%);
 }
 
+/* Keep visitor action icons readable when their collectible is flipped. */
+.flipped .monument-help-action img {
+  transform: none;
+}
+
 /* Custom pixel art cursor (Sunnyside). While the mouse button is held
    anywhere on the page, html.custom-cursor:active swaps in a variant of the
    same art rotated 20deg counter-clockwise. Its hotspot (8 0) keeps the


Dionis404:
Nice catch on the monument flip bug. One thing worth flagging: this same "visitor help" icon markup also exists in `Project.tsx` (village projects — Big Apple, Big Orange, Big Banana, the Cooking Pots), and it's rendered through the exact same pipeline as `Monument.tsx`:

- Both `Monument` and `Project` are placed via `CollectibleCollection.tsx`'s `COLLECTIBLE_COMPONENTS` map.
- Both get wrapped by `Collectible.tsx:174-177`: `<div className={classNames("h-full", { flipped })}>` — the same global `.flipped img` rule applies to either.
- `Project.tsx:682-696` has the identical help-icon structure (`div.relative.mr-2` with `disc`/`help` icons inside `pointer-events-auto`) that `Monument.tsx` had before this fix.

Since the CSS fix here targets the new `monument-help-action` class specifically (only added in `Monument.tsx`), a flipped village project (e.g. a flipped Big Apple/Big Orange/Big Banana) would still show its help icon mirrored — same bug, uncovered path.

Suggested fix: add the same class to the equivalent div in `Project.tsx` (or rename it to something shared like `visitor-help-action` and apply it in both places), so the CSS rule covers both:

```diff
// Project.tsx
-              <div
-                className="relative mr-2"
+              <div
+                className="monument-help-action relative mr-2"
                 style={{ width: `${PIXEL_SCALE * 20}px` }}
               >
```

Happy to open a follow-up PR for this if useful.

adamhannigan (approved):
Good attention to detail 🙏 Cheers!

PR #7570, opened 2026-08-22
[CHORE] Hide unavailable chapter artefacts in Jafar shop

alexandrdragin:
## Problem

image

## Summary

- hide unavailable chapter artefacts with a zero inventory balance from Jafar's Sell tab
- keep the current chapter artefact visible even when the player does not own it
- keep previous chapter artefacts visible when the player still has copies to sell

image

## Logic

A treasure is hidden only when all three conditions are true:

1. it is a chapter artefact derived from `CHAPTER_ARTEFACT`
2. it is not the current chapter artefact
3. its inventory amount is zero or lower

The Sell tab snapshots owned chapter artefacts when the window opens. If the player sells the final copy of one of those artefacts, its selected card remains visible with a zero balance until the window closes. This avoids moving the selection underneath the player's next click. The list is recalculated the next time the shop opens, so the depleted previous-chapter artefact is then hidden.

## Summary by CodeRabbit

* **New Features**
  * Treasure shops now prioritize relevant chapter artefacts based on the current chapter.
  * Previously owned chapter artefacts remain visible for selling.
  * Treasure selection is preserved while browsing sellable items.

* **Bug Fixes**
  * Prevented unavailable chapter artefacts with no inventory from appearing in treasure shops.
  * Improved treasure sorting and inventory handling for a smoother selling experience.

### DIFF
--- a/src/features/game/types/desert.test.ts
+++ b/src/features/game/types/desert.test.ts
@@ -1,0 +1,63 @@
+import Decimal from "decimal.js-light";
+import {
+  CHAPTER_ARTEFACT,
+  isChapterArtefact,
+  shouldHideChapterArtefact,
+} from "./desert";
+
+describe("isChapterArtefact", () => {
+  it.each([...new Set(Object.values(CHAPTER_ARTEFACT))])(
+    "identifies %s as a chapter artefact",
+    (name) => {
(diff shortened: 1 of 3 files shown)


adamhannigan (approved):
LGTM! ✅

PR #7571, opened 2026-08-23
[FIX] Keep minigame loader stable until iframe loads

alexandrdragin:
## Problem

image

## Summary

- keep the existing minigame loading panel visible while the iframe loads in the background
- reveal the iframe only after its `load` event, preventing the empty collapsed modal frame
- reserve enough loading-panel space for the close button
- show a retry action when portal authentication or iframe loading fails
- accept portal messages only from the current iframe window and expected origin
- fail a loading attempt after 30 seconds so retry remains available

## User impact

Players now see a stable loading panel with an accessible close button until the selected minigame is ready. The transition no longer flashes a compressed empty panel before the iframe renders.

## Validation

- `yarn eslint --quiet src/features/world/ui/portals/Portal.tsx`
- `yarn tsc --noEmit`
- `git diff --cached --check`

## Notes

- The change is shared by all minigames using `Portal`.
- The deployed minigames do not currently emit a render-ready message, so `iframe.onload` remains the supported success signal. Waiting for a new readiness event in this PR would prevent current games from opening.
- DevStand was not started because visual preview was not requested for this implementation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved portal loading reliability and error handling.
  - Added automatic timeout detection for stalled loads.
  - Added a retry option when loading fails.
  - Prevented cancelled or outdated requests from affecting the active portal.
  - Improved security and reliability of iframe communication.

- **UI Improvements**
  - Portal displays clear loading, error, retry, completion, and content states.

### DIFF
--- a/src/features/world/ui/portals/Portal.tsx
+++ b/src/features/world/ui/portals/Portal.tsx
@@ -1,5 +1,11 @@
 import { createPortal } from "react-dom";
-import React, { useContext, useEffect, useRef, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
 import { useActor } from "@xstate/react";
 
 import * as AuthProvider from "features/auth/lib/Provider";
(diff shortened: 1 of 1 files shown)


adamhannigan (approved):
Great fix - this always bugged me while waiting for portals to load 👏

PR #7408, opened 2026-07-06
[FEAT] Speed boosts: barn/hen house animals

eliasSFL:
## Speed boosts: barn / hen house animals (frontend)

9th activity slice of the SPEED_BOOSTS windowed-boost migration. Moves barn/hen-house **animal sleep** from a one-shot baked discount onto the live speed-window model, behind the `SPEED_BOOSTS` flag. Pairs with the BE PR (sunflower-land-api#3474).

### What changes
- **Boosts**: Collie Shrine (Cow/Sheep) + Bantam Shrine (Chicken) sleep `×0.75` → live **1.35× windows** (`ANIMAL_BOOST_SPEED` + `getAnimalBoostWindows`). No totems/hourglasses. The shrines' **feed `×0.95` half stays baked**; El Pollo Veloz −2h + all permanents stay baked.
- **Model**: `Animal.baseDurationMs?` marker (no `boostedTime` — countdown only). `claimProduce` writes `asleepAt`/`baseDurationMs`(shrines-excluded)/denormalised `awakeAt`. New `getAnimalReadyAt(animal, game)` threaded through every raw-`awakeAt` gate (feed, love, wake-up doll, feed-buff, sell/`isValidDeal`, Barnyard Rouse).
- **Love cadence compresses** with the boost (`getNextLoveAvailableAt`/`isAnimalNeedingLove` take `game`). Instant-wake sets `baseDurationMs = 0`; `placeBuilding` Barn/Hen House pause converted to per-animal `pauseWindowedTimer`.
- **UI**: `animalMachine` gets `game` in context + an `UPDATE_GAME` event so a shrine placed mid-sleep recomputes the ready/love times live. Cow/Sheep/Chicken derive `readyAt = getAnimalReadyAt(...)`, tick with `useNow({ live: true, autoEndAt: readyAt })` (stops re-rendering once the animal wakes), and feed the derived time into `SleepingAnimalModal` + the Barn/Hen House hungry/love indicators.
- **Labels**: flag-gated `.speed` variants for the Collie (cow + sheep) and Bantam (chicken) sleep entries in `collectibleItemBuffs.ts` + `dictionary.json` (feed entries stay legacy).

### Verification
- `tsc` source clean; `eslint` clean.
- Full jest suite green (3968 passed). Inverted the two `claimProduce` "shrine reduces sleep ×0.75" tests to the windowed model.

[...]

PR #7578, opened 2026-08-25
[FEAT] Speed boosts: animal sleep

eliasSFL:
Moves barn / hen house sleep onto the speed-rate (`SPEED_BOOSTS`) model. Frontend half of sunflower-land/sunflower-land-api#3555 — **merge the backend first.**

Supersedes #7408 (abandoned earlier attempt, rebuilt from scratch on current `main`).

## What changes

Collie Shrine (Cow/Sheep) and Bantam Shrine (Chicken) are the only *temporary* sleep boosts. Today `claimProduce` bakes their ×0.75 into a stored `awakeAt`, so a shrine placed while an animal is asleep does nothing, and one that expires ten minutes into a 24h sleep still credits the whole sleep.

They now apply as **1.35× rate windows**. Their ×0.95 **feed-cost** half stays baked — the same mixed-boost split Stag and Moth Shrine already use.

The game-logic half mirrors the backend exactly (`ANIMAL_BOOST_SPEED`, `Animal.baseDurationMs`, `getAnimalReadyAt` keyed off the marker rather than the flag, instant wake clearing the marker, `pauseWindowedTimer` on the Barn/Hen House lift, and the love cadence measured against the projected wake time).

## Frontend-only pieces

**`resolveAnimal(animal, game)`** — the one thing worth a close look. `animalMachine` holds only `{ animal }` in context and has no access to game state, so its `isAnimalSleeping` / `isAnimalNeedsLove` guards can't derive a windowed wake time. Rather than add `readyAt` to all ~18 `send({ animal })` payloads across Chicken/Cow/Sheep, each component `useMemo`s a **display-only copy** with the live wake time substituted into `awakeAt` and hands *that* to the machine and the sleeping modal. The machine itself is untouched. It's never persisted or sent — safe because every mutation goes through `gameService` by animal id.

**`SleepingAnimalModal` moves onto `useNodeTimer`**, so "Wakes in" follows the *Show actual time* setting like the other 11 node components: remaining work by default (ticking at the boosted rate, with the ⚡ rate label), real wall-clock countdown when toggled. The love countdown stays a plain `useCountdown` — it's a wall-clock slot, not a work timer.

**The buy panel** (`AnimalBuildingModal`) projects its maturity time and re-names the shrine through a new `getAnimalBoostContributions`, matching what the seed and tool guides already do. That needed `timeSpeed` on `CraftingRequirements`, which now shares `isPreActionBoosted` with `InventoryItemDetails`; the boost breakdown only opens when there are named boosts to list.

**Boost descriptions** flip to `.speed` variants under the flag — `1.35x Cow / Sheep / Chicken Sleep Speed`. The shrines' feed-cost entries stay legacy, since that half is still baked. New keys are in `dictionary.json` only.

## Notes

- `WakesIn.tsx` is dead code (no call sites) and one of the last `TimeLeftPanel` users. Flagged here, deliberately left alone.
- Jest runs amoy, so `SPEED_BOOSTS` is **on** by default in the frontend suite too. The two pre-existing shrine assertions are pinned to mainnet with the `CONFIG.NETWORK` pattern `drillOilReserve.test.ts` uses, and windowed counterparts sit beside them.

## Testing

`tsc --noEmit` and eslint clean. Full `src/features/game`: **301 suites, 4798 tests passing**.

New coverage for the two behaviours the change exists for (a shrine placed *after* the animal fell asleep, and one expiring mid-sleep), flag-off parity, instant wake, the lift, the love cadence on both models, and an anti-drift case asserting the named boost set flattens to exactly `getAnimalBoostWindows` for all three animal types.

Per repo convention there are no component tests; the UI paths are covered by tsc/eslint and in-game checks.

## Summary by CodeRabbit

* **New Features**
  * Added windowed 1.35× sleep-speed boosts for animals from applicable shrines.
  * Animal timers now update as active boost windows change.
  * Added clearer boost details, speed indicators, and localized shrine descriptions.

* **Bug Fixes**
  * Improved accuracy for animal wake, feeding, love, selling, and maturity checks.
  * Preserved sleep progress when moving animal buildings.
  * Fixed incorrect wake times after animals wake or begin sleeping again.
  * Corrected crafting-time boost indicators and breakdown visibility.


src/features/game/events/landExpansion/loveAnimal.ts
```diff
@@ -21,14 +22,25 @@ import { produce } from "immer";
...
+ * players actually schedule around. Required rather than defaulted so every
+ * caller has to decide where its wake time comes from.
  */
-export function getNextLoveAvailableAt(animal: Animal): number {
-  const third = (animal.awakeAt - animal.asleepAt) / 3;
+export function getNextLoveAvailableAt(
+  animal: Animal,
+  readyAt: number,
+): number {
+  const third = (readyAt - animal.asleepAt) / 3;
```
eliasSFL, line 43: **Bug: an already-spent love slot re-opens when the shrine window shrinks mid-sleep.**

The cadence is now thirds of a value that can *grow* after slots have been consumed. Let `R = readyAt - asleepAt` and `L = lovedAt - asleepAt`. A slot is open iff `L + R/3 < R`, i.e. `L < 2R/3`. After the second love, `L == 2R₀/3` exactly — so **any** later increase in `R` re-opens a slot.

Concrete (verified numerically against `computeReadyAt`), 24h base chicken with a Bantam Shrine covering the whole sleep:

```
R0 = 17.78h, third = 5.93h
love 1 @ 5.93h, love 2 @ 11.85h        -> next = 17.78h == readyAt, no third love ✓
player LIFTS the Bantam Shrine at 12h  -> window truncates at removedAt
R1 = 19.80h, third = 6.60h
next = max(6.60, 11.85 + 6.60) = 18.45h < 19.80h  -> THIRD LOVE ACCEPTED ✗
```

`loveAnimal` accepts it, so it's free animal XP, repeatable every night by toggling the shrine in landscaping.

This is the same failure the `placeBuilding` hunk in this PR guards against — and that guard doesn't hold either for a windowed animal: shifting `asleepAt` also changes the overlap with a shrine window whose `to` falls near the sleep, so `R` moves and a spent slot re-opens. The new regression test only exercises a farm with no shrine placed.

Suggest anchoring the cadence to something stamped at sleep (thirds of `baseDurationMs`, or a `loveCadenceMs` on the animal) rather than a quantity that can grow after slots have been spent.

src/features/island/buildings/components/building/barn/Barn.tsx
```diff
@@ -268,7 +270,7 @@ export const Barn: React.FC<BuildingProps> = ({ isBuilt, island, season }) => {
       !isAnimalCoveredByGoldenAsset({
         state: game,
         animalType: animal.type,
-      }) && isAnimalReadyForLove(animal, now),
+      }) && isAnimalReadyForLove(animal, now, getAnimalReadyAt(animal, game)),
```
eliasSFL, line 273: **Perf: the shrine window set is rebuilt per animal, inside per-second render loops.**

`getAnimalReadyAt` calls `getAnimalBoostWindows` → `getBoostWindows`, which does a `getCollectiblesAcrossLocations` scan (4 array spreads), a filter, a `boostHistory` map and a `mergeWindows` sort, plus `getExpiryCooldown`/`hasFeatureAccess`. But only **two** distinct window sets exist for a given game state — Bantam for `Chicken`, Collie for `Cow`/`Sheep`.

This file drives `useNow({ live: true })`, so it re-renders once a second for as long as the island is open, and each render now runs that rebuild once per animal for the love alert (this line) — same in `HenHouse.tsx:63`. `_hasHungryAnimals` / `_hasHungryChickens` (`Barn.tsx:234`, `HenHouse.tsx:27`) repeat it on every gameService emission. With a stocked barn *and* hen house that's O(animals) collectible scans + sorts every second, forever, to recompute two identical results.

Same shape elsewhere in the PR:
- `FeedAllButton.tsx` — `.map(animal => getAnimalReadyAt(animal, game))`
- `BarnInside.tsx` / `HenHouseInside.tsx` — `isValidDeal({ animal, deal, game })` per animal, per render
- `getFeedAllTargets` — per-animal rebuild against an **immer draft** in the reducer (draft proxy traps on every collectible read)

Hoisting `getAnimalBoostWindows(game, type)` once per animal type (or memoising it) and threading the windows down would collapse all of these to two builds.

src/features/barn/components/Cow.tsx
```diff
@@ -104,8 +105,12 @@ export const Cow: React.FC<{ id: string; disabled: boolean }> = ({
 }) => {
   const { gameService, selectedItem, shortcutItem } = useContext(Context);
 
-  const cow = useSelector(gameService, _cow(id));
+  const storedCow = useSelector(gameService, _cow(id));
   const game = useSelector(gameService, _game);
+  // The animal machine has no access to game state, so every consumer below —
+  // the machine included — is handed the animal with its live windowed wake time
+  // substituted in. Read-only: nothing here writes an animal back.
+  const cow = useMemo(() => resolveAnimal(storedCow, game), [storedCow, game]);
```
eliasSFL, line 113: **The machine's copy of the animal can hold a stale `awakeAt` for the rest of the sleep.**

`resolveAnimal` makes `awakeAt` a *derived* value that can now change mid-sleep (it moves whenever a Collie/Bantam Shrine is placed, lifted or burned). But `animalMachine` only ever gets a fresh animal via an `assign` on FEED / LOVE / CLAIM_PRODUCE / CURE / INSTANT_WAKE_UP, and its `sleeping` state TICKs once a second evaluating `isAnimalSleeping` / `isAnimalNeedsLove` against `context.animal.awakeAt`.

The only effect watching the derived value fires on a single condition:

```ts
useEffect(() => {
  if (cow.awakeAt < Date.now() && cowMachineState === "sleeping") {
    cowService.send({ type: "INSTANT_WAKE_UP", animal: cow });
  }
}, [cow.awakeAt]);   // Cow.tsx:127, Sheep.tsx:139, Chicken.tsx:138
```

So when `awakeAt` changes but is still in the **future**, the effect runs, does nothing, and the machine keeps the old value — and since `getAnimalReadyAt` is a pure function of `(asleepAt, baseDurationMs, windows)`, nothing later re-fires it. Result, with the interior mounted and the shrine set changing underneath it (another session, or a server-refreshed farm):

- readyAt moved **earlier** → the sprite ticks against the old, later value and stays visually asleep past its real wake time.
- readyAt moved **later** → the machine wakes the sprite early, and the first feed click throws `"Animal is asleep"` from the reducer (`feedAnimal` uses the live `getAnimalReadyAt`).

Sending the resolved animal into the machine on *any* `awakeAt` change — not just one already in the past — would keep the guard in step with the reducer.

PR #7579, opened 2026-08-25
[FEAT] Wasm token approach

adamhannigan:
# Pull request description

POC of using WASM tokens for added security 🔒


adamhannigan:
Didn't work

PR #7592, opened 2026-08-31
[FEAT] Load saved layouts lazily from the layouts endpoint

eliasSFL:
## Summary

FE half of sunflower-land/sunflower-land-api#3569, reworked per Adam's review: `GameState.layouts` is deleted, the five layout reducers are gone, and every layout action is an explicit async effect.

## Changes

- **Saved Layouts modal** (still in landscaping, still `SAVED_LAYOUTS`-gated): holds the fetched list in local state (`loadLayouts` on first open, loading/retry UI) and posts `layout.created/edited/deleted/applied` via `postEffect` directly — the established non-machine-effect path (`StateMachineEffectName` exclude list), which is what makes effects usable from the `landscaping` state. Every mutation first awaits `flushPendingActions` (new helper: triggers/waits on the landscaping child's autosave, or the parent's in `playing`) so server snapshots and applies see the current farm. Toasts moved from optimistic to post-response; layouts are addressed by their stable server ids.
- **`LAYOUT_APPLIED`** machine event (assign-only, handled in `playing` + `landscaping`) pushes the farm returned by `layout.applied` into the machine; the apply toast shows the server's `applied/skipped/noInventory` counts.
- **Island upgrade prompts** (`IslandUpgrader`): ascension confirmations gain a "save my current layout" checkbox (creates a normal cap-counted layout marked as the `ascensionLayoutId` target; replaced by a hint when slots are full), and on later ascensions — which now reset to the initial land instead of silently re-applying a protected auto layout — the post-upgrade modal offers a one-tap "Re-apply '<name>'" button.
- `upgradeFarm` sheds all layouts logic (mirrors the BE); the protected `auto` layout concept and its modal special-casing are retired (the ascension-target layout just gets a badge).
- `applyLayout.test.ts` ported to `lib/layouts.test.ts` (22 tests against `applyFarmLayout` directly); new/updated i18n keys in dictionary.json only.

## Rollout

Release the API PR first (after its migration seed phase), this ~5 min after.

## Testing

Full FE suite green: 338 suites / 5,207 tests; `tsc --noEmit` + eslint clean. In-browser on testnet still to verify: modal CRUD from landscaping (including pending-moves-then-save), apply without leaving landscaping, and both ascension prompts.

## Summary by CodeRabbit

- **New Features**
  - Saved layouts load from the server when the panel opens.
  - Create, rename, overwrite, apply, and delete saved layouts.
  - Ascensions reset the farm, with options to save and reapply a layout afterward.
  - Added layout slot status and ascension-layout indicators.
  - Added saved-layout support in offline art mode.

- **Bug Fixes**
  - Improved synchronization before upgrades and layout application.
  - Added clearer handling for layout action failures.
  - Applying a layout now consistently updates the farm state.


adamhannigan (approved):
Tested and working well 👏 ✅

PR #7598, opened 2026-09-01
[FIX] Clarify crustacean chum selection

alexandrdragin:
## Problem

The water trap modal only showed chum types that the player already owned. This made it difficult to understand which chum options were supported, inspect their requirements, or see a previously discovered catch when the required chum was not currently available.

Placing a trap without chum was also supported, but the only way to reach that state was to click the selected chum again. That hidden deselection behavior made the chum-free option easy to miss.

## Changes

- Show every chum supported by the selected water trap, including chum with a zero inventory balance.
- Allow unavailable chum to be selected for inspection while keeping the Place action disabled until the required amount is available.
- Keep stored chum selection valid based on trap compatibility instead of current inventory quantity.
- Replace implicit click-again deselection with an explicit empty selection cell at the end of the chum list.
- Label the empty option as `Without chum` and add a short description.
- Show `Undiscovered` for combinations whose resulting catch has not been discovered, without revealing the catch.
- Preserve the existing discovered-catch label and chum requirement behavior.

image
image

## Validation

- Prettier
- Targeted ESLint
- `yarn tsc --noEmit`
- `git diff --check`
- DevStand preset on Volcano with one of each water trap and a mix of available and unavailable chum; HTTP 200
- Iterated using user-provided DevStand screenshots and feedback

## Summary by CodeRabbit

* **New Features**
  * Players can now place water traps without owning or selecting chum.
  * Added a dedicated option to clear the selected chum.
  * Added guidance explaining the no-chum option.
  * Unknown potential catches now display as “Undiscovered.”

* **Updates**
  * The water trap chum panel now consistently displays all allowed chum options.


Xeko-92:
I believe you're making these changes due to the message I posted in the wishlist channel, and I appreciate that, but the main issue is still that we can't see how many crustaceans we have and this doesn't really fix that issue. The crustacean catching process needs a pretty big UX rework.

alexandrdragin:
@Xeko-92 can you please specify the user story when you need know all amounts of crustaceans until you start catching some?

Xeko-92:
> @Xeko-92 can you please specify the user story when you need know all amounts of crustaceans until you start catching some?

I'm not sure I understand the question. What I'm saying is that a small change doesn't really fix the core UX problems that the crustacean system is facing.

alexandrdragin:
@Xeko-92 my question is about why exactly you need to know all numbers of your crustacean before catching some. For example, the interface for fish catching of flowers planting doesn't answer that question.

Xeko-92:
> @Xeko-92 my question is about why exactly you need to know all numbers of your crustacean before catching some. For example, the interface for fish catching of flowers planting doesn't answer that question.

I'm not really worried about the process before catching some. That lasts for 1 day for most people. The problem is the on-going process of catching them.

alexandrdragin:
> The problem is the on-going process of catching them.

@Xeko-92 im totaty lost
 please tag me (@crimewaveboat) at any discord tread and show me how cathing crustacean may > lasts for 1 day for most people

src/features/island/fisherman/WaterTrapModal.tsx
```diff
@@ -17,7 +17,7 @@ import {
   getChestItems,
 } from "../hud/components/inventory/utils/inventory";
 import { SUNNYSIDE } from "assets/sunnyside";
-import type { GameState, WaterTrap } from "features/game/types/game";
```
AhmedMuhammedMuhammed4, line 20: 2f0dcf1d95f00e4ed618f3407d012c0374e90e8a

alexandrdragin: Could you clarify what change you’re suggesting here? The comment currently contains only the SHA of this PR’s existing commit.

PR #7634, opened 2026-09-07
[FIX] Cancel Fish Market items from unordered queues

alexandrdragin:
## Problem

https://discord.com/channels/880987707214544966/1347190192745742336/1546507600151974048

Cancelling a queued Fish Market product can stop working after the game is refreshed. The confirmation modal closes, but the selected product remains in the queue and its ingredients are not refunded.

The Fish Market UI displays a copy sorted by `readyAt`, while the cancellation reducer previously looked up `productIndex` in the stored array and then sorted that same array in place while finding the active product. When a server-loaded processing array was not chronological, the saved index no longer referred to the selected product. The reducer could therefore treat the wrong product as active, refund the wrong requirements, or remove the wrong queue entry.

## Solution

- Create a chronological copy of the processing queue before identifying the selected and active products.
- Stop sorting the reducer queue in place inside `getCurrentProcessingItem`.
- Pass the same ordered queue through identification, removal, and queue-time recalculation so indices remain stable.
- Add regression coverage using a non-chronological server-loaded Fish Market queue and verify the selected product, refund, and recalculated ready times.

## Validation

- `yarn test src/features/game/events/landExpansion/cancelProcessedResource.test.ts src/features/game/events/landExpansion/speedUpProcessing.test.ts src/features/game/events/landExpansion/processResource.test.ts src/features/game/events/landExpansion/collectProcessedResource.test.ts --runInBand` (42 tests passed)
- `yarn tsc --noEmit`
- ESLint on the two changed files
- Prettier check on the two changed files
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed resource cancellation when the processing queue is not in chronological order.
  - Ensured the selected product is removed, requirements are refunded, and remaining items retain the correct order.
  - Corrected recalculation of processing start and ready times after cancellation.

- **Tests**
  - Added regression coverage for cancellation and queue reordering scenarios.

### DIFF
--- a/src/features/game/events/landExpansion/cancelProcessedResource.test.ts
+++ b/src/features/game/events/landExpansion/cancelProcessedResource.test.ts
@@ -250,4 +250,79 @@ describe("cancelProcessedResource", () => {
       readyAt: firstReady + DURATION("Fish Oil"),
     });
   });
+
+  it("cancels the selected product when the server queue is not chronological", () => {
+    const firstReady = createdAt + DURATION("Fish Flake");
+    const secondReady = firstReady + DURATION("Fish Stick");
+    const thirdReady = secondReady + DURATION("Fish Oil");
+
+    const state: GameState = {
+      ...BASE_STATE,
+      buildings: {
+        "Fish Market": [
+          {
+            ...(BASE_STATE.buildings["Fish Market"]?.[0] as PlacedItem),
+            processing: [
+              {
+                name: "Fish Oil",
+                readyAt: thirdReady,
+                startedAt: secondReady,
+                requirements: { Anchovy: new Decimal(4) },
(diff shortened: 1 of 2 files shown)


eliasSFL:
Thanks a lot for digging into this and for the clear write-up and regression test! 🙏

We looked into it further with a reporter's farm, and it turned out their queue was already in chronological order, so the unordered-queue path wasn't what players were hitting. The real cause was elsewhere: the ingredients stored on each queued item (used for the refund) weren't being saved in a format the game could read back after a reload, so the refund step failed and the cancel was dropped. That's been fixed on our side, so this change isn't needed anymore and we're closing it.

Really appreciate the contribution!

PR #7420, opened 2026-07-08
[FEAT] Add MoonForge analytics and error tracking

7hakurg:
## Summary

Adds MoonForge analytics and error tracking to the game. A self-contained SDK
(`src/lib/moonforge/`, no new npm dependencies) plus a thin typed wrapper
(`src/lib/moonforgeAnalytics.ts`) instrument the core gameplay loop — 13 events
(harvest, plant, gather, craft, build, deliveries, level-ups, achievements,
marketplace sells, faction joins, IAP, auction bids), player identity, per-scene
error context, and automatic capture of sessions, screen views, unhandled JS
errors, and failed API calls. It runs alongside the existing GameAnalytics,
Firebase, and Meta pixel integrations — nothing is replaced or removed, and all
calls are fire-and-forget no-ops if initialization fails, so gameplay is never
affected.

## Context / motivation

We are onboarding MoonForge for gameplay analytics and error monitoring
(retention/funnel views per scene, plus crash and API-failure visibility with
game-state context). The existing GameAnalytics setup stays as-is; this adds the
MoonForge pipeline next to it using the same wiring points
(`gameMachine.initialiseAnalytics`, the `mmoMachine` scene switch), following
the conventions of `src/lib/gameAnalytics.ts`.

## How to test

1. `yarn install && yarn dev`, open the game in a browser.
2. Open DevTools → Network, filter by `collector.moonforge.co`.
3. On load: one `POST /api/send` with `session_start` returns 200 (response
   contains a `cache` token + `sessionId`).
4. Walk between world scenes (e.g. Plaza → Beach): a `screen_view` event fires
   per scene change.
5. Harvest a crop / chop a tree / place a building: `crop_harvested` /
   `resource_collected` / `building_placed` events fire with their properties.
6. Type-check: `npx tsc --noEmit` passes (strict mode, SDK ships with its own
   `.d.ts`).
7. Optional: throw a test error in the console — a `POST /api/errors` fires with
   the current scene in `gameState`.

## Screenshots or screen recording

N/A — no UI or copy changes; all additions are analytics calls and a library
folder.

---

### PR and scope

- [x] Title is clear and uses a prefix: `[FEAT]`, `[CHORE]`, or `[FIX]`
- [x] I have read the contributing guidelines and agree to the T&Cs

### Code quality

- [x] I have performed a self-review of my own code
- [x] I have commented code only where it helps future readers (non-obvious logic, invariants, gotchas)
- [x] I have updated documentation if behaviour or public APIs changed
[...]

PR #7641, opened 2026-09-08
feat(analytics): revenue, FTUE and economy_transaction instrumentation

7hakurg:
Follows on from #7420 and #7549. Those added the analytics wrapper, the
tutorial funnel and identity ordering. This one covers three questions the
current events cannot answer: what players spend real money on, where new
players drop out before the first session ends, and where in-game currency
enters and leaves the economy.

## Revenue

`iap_initiated` and `iap_completed` wrap the Xsolla card flow in
`CurrenciesModal`. Both carry the same transaction id, generated once when
checkout starts and held in a ref.

The shared id matters because Xsolla can fire its success callback more than
once. Without a stable id the same purchase is counted as two sales, and
revenue numbers drift upward with no obvious cause. The id also matches the
payment record, so a figure can be traced back to a specific transaction
rather than taken on faith.

## First-time user experience

`account_created` fires once per account, from the game machine, after the
player is identified. The auth machine leaves a marker at `creating.onDone`.

That location is deliberate. `authorising` only means the player was
authorised, and a welcome-screen button press only means they intended to
sign up, so emitting from either would count accounts that do not exist. The
marker is keyed to the farm id carried on the signup token, so a second
account created on a shared device does not inherit the first one's marker.

`signup_method` is a fixed set: social, email, platform, grant, or
guest_upgrade. Social carries the provider alongside it.

## Economy

`economy_transaction` fires wherever currency is created or destroyed:
crafting, cooking, mining, deliveries, land expansion, shop purchases and
reward claims. Each event carries the delta and the resulting balance, with
inputs and outputs recorded separately so a single craft shows both what it
consumed and what it produced.

The emitters live in the individual game events rather than in a shared
helper. A helper sees only that a balance moved; the event knows whether it
was a sale, a craft or a claim, which is the part worth recording. This is
the same reasoning behind the domain-level placement suggested during review
of the API-side work.

## Scope and safety

Every call is fire and forget and cannot throw into gameplay, matching the
existing wrapper. Where a call site turned out to be unreachable it was
removed rather than kept for symmetry, so the emitters map to paths players
actually take.

Nothing existing changes behaviour. The only non-analytics edit is a ref in
`CurrenciesModal` to hold the transaction id.

## Testing

46 tests in `moonforgeAnalytics.test.ts`, `moonforgeSessionBuffer.test.ts`
and `moonforgeTutorial.test.ts`, covering the double-fire de-duplication, the
farm scoping on the signup marker, and event ordering against the identify
buffer. `tsc --noEmit` is clean and the 61 changed files add no new eslint
warnings against the baseline.

`src/features/game/events` has 58 failing tests across 9 suites. These fail
identically on `main` without this branch applied and are unrelated.

## Summary by CodeRabbit

- **Analytics**
  - Expanded tracking for in-game economy activity, including purchases, crafting, harvesting, mining, exchanges, rewards, sales, and speed-ups.
  - Added detailed before-and-after currency and inventory balance reporting.
  - Added signup, account creation, tutorial progress, and referral attribution tracking.
  - Improved purchase analytics with initiation and completion events linked by transaction.
- **Reliability**
  - Analytics events now fail safely without interrupting gameplay or purchases.

### DIFF
--- a/src/features/auth/lib/authMachine.ts
+++ b/src/features/auth/lib/authMachine.ts
@@ -13,6 +13,7 @@ import {
   trackTutorialStep,
   type TutorialAuthMethod,
 } from "lib/moonforgeTutorial";
+import { markSignupPending } from "lib/moonforgeAnalytics";
 import type { loadSession } from "features/game/actions/loadSession";
 import { getToken, removeJWT, saveJWT } from "../actions/social";
 import { signUp, type UTM } from "../actions/signup";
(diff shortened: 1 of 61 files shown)


7hakurg:
Hello @adamhannigan - requesting merge of this PR to main for us to collect a few additional info to improve MoonForge insights.

franksunnyflower (changes requested):
Thanks for the contribution @7hakurg 

The `game/events` code is used exclusively for functional game events - it cannot perform anything asynchronous or send off additional side effects.

If you are looking to fire off side-effects/analytics, I would consider looking into setting it up elsewhere (see autosave.ts) or in a separate Xstate event listener

PR #7655, opened 2026-09-09
[FEAT] Complete the big fruits instantly instead of through the server

eliasSFL:
Backend: sunflower-land/sunflower-land-api#3610 — **must deploy first.**

## Why

Completing a village project went through the `project.completed` **effect** for every project. That is only necessary for the cooking pots: completing one also airdrops its raffle winner a food box, and writing to another player's farm can only happen server side.

The big fruits (Big Orange / Big Apple / Big Banana) share that handler but touch nothing except the owner's own inventory — so the modal sat on a loading spinner and then a success screen for a reward that was already earned.

## What

The two halves are now fully disjoint:

- **`project.collected`** — a new local event. Banks the reward, tracks the farm activity, clears the village project and adds it to `completedProjects`. Same work as the effect, minus the airdrop, and it syncs on the next autosave.
- **`project.completed`** — keeps the cooking pots, the only projects that reach off the farm, and keeps the loading and success screens with them.

`RaffleProjectName` and `SoloProjectName` split `VillageProjectName` in two, and each path takes only its own half. `RAFFLE_REWARDS` is now exhaustive over `RaffleProjectName`, which makes `hasProjectRaffle` a real type guard (`project is RaffleProjectName`) — both the completion branch and the modal's raffle-winner section go through it instead of hand-rolling the check.

The reducer still rejects a raffle project at runtime. The action is JSON on the wire, so the type is not the thing enforcing this; the backend mirrors the same rejection.

Completing a fruit now just closes the modal — the reward toast fires off the inventory diff like every other instant event.

## Drive-by fix

`getProjectReward` returned `BoostName[]`, which `updateBoostUsed` does not accept, so the Cornucopia boost was never recorded on this side. It now returns the `{ name, value }` shape every other call site uses. Nothing consumed the old return value.

## Tests

`collectProject.test.ts` adds 9 cases: not started, not enough cheers, raffle rejection, reward, stacking onto an existing amount, the Cornucopia +1, project cleared and marked completed, no duplicate entry, farm activity.

- `npx tsc --noEmit` clean
- The 11 monument/project suites pass (123 tests)

## Deploy

Not backwards compatible in either direction: the backend effect now refuses a big fruit, and `project.collected` needs the backend deploy to validate. A stale tab gets an error until it refreshes, which the new-version refresh prompt covers.

## Summary by CodeRabbit

- **New Features**
  - Added support for collecting completed solo village projects.
  - Rewards, bonus boosts, completion progress, and activity records are now updated when a project is collected.
  - Collected projects are removed from active project listings and duplicate collection is prevented.
  - Project details close automatically after successful collection.
- **Bug Fixes**
  - Added validation to prevent collecting missing, incomplete, or unsupported projects.

### DIFF
--- a/src/features/game/events/index.ts
+++ b/src/features/game/events/index.ts
@@ -751,6 +751,10 @@ import {
   type StartProjectAction,
 } from "./landExpansion/startProject";
 import {
+  collectProject,
+  type CollectProjectAction,
+} from "./landExpansion/collectProject";
+import {
   instaGrowFlower,
   type InstaGrowFlowerAction,
 } from "./landExpansion/instaGrowFlower";
@@ -1025,6 +1029,7 @@ export type PlayingEvent =
   | BurnClutterAction
   | InstantGrowProjectAction
   | StartProjectAction
+  | CollectProjectAction
   | InstaGrowFlowerAction
   | UpgradeRockAction
   | UpgradeTreeAction
@@ -1334,6 +1339,7 @@ export const PLAYING_EVENTS: Handlers<PlayingEvent> = {
   "clutter.burned": burnClutter,
   "project.instantGrow": instantGrowProject,
   "project.started": startProject,
+  "project.collected": collectProject,
   "trial.started": startTrial,
   "rock.upgraded": upgradeRock,
   "tree.upgraded": upgradeTree,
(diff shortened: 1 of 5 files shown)


eliasSFL:
@coderabbitai review

PR #7671, opened 2026-09-12
[FIX] Preserve cooking queues when using Instant Gratification

alexandrdragin:
Instant Gratification can finish queued meals as well as the currently cooking meal in Deli, Fire Pit, Kitchen, Smoothie Shack, and Bakery when recipes use windowed timers.

The skill zeroes the current recipe's `baseDurationMs` but leaves its old `startedAt`. The queue resolver therefore places completion at the original start and credits that elapsed time to subsequent recipes. For example, a recipe started 10 hours ago with 2 hours remaining, followed by two 2-hour recipes, makes the entire queue ready after activation.

Anchor the completed recipe to the activation time so only the current meal finishes immediately in each building and subsequent meals retain their cooking durations. In that example, readiness now becomes immediate, +2 hours, and +4 hours. Legacy timers are unchanged.

Validation:
- Added a parameterized regression covering all five cooking buildings, checking both cached and derived readiness. All five cases fail before the fix and pass afterward.
- `VITE_NETWORK=amoy jest ... --runInBand --silent`: 3 suites / 80 tests passed (`skillUsed`, `cancelQueuedRecipe`, `cookingReadiness`).
- `tsc --noEmit`, scoped ESLint, Prettier, and `git diff --check` passed.
- Without the network setting, two existing speed-boost tests fail identically on clean upstream/main; both pass in the amoy run above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Instant Gratification now completes only the active recipe across cooking buildings.
  * Queued recipes recalculate their readiness times correctly after activation.
  * Completed recipes no longer pass elapsed cooking time to queued recipes.
  * Chained recipes complete correctly even when an earlier recipe is finished but uncollected.
  * Recipes queued during Gourmet Hourglass retain their boosted durations.

### DIFF
--- a/src/features/game/events/landExpansion/skillUsed.test.ts
+++ b/src/features/game/events/landExpansion/skillUsed.test.ts
@@ -5,6 +5,9 @@ import { COOKABLES } from "features/game/types/consumables";
 import { FLOWER_SEEDS, FLOWERS } from "features/game/types/flowers";
 import { getFlowerReadyAt } from "features/game/lib/flowerBedReadiness";
 import { harvest } from "./harvest";
+import { getCookingQueueReadyAts } from "features/game/lib/cookingReadiness";
+import { BUILDING_DAILY_OIL_CAPACITY } from "./supplyCookingOil";
+import { getKeys } from "lib/object";
 import Decimal from "decimal.js-light";
 
 describe("skillUse", () => {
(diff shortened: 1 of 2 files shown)


src/features/game/events/landExpansion/skillUsed.ts
```diff
@@ -215,6 +215,9 @@ function useInstantGratification({
     // undo the instant completion (mirrors the oil reserve above / instaGrowFlower).
     if (queue[recipeIndex].baseDurationMs !== undefined) {
       queue[recipeIndex].baseDurationMs = 0;
+      // Complete at activation, not at the old start: queued recipes derive their
+      // start from this recipe and must not inherit its elapsed cooking time.
+      queue[recipeIndex].startedAt = createdAt;
```
eliasSFL, line 222: Nit on the comment: the tails do not end up chained off this recipe after the rebuild (`recalculateQueue` re-anchors the first upcoming recipe at `createdAt`). What the `startedAt` write really fixes is the ready/upcoming *split*, which is computed from the derived chain before the rebuild: a zero-work recipe derives `readyAt = startedAt`, so a stale start drags every chained successor into the ready half.

src/features/game/events/landExpansion/skillUsed.test.ts
```diff
@@ -1229,6 +1232,72 @@ describe("skillUse", () => {
...
+          },
+        };
+        const state = skillUse({
+          state: game,
+          action: { type: "skill.used", skill: "Instant Gratification" },
+          createdAt: now,
+        });
+        const queue = state.buildings[buildingName]![0].crafting!;
+        const expected = [now, now + 2 * HOUR, now + 4 * HOUR];
+        expect(queue.map((recipe) => recipe.readyAt)).toEqual(expected);
```
eliasSFL: Coverage: the fixture only exercises an anchored head with no boost windows, so the expected values are plain `startedAt + baseDurationMs` arithmetic. Two shapes the fix covers today but nothing pins:

1. The cooking recipe is **chained** (`recipeIndex > 0`, no `startedAt`, behind an uncollected ready recipe). Queue `[A ready at now-3h, B chained 5h, C chained 2h]`: pre-fix resolves to `[-3h, -3h, -1h]` (C is free), fixed to `[-3h, 0, +2h]` with `B.startedAt === now`. A future "only re-anchor recipes that already have a `startedAt`" refactor would pass the current suite and regress this.
2. An **active 2x Gourmet Hourglass** placed before the skill: tails should resolve to `[now, now+1h, now+2h]`.

Minor: the second `getCookingQueueReadyAts` assertion cannot diverge from the cached one (`recalculateQueue` writes the cache from that derivation); `Beta Pass` is not needed since jest runs with `VITE_NETWORK=amoy`; and the last assertion checks immer's input immutability rather than this fix.

eliasSFL (commented):
Verified the issue and the fix locally.

- **Issue is valid.** With the PR fixture and the fix line removed, the queue resolves to `[now-10h, now-8h, now-6h]` (whole queue ready). With the fix it resolves to `[now, now+2h, now+4h]`.
- **Fix is correct.** It also handles the shapes the new test does not cover: a *chained* cooking recipe (no `startedAt`, behind an uncollected ready recipe, windowed or legacy predecessor) and an active 2x Gourmet Hourglass.

Codex CI failing here is unrelated: secrets are unavailable to fork PRs (`Secret source: None`).

eliasSFL:
Ignore the Comments above was posted by the AI. 
I'll fix the api side of things

PR #7688, opened 2026-09-17
[FIX] Break crop sort ties consistently when checking weather-destroyed plots

adamhannigan:
## What

Client half of the fix for the `Plot is affected by greatFreeze` autosave 500s. API PR: https://github.com/sunflower-land/sunflower-land-api/pull/3641

`isCropDestroyed` sorts plots oldest-first to decide which half the Great Freeze, Tsunami or Tornado destroys. Its comparator (`b.createdAt > a.createdAt ? -1 : 1`) returns `1` both ways for plots with the same `createdAt`, so each JS engine orders ties differently. Expansion plots often share a millisecond. On Firefox the client could show a plot as harvestable that the server (Node) considers frozen, and the autosave failed.

## Fix

Ties now compare as equal (`a.createdAt - b.createdAt`). Sorts must be stable in every engine, so all engines pick the same plots. For Chrome/V8 the result is unchanged, which also matches the server, so this can ship before or after the API PR.

## Tests

Added `getAffectedWeather` tests in `plant.test.ts` that swap in a non-V8 (insertion) sort to stand in for Firefox. They fail with the old comparator and pass with the fix.

Note: 23 tests in `plant.test.ts` already fail locally on `main` without this change. They are unrelated to this fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved crop destruction ordering during great freezes.
  - Ensured plots with matching creation times are handled consistently across different environments, keeping client and server results aligned.


RyrukxD:
Dear Adam,

It is with a heavy heart that I write this message.

As a long-time Sunflower Land fan (Land 4019), I somehow managed to mess up my eligibility for the Yakkamon mint.

While I was in the top 20k, I underestimated the importance of the flower deposit. I mistakenly believed that the snapshot would capture my position on the leaderboard, allowing me to make the deposit afterward and still remain eligible.

I was a fool, and I understand my mistake now. However, I’m writing with the hope of a tiny mouse drowning in a bucket of water, desperately waiting to be rescued by a greater force, and perhaps given the chance to mint during Wave 4.

Would it be possible to add me to the eligible list? given that the attached screenshot shows that I made the deposit before the wave began sadly after the snapshot

One can only hope that this humble prayer reaches your heart and allows a lowly peasant like me to mint in 20 minutes or so.

My wallet address is:

"0x19eCf3906FA78461BF37e1A7fdb3Abc4E94b8854"

Thank you very much for your time, and especially for taking the time to read this.

Have a great afternoon!

Kind regards,
Elier Reyes

Screenshot_2026-09-17-01-31-15-826_com brave browser_beta

PR #7695, opened 2026-09-21
[FIX] Parse Fish Market queue requirements from the API into Decimals

eliasSFL:
## Problem
Players can't cancel queued Fish Market items: the confirm modal closes and nothing happens (reproduced on farm 62559).

Queued Fish Market `requirements` were persisted by the BE as decimal.js-light's raw internals (`{ "s": 1, "e": 0, "d": [8] }`) and returned by the API in that shape, so the cancel refund's `Decimal.add` threw and the event was dropped.

The root fix and the data migration live in sunflower-land/sunflower-land-api#3648: the BE now persists requirements as strings, and migration 028 rewrites the legacy shape.

## This PR
`makeGame` runs `buildings` through `makeBuildings`, which parses `processing[].requirements` strings back into Decimals, the same way inventory already is. That keeps the FE state matching its `Inventory` type.

This is type correctness rather than the fix itself: `Decimal.add("8")` already accepts strings, so the current FE works once the BE ships.

## Tests
New `transforms.test.ts`: string requirements become Decimals, and a queued item loaded from the API can be cancelled and refunded.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fish Market processing data now loads reliably from saved game data.
  - Processing requirements retain their correct decimal values after loading.
  - Cancelling a queued Fish Oil item now correctly refunds the required Tuna inventory.

- **Tests**
  - Added coverage for loading Fish Market data and cancelling queued processing items.


eliasSFL:
Closing: not needed. The fix lives entirely on the backend, and the current client already handles the string-serialised requirements.

